PR #9, opened 2019-05-02
Incorporate herbert

mducle:
This PR adds `Herbert` as a submodule within the Horace repository and modifies the initialisation file for Horace to look for Herbert in this folder and if it exists to add it to the path and run `herbert_init`.


mducle:
Decided not to pursue this action and to leave Herbert and Horace separate.

PR #11, opened 2019-05-09
fake sqw with q range

abuts:
This modifies fake_sqw to accept range of hkl values instead of detector positions to be able to compare the generated sqw object with the results of theoretical simulations. 

The validity of the operations are illustrated by unit test provided. If one generates sqw object using detectors and then deploy q-range as input to another fake sqw, the fake_sqw correctly recovers the detectors positions and produce the result, similar to the one, generated by initial fake_sqw

### DIFF
--- a/algorithms/build_det_from_q_range.m
+++ b/algorithms/build_det_from_q_range.m
@@ -1,5 +1,5 @@
 function  [det_pos,par_file_name] = build_det_from_q_range(q_range,efix,alatt,angdeg,u,v,psi,omega,dpsi,gl,gs,filename)
-% create fake detector file which would cover the q-range provided as input
+% Create fake detector file which would cover the q-range provided as input
 %
 % Inputs:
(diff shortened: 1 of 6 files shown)


algorithms/build_det_from_q_range.m
```diff
@@ -0,0 +1,92 @@
...
+end
+if size(q_range,2) ~=3
+    error('FAKE_SQW:invalid_argument',...
+        'second dimension of the q-range matrix should is not 3 but %d',...
+        size(q_range,2))
+end
+if all(size(q_range) == [3,3])
+    [q1,q2,q3] = ndgrid(q_range(1,1):q_range(1,2):q_range(1,3),...
+        q_range(2,1):q_range(2,2):q_range(2,1),...
+        q_range(3,1):q_range(3,2):q_range(3,1));
```
mducle: bug here. `q_range(2,1)` and `q_range(3,1)` should be `q_range(2,3)` and `q_range(3,3)`.

PR #16, opened 2019-09-19
Fix typo in combine_sqw_pix_job.m error message

g5t:
This very minor error-message bug will probably never be seen by anyone who is not also (incorrectly) modifying temporary SQW objects which they then write to disk before combining.


abuts:
Correct. Who would think one can work this way too :)

Issue #37, opened 2019-11-12
Document testing strategy


nickbattam-tessella:
It is essential that the updates to Horace

i) do not break existing functionality
ii) returns the same results when run in single threaded, multi-threaded (shared memory), and fully distrusted environments
iii) is performant and that performance scales

Outputs: Test Plan document.

Kevin-Woods-Tessella:
Initial draft looks good.  Let's publish it and update it as the project evolves.

Issue #32, opened 2019-11-11
C++ unit testing


nickbattam-tessella:
Identify the most appropriate C++ testing library and migrate any existing tests to this.

Outputs:
- ADR documenting selection
- Implementation of at least one test

hsautessella:
Googletest was identified as the preferred choice for testing framework.

1) It is in active development - receiving bug fixes
2) It is in use by other teams within ISIS (IBEX, Mantid uses it for the built-in mocking framework)
3) It is required by [Google Benchmark](https://github.com/google/benchmark#requirements) which may become useful in the future.

The fact that it was in use by other teams within ISIS outweighed the fact that Catch2 is more light-weight.

abuts:
Bearing in mind that the only unit tests currently existing for C code are the tests for CombineSQW, do not see any problem in changing to this. 

Matlab code tests should be still written on the basis of matlab_xunit  tests framework

PR #38, opened 2019-11-12
Initial draft of a Horace test plan

nickbattam-tessella:
Some initial thoughts on a testing strategy for Horace.

All feedback welcomed.

Issue: #37

### DIFF
--- a/documentation/stp/test-plan.md
+++ b/documentation/stp/test-plan.md
@@ -34,7 +34,7 @@ In addition:
 All (new) code will be tested at UNIT level
 
 - Where code is parallelized tests must be extended to cover execution on 1 and N (>1) threads to verify behaviour is unchanged
-- C++ code will be unit tested with CXX
+- C++ code will be unit tested with WHICH-TEST-FRAMEWORK
 - Matlab will be tested with Matlab class-based unittest Framework
 - Unit tests will be executed as part of the CI build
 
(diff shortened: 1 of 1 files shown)


documentation/stp/test-plan.md
```diff
@@ -0,0 +1,62 @@
...
+In addition:
+
+-	Automated system tests will be written by the developers to script the execution of a number of standard end-to-end workflows to measure performance on representative data sets and hardware.
+
+### Unit Testing
+
+All (new) code will be tested at UNIT level
+
+- Where code is parallelized tests must be extended to cover execution on 1 and N (>1) threads to verify behaviour is unchanged
+- C++ code will be unit tested with CXX
```
hsautessella: CXXTest has not had a release in 5 or 6 years https://sourceforge.net/p/cxxtest/blog/. Some more modern frameworks are [googletest](https://github.com/google/googletest) or, my personal preference (which I've used before), [Catch2](https://github.com/catchorg/Catch2). Catch has the advantage of being header only so is pretty painless to set up

nickbattam-tessella: The decision of which specific C++ test framework is covered under  #32. I can replace "CXX" in this document with whichever decision is reached there.

While Catch2 may be easier to setup, googletest is being used by the IBEX team and the googletest mocking framework by Mantid.

PR #51, opened 2019-11-13
Matlab license requirements documentation

nickbattam-tessella:
Document of which Matlab licences are needed on my current understanding of which our of build / test / execution platforms, looking at the range of deployment modes that are being discussed.

Please comment with any errors or additions.

You can view the rendered markdown here: https://github.com/pace-neutrons/Horace/blob/matlab-license/documentation/smg/03_matlab_licensing.md

### DIFF
--- a/documentation/smg/01_smg.md
+++ b/documentation/smg/01_smg.md
@@ -1,0 +1,4 @@
+# PACE - Horace System Maintenance Guide
+
+- [Development Environment](./02_dev-environment.md) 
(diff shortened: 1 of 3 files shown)


nickbattam-tessella:
@abuts that is an excellent question. I'll raise it with Toby.

nickbattam-tessella:
From Toby:  "if a pure Matlab fallback is to be retained, we only provide a single threaded version."

nickbattam-tessella:
This is part of #53

Issue #36, opened 2019-11-11
Use CMake to generate build files for mex libraries


hsautessella:
CMake offers a way to generate build files (e.g. Visual studio projects, makefiles, etc.) for multiple platforms (Windows, Linux, Mac), compilers and architectures. 

CMake is shipped with a [`FindMatlab` script](https://cmake.org/cmake/help/v3.16/module/FindMatlab.html) that can be used to find the Matlab compiler and generate mex files compatible with the target system.

PR #56, opened 2019-11-15
Use CMake to generate build files for mex libraries

hsautessella:
#### Description:
This PR writes some CMakeLists files to generate build scripts (e.g. Makefiles or Visual Studio Projects) for Horace's mex libraries in a platform agnostic way. 

The Visual Studio projects and pre-compiled mex files that were previously committed to the repository have been removed. The plan is that the build servers will produce compiled mex files, we will then distribute these files in Horace releases.

The structure of the `_LowLevelCode` directory has been changed slightly as well. There is no longer a `build_all` directory as CMake will generate a `BUILD_ALL` target for us. Instead, the directory contains a `cpp` folder containing the C++ source code, and an `external` directory, which currently holds the HDF5 libraries.

As part of setting up the CI environment a build script that will run CMake and build will be produced, but we leave this to a separate PR. For now, we must do this manually. Instruction are provided as part of this PR.

(Note there are a lot of files changed in this PR, however the vast majority of these are either deleting or moving files. So don't be put off!)

#### To test:
- Follow the instructions in `documentation/Building_with_CMake.md` to build the mex files.
- Launch Matlab (making sure you've run the Herbert and Horace `init` scripts) and use the `check_horace_mex` to validated that the mex files have been built correctly and can be read by Matlab.

Refs #36


abuts:
I think, necessary part of this work to be completed should be a Matlab script, which would allow users to build mex files on their platform without having cmake. Currently we have horace_mex (herbert_mex) files which do this job (herbert_mex is not automated so should be modified) for Unix but windows version would do. 

I see hdf headers removed. How would users (under linux) would build Matlab-hdf correspondent libraties?

hsautessella:
I think the idea here is that developer builds will require CMake, however the releases we ship to users will have pre-built mex libraries included (so users will have no need to build anything). If users wish to compile their own mex libraries, they can do this in the usual way - using the `mex` command within the Matlab command window.

I haven't removed any HDF headers, I've only moved them and added some UNIX libraries to v1.8.12, that were missing.

abuts:
Not sure it is possible to do for all range of the architectures, the Matlab is running.  If it is the case, than happy to go this way

Have not noticed the move of the headers. Have not looked too accurately into that yet. Under Unix we need only the headers, as the build shold be done against Matlab libraries as any other hdf libraries on the system will cause a lot of problem. Under Windows, you also should build agains Matlab dll-s but lib-s are, of course, necessary .

nickbattam-tessella:
@abuts The generated makefiles (or equivalent .vcproj files for VStudio) can be included in the build artifacts and shipped with the release.

From the end-users point of view this means there would be difference between these auto-generated build files and the current manually maintained ones. 
Critically they would need no knowledge of CMake to use them, the will probably just need to update library paths to something appropriate for their environment.

abuts:
@nickbattam-tessela -- I do not think users should be aware of VS too -- majority do not want to know what it is. They want either not do anything (mex-files provided) or run simple command (horace_mex) to get the mex files running. 

make is usually on a system, so If if is launched from Matlab it would be ok. Do not know how to easily modify the path-s though without running cmake.

nickbattam-tessella:
The `horace_mex.m` script can be updated to build using these CMake files. We *may* need to add cmake as a dependency but that is simple for anyone with the necessary build tools (gcc/visual studio) to install with some document guidance.

Added new issues #57 to cover this development

PR #61, opened 2019-11-22
Set up cpp testing framework

hsautessella:
This PR sets up a C++ testing framework in Horace.

- CMake has been configured to download GoogleTest at configure time. Targets are then generated to build it.

- A `BUILD_TESTS` switch has been added to CMake so the tests need not be built if they are not required (e.g. for a release candidate).

- A `horace_add_unit_test` function has been defined in CMake to make it easy to add tests.

- One of the `combine_sqw` tests has been ported to use GoogleTest.

- A new entry has been added to the SMG describing how to add and run C++ tests.

refs #32


_LowLevelCode/cpp/test/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+set(TESTS_BIN_DIR "${CMAKE_BINARY_DIR}/tests/bin")
+
+# On Windows we need to point the tests to the Matlab dlls. Define a
+# TEST_ENV_PATH variable that has the location prepended to the system path.
```
nickbattam-tessella: the Matlab path is appended to the PATH not prepended

_LowLevelCode/cpp/test/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+# When adding tests, this can be set as the system path for those tests.
+if(WIN32)
+    file(TO_CMAKE_PATH "$ENV{PATH}" SYS_PATH)
+    set(TEST_ENV_PATH "${SYS_PATH};${Matlab_DLL_DIR}")
+    string(REPLACE ";" "\;" TEST_ENV_PATH "${TEST_ENV_PATH}")
```
nickbattam-tessella: what's this `;` magic doing?

_LowLevelCode/cpp/test/combine_sqw.tests/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+set(TEST_NAME "combine_sqw.test")
+
+set(TEST_SRC_FILES
+    "combine_sqw.test.cpp"
+)
+
+set(SRC_FILES
+    "combine_sqw.test.cpp"
```
nickbattam-tessella: this has now moved to `TEST_SRC_FILES`

_LowLevelCode/cpp/test/combine_sqw.tests/combine_sqw.test.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include <vector>
+
+/*
+ * Retrieve an environment variable from the system.
+ * This function is necessary as `std::getenv` raises security warnings on
+ * Windows.
+ *
+ * @param key :: The key of the environment variable
+ */
+std::string get_enviornment_var(const std::string &key) {
```
nickbattam-tessella: this should be defined outside this file -- it's a env_helper utility file that will be potentially used by every test file

cmake/horace_FindGTest.cmake
```diff
@@ -0,0 +1,78 @@
+#[=======================================================================[.rst:
+horace_FindGTest
+-----------------
+
+Download and build GoogleTest and GMock. Then export variables with which we
```
nickbattam-tessella: should this documentation include a link to the GoogleTest site?

documentation/smg/05_adding_and_running_cpp_unit_tests.md
```diff
@@ -0,0 +1,94 @@
...
+
+The test executables can also be run directly, with a couple of caveats:
+
+1) On Windows, if the test uses any mex libraries, then it must be able to
+locate some Matlab DLLs; the path to these needs to be on your system path. The
+DLLs are located at `<Matlab_Root>/bin/win64` for a 64-bit build (when running
+the tests with CTest, this is automatically added to your path for the duration
+of the tests).
+
+2) Since all data file paths are written relative to the `HORACE_ROOT`
```
nickbattam-tessella: this could be clearer; explicitly say the if not defined it will be CWD

_LowLevelCode/cpp/test/combine_sqw.tests/combine_sqw.test.cpp
```diff
@@ -0,0 +1,193 @@
+#include "combine_sqw/combine_sqw.h"
```
nickbattam-tessella: where is the original CppTest version of this test?

nickbattam-tessella: it's not been deleted

_LowLevelCode/cpp/test/combine_sqw.tests/combine_sqw.test.cpp
```diff
@@ -0,0 +1,162 @@
...
+protected:
+  static std::vector<uint64_t> sample_npix;
+  static std::vector<uint64_t> sample_pix_pos;
+  static std::vector<float> pixels;
+  static std::string test_file_name;
+  static std::size_t num_bin_in_file, bin_pos_in_file, pix_pos_in_file;
+
+  // Called once, before the first test is executed.
+  static void SetUpTestSuite() {
+    const auto HORACE_ROOT_DIR = get_env_variable("HORACE_ROOT");
```
nickbattam-tessella: would it be helpful to define `HORACE_ROOT` as a const somewhere? that somewhere would then be a good reference for "these are the environment variables we're going to use in the code"

_LowLevelCode/cpp/test/utility/environment_helpers.cpp
```diff
@@ -0,0 +1,33 @@
...
+ * Windows.
+ *
+ * @param key :: The key of the environment variable
+ */
+std::string get_env_variable(const std::string &key) {
+#ifdef _WIN32
+  // Get the required size of the variable
+  std::size_t requiredSize;
+  getenv_s(&requiredSize, NULL, 0, key.data());
+  if (requiredSize == 0) // No env variable called 'key' exists
```
nickbattam-tessella: should we be using `if (...)` without `{ }` ?

_LowLevelCode/cpp/utility/environment.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Horace {
+namespace Utility {
+namespace Environment {
+/*
+ * Retrieve an environment variable from the system.
+ * This function is necessary as `std::getenv` raises security warnings on
+ * Windows. Also note, GCC does not implement `getenv_s`.
+ *
+ * @param key :: The key of the environment variable
```
nickbattam-tessella: missing a note on the return parameter saying it returns empty if variable not found

_LowLevelCode/cpp/utility/environment.h
```diff
@@ -0,0 +1,13 @@
+#include <string>
+
+namespace Horace {
+namespace Utility {
+namespace Environment {
+// Get an environment variable
+std::string get_env_variable(const std::string &key);
+// The HORACE_ROOT environment variable
+const std::string HORACE_ROOT{get_env_variable("HORACE_ROOT")};
```
nickbattam-tessella: I hadn't thought of doing it this way; I was expecting to just have a `HORACE_ROOT` defined as the constant ("HORACE_ROOT") and leaving it to users to call the function as needed.

nickbattam-tessella: this has an overhead as we add more env variables and doesn't make the "failure modes" explicit
.

_LowLevelCode/cpp/utility/environment.cpp
nickbattam-tessella: would it make it "easier" for users if the API was `string get_variable(string key, string default)` where default (not empty) was returned if the key doesn't exist.

That way anyone calling can be sure the value is populated correctly on return and doesn't need to add a `if(isempty(value)) { value = default }` block of code  Thinking specifically of needing `.` for HORACE_ROOT.

_LowLevelCode/cpp/utility/environment.h
hsautessella: Ah I see, no problem

_LowLevelCode/cpp/test/CMakeLists.txt
hsautessella: `;` is a list separator in CMake, so to get the string, we need to escape it. I've added a comment explaining the need for the escape character.

Issue #62, opened 2019-11-26
Use the HDF5 libraries shipped with Matlab on Linux


hsautessella:
Using the HDF5 libraries that are shipped with Matlab should hopefully guarantee compatibility if we wish to read/write the same HDF5 files in Matlab and C++. 

The HDF5 libraries for Linux we currently have in this repository only compile when linked via `target_link_libraries(myLibrary ${HDF5_CXX_LIBRARIES})` whereas on Windows we use `${HDF5_LIBRARIES}`.

hsautessella:
Closed by #63

Issue #31, opened 2019-11-11
Dependency management plan


nickbattam-tessella:
C++ build will include dependencies on a number of third-party libraries, e.g. boost, mpi, hdf5, testing framework...

Output:
- A plan for managing this should be created and filed as an ADR

nickbattam-tessella:
MATLAB enforces strict versioning of libraries, often not-current releases.

As part of this investigation into dependency management we must find a solution to map the correct library to the required MATLAB version

e.g.
- HDF5 1.8.12 for <current releases of MATLAB>
- HDF5 1.8.6 for <legacy releases of MATLAB> TODO: confirm WHICH version
- build against HDF5 libraries found on the MATLAB path for Linux
- build against HDF5 libraries managed locally for Windows as the MATLAB doesn't include the static libraries
- FAIL the build if the required headers or libraries matching the found MATLAB version don't exist

There is no simple reference on the Mathworks site for which MATLAB releases are shipped with which library.

hsautessella:
As part of this ticket, we should also look into how we deal with other PACE dependencies and modelling codes, e.g. [Herbert](https://github.com/pace-neutrons/Herbert) and [Euphonic](https://github.com/pace-neutrons/Euphonic)

nickbattam-tessella:
Time boxed based document decisions based on current understanding

hsautessella:
Closed by #132

PR #63, opened 2019-11-26
Use the HDF5 libraries shipped with Matlab on Linux

hsautessella:
Tell CMake to link against the HDF5 libraries within Matlab's bin directory. 

Delete the Linux HDF5 libraries that exist within this repo.

refs #62


nickbattam-tessella:
Management of the tight coupling between HDF5 version and MATLAB release will be handled as part of #31

Issue #75, opened 2019-12-03
tmp folder in iDaaaS


abuts:
currently unit tests and some small parts of Horace code use /tmp folder to keep intermediate information.  This does not work on iDaaaS as this folder is cleared up as soon as any substantial node load is encountered. Custom tmp folder function written in Herbert and this changes to use this function in Horace

abuts:
the task is successfully completed

PR #76, opened 2019-12-03
Re #75 Replaced all occurencies of tempdir by tmp_dir

abuts:
This simple change should allow running all unit tests on iDaaaS


abuts:
Run the modified tests on isiscompute, small machine trhough validate_horace/validate_herbert scripts and only three Horace tests  have failed -- one looks like new problem with update_sqw operation -- need investigations but this is the part of code can not be affected our changes and two old failures -- Tobyfit gets different results on different machines and I actually do not know what to do about this. The change is very useful and big success, as before it it was like ~50 tests failing.

PR #77, opened 2019-12-05
71 code relocation

abuts:
the code is relocated and all unit tests work as before. The further development can be continued under another tickets.


_test/test_sqw/test_rundata_vs_sqw.m
```diff
@@ -37,7 +37,7 @@ function rm_sqw(filename)
                 name = varargin{1};
             end
             this = this@TestCase(name);
-            root_dir = fileparts(which('horace_init.m'));
+            root_dir = fileparts(fileparts(which('horace_init.m')));
```
nickbattam-tessella: `fileparts(fileparts(...))` ?

nickbattam-tessella: this is repeated in a few places; in other places you have `fullfile(fileparts(...))`

_test/test_change_crystal/test_change_crystal_1.m
```diff
@@ -9,8 +9,9 @@
 
 % -----------------------------------------------------------------------------
 % Add common functions folder to path, and get location of common data
-addpath(fullfile(fileparts(which('horace_init')),'_test','common_functions'))
-common_data_dir=fullfile(fileparts(which('horace_init')),'_test','common_data');
+horace_root = fileparts(fileparts(which('horace_init')));
```
nickbattam-tessella: this single line appears dozens of times; would it be helpful to define it as a tiny helper function `get_horace_root()`?

_test/test_sqw/test_rundata_vs_sqw.m
abuts: I though about it and had that temptation. Left this as it is for one reason: How many times we will relocate the code? Hopefully 0. How would a new person who comes to do some job with the code read the code? He would have a look and the code and this code is clear.  If there is some function, doing the same job (I tried it in Herbert) one needs to have a look and understand what is going on. Similarly, if one writes new unit test and relies on its location within the package -- its trivial to drop a line like this. A new person or I myself coming there after 2years break will not know where to look for the function, providing the correct path. This is proven by my experience with Herbert where I tried to make such function and already forgot its name. 
So that was the reason behind this decision -- ready to discuss if it valid or not.

_test/test_change_crystal/test_change_crystal_1.m
abuts: again, the code becomes unclear -- root -- is it the place where matlab code resides? is it where .git is located?

_test/test_sqw/test_rundata_vs_sqw.m
nickbattam-tessella: The function documentation would make it clear what was mean by "root"; you obviously have a clear idea as that definition is being used consistently across the test and source files.

Based on that it's then much easier to have a single "get_root()" function call than remembering that you mean the folder two up from where `horace_init` lives?

I totally agree that adding the line you have to the top of the files is trivial, but it's not obvious what it's doing (to someone like me who is not familiar with the code or folder structure) and is easy to get wrong.

I think creating a get_root() helper and putting it in admin or utilities (as appropriate) would be a good change in Horace and Herbert

PR #83, opened 2019-12-11
39 anvil win node

abuts:
Small document describing current state of the Anvil node


documentation/smg/07_Windows_Anvil_Node.md
```diff
@@ -0,0 +1,23 @@
+# Windows Anvil Node
```
nickbattam-tessella: can you add:
- the name of the build node as it appears in Anvil somewhere `PACE Windows (Private)`
- for completeness (in case your Download folder is emptied) the java command to launch the agent (as in the .bat file).

Issue #43, opened 2019-11-13
Create Linux CI build


nickbattam-tessella:
Create a Jenkins build and test job for the LINUX build.

Part of #30 

Output:
- bash script to build and test
- jenkins script to conditionally run 
- jenkins configuration to trigger builds

nickbattam-tessella:
Blocked waiting for plugins to be installed on ANVIL: 
•	Generic Webhook Trigger https://plugins.jenkins.io/generic-webhook-trigger - Allows GitHub to trigger builds
•	httpRequest https://plugins.jenkins.io/http_request - Send GitHub statuses
•	xUnit https://plugins.jenkins.io/xunit - Test parsing and visualisation
•	Rebuilder https://plugins.jenkins.io/rebuild - Rebuild failed builds
•	Blue Ocean https://plugins.jenkins.io/blueocean - Better pipeline construction and visualisation 
o	  Including Git Pipeline for Blue Ocean and GitHub Pipeline for Blue Ocean
•	Delivery Pipeline https://plugins.jenkins.io/delivery-pipeline-plugin - Multi-job pipeline and deployment visualisation
•	Performance https://plugins.jenkins.io/performance - Display and track benchmarking results
•	Warnings Next Generation https://plugins.jenkins.io/warnings-ng - Parse and display warnings from builds

hsautessella:
Removing blocked label, all plugins above except httpRequest and Blue Ocean have been installed. 

There are plans to install blue ocean down the line

PR #89, opened 2019-12-19
Create Scientific Linux CI build

hsautessella:
This PR sets up our Jenkins CI to automatically build pull requests submitted to Horace, currently these builds are solely on Scientific Linux.

The basic logic is:
- Setup a webhook on GitHub to trigger a build on Jenkins after a pull request event
- Checkout the branch linked to the pull request and merge with master
- Run the `Jenkinsfile` in `tools/build_config`. This:
    - calls the build script with `./build.sh --build`  - builds mex files
    - calls the build script with `./build.sh --test`  - runs tests using CTest
    - calls the build script with `./build.sh --package`  - packages Horace into zip (not yet implemented)
- Post statuses back to GitHub with whether the build succeeded or failed.

Some work is obviously done within the Jenkins GUI (the logic for checking out the PR) so reviewers should check the `PR-Scientific-Linux-7-2018b` job on ANVIL as well.

Matlab tests are currently not run, but this PR provides the basic CI setup. Getting the Matlab tests running will require putting the relevant commands in the `tools/build_config/build.sh` script (including cloning Herbert).

There is also more work to be done to set up the Windows and Mac builds.

refs #43


tools/build_config/build.sh
```diff
@@ -0,0 +1,101 @@
...
+    echo "+ $1"
+    eval "$1"
+}
+
+function warning {
+    echo -e "\e[33m$1\e[0m"
+}
+
+function run_build() {
+    echo_and_run "cd ${build_dir}"
```
nickbattam-tessella: it would be cleaner to pass the variables in as arguments (`$build_config`; `$build_dir`) rather than use global variables

```
$build_dir = $1
$build_config = $2
```
and similarly `$install_dir` in the package step.
You can use `local` to control the scope of these function-scoped variables

tools/build_config/build.sh
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env bash
+
+set -o errexit  # exit early on any error
+set -o nounset  # raise error using unused variables
+
+FALSE=0
+TRUE=1
+CMAKE_GENERATOR="Unix Makefiles"
+HORACE_ROOT=$(realpath $(dirname "$0")/../..)
```
nickbattam-tessella: it may be helpful to add some comments explaining these two magic path calculations

tools/build_config/build.sh
```diff
@@ -0,0 +1,101 @@
...
+    echo_and_run "${build_cmd}"
+}
+
+function run_tests() {
+    echo -e "\nRunning test step..."
+    test_cmd="ctest -T Test --no-compress-output"
+    echo_and_run "$test_cmd"
+}
+
+# not yet implemented
```
nickbattam-tessella: arguably if it's not implemented it shouldn't be here....

tools/build_config/build.sh
```diff
@@ -0,0 +1,101 @@
...
+        -p|--package) package=$TRUE; shift ;;
+        # options
+        -X|--build_tests) build_tests="$2"; shift; shift ;;
+        -C|--build_config) build_config="$2"; shift; shift ;;
+        -O|--build_dir) build_dir="$(realpath $2)"; shift; shift ;;
+        -I|--install_dir) install_dir="$(realpath $2)"; shift; shift ;;
+    esac
+done
+
+# print versions
```
nickbattam-tessella: Is "print versions" a useful comment? that whole block could move into a function `print_tool_versions()` which would remove the need for the comment completely.

tools/build_config/Jenkinsfile
```diff
@@ -0,0 +1,99 @@
...
+
+        stage('Package') {
+            steps {
+                sh './tools/build_config/build.sh --package'
+            }
+        }
+    }
+
+    post {
+        always {
```
nickbattam-tessella: does always happen even if the build fails? does that archive no files.

xunit() is instructing the XUnit plug in to display the archived XML file(s)?
why `deleteOutputFiles: true,` if you're separately archiving them?

PR #94, opened 2019-12-23
Jenkins ADR

nickbattam-tessella:
New Decision Records covering decisions made regarding Jenkins build pipelines, job structures and the management of Horace/Herbert dependency in them.

### DIFF
--- a/documentation/adr/0001-record-architecture-decisions.md
+++ b/documentation/adr/0001-record-architecture-decisions.md
@@ -2,7 +2,7 @@
 
 # 1. Record architecture decisions
 
-Date: 2019-11-04
+Date: 2019-Nov-04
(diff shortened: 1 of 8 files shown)


nickbattam-tessella:
After an update to Herbert the nightly build of Horace will pick up the changes and trigger a build+test.
If there are any problems we will see them the following morning. A manual build can be triggered if feedback is required more rapidly.

It should be possible to configure the Herbert build to trigger a (parameterised) Horace master build automatically if desired with an additional build step, e.g. 
```
steps {
    build job: 'downstream-job-name', parameters: [
        string(name: 'upsteam_project_name', value: env.NAME)
    ], wait: false
}
```

PR #95, opened 2019-12-23
Initial version of DoD

nickbattam-tessella:
Initial version for discussion.


nickbattam-tessella:
I agree; code reviews shouldn't take long but will add to knowledge sharing and hopefully catch some errors before they're merged in to `master`.
Quick reviews are easiest when there are only a small number of related changes on the pull-request; so keeping things small should be our goal.

Issue #13, opened 2019-07-03
apply_symop method on sym_op class


abuts:
Currently sym_op class works as input to cut_sym_sqw methods transforming range of the cuts, produced by this method before adding these cuts together. 

cut_sym_sqw does not implement full range of possible cuts (e.g. cellarray of projections does not work) 
In addition to that, it is currently impossible to make modifications to cuts, produced by cut_sym_sqw (e.g. fix magnetic form factor or remove background) before combining these cuts together. 

It would be beneficial to modify sym_op to add the method which apply the appropriate sym_op operation to input sqw object. It should be possible to combine number of such objects using usual combine operations. 

All necessary code is present in Horace, so ony minor refactoring is necessary to implement such operation.

abuts:
This is superseedeed by #107

PR #107, opened 2020-01-22
symop improvements

g5t:
Partially addressing the issues raised in https://github.com/pace-neutrons/Horace/issues/13:

- Fixed a bug whereby `[sym_sqw, sym_cuts] = cut_sqw_sym(sqw, proj, ...)` returned the same information for its first and second outputs instead of the expected vector (or cell array) of symmetry related cuts.
- Fixed a buy where empty symmetry-equivalent cuts could not be handled by `cut_sqw_sym`.
- Added the ability to define a `symop` by its matrix representation.


g5t:
I've added unit testing for the added and fixed features, and rebased my changes into a single commit.

PR #110, opened 2020-01-29
Run matlab tests in ci builds

### LINKED ISSUE #91: Run Matlab tests in CI builds
Currently, Matlab tests are not run in our CI builds.

To run the tests we will need to clone Herbert somewhere in the Jenkins workspace. We'll need to build Herbert's mex files as well unless we can grab the binaries from elsewhere?

We'll also need to resolve #86 so that Matlab can find the mex files that are compiled, and we can test against them.

Outcome:
- [x] MATLAB tests executed as part of HERBERT build on Jenkins
- [ ] MATLAB tests executed as part of HORACE build on Jenkins
- [ ] MATLAB tests can still be executed on Developer machine

hsautessella:
This PR enables Matlab tests for Horace on Jenkins.

- Horace pipelines pull the most recent (successful) nightly version of Herbert from Jenkins
- Built mex libraries are copied into `horace_core/DLL`
- The most recent CMake FindMatlab script is included in the repo, so we get consistent build behaviour regardless of the version of CMake you're using (as long as it's >=3.7)

Several tests that were failing have been disabled in order to get a successful build running. An issue will be opened to fix and re-enable these tests.

Fixes #91


admin/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+# Copy template files into <build_dir>/horace_config
+set(WORKER_CONFIG_DIR "${CMAKE_BINARY_DIR}/horace_config")
+file(MAKE_DIRECTORY ${WORKER_CONFIG_DIR})
+
+# Format and copy horace_on template
+file(READ "horace_on.m.template" _horace_on_template)
+string(REPLACE
+    "default_horace_path ='/usr/local/mprogs/Horace/horace_core';"
```
nickbattam-tessella: these explicit replacements are very fragile; if the template is ever updated this will fail; can you replace it with a "replace `default_horace_path =*`" match?

tools/build_config/build.sh
```diff
@@ -12,7 +12,7 @@ readonly HORACE_ROOT="$(realpath $(dirname "$0")/../..)"
 # matlab executable. The Matlab on the path will likely be a symlink so we need
 # to resolve it with `readlink`
 readonly MATLAB_ROOT="$(realpath $(dirname $(readlink -f $(which matlab)))/..)"
-readonly MAX_CTEST_SUCCESS_OUTPUT_LENGTH=4096 #bytes
+readonly MAX_CTEST_SUCCESS_OUTPUT_LENGTH="10000" # 10 kilobytes
```
nickbattam-tessella: 10kb is 10240 bytes (using the same definition of KB as we were for 4k == 4096 kytes)

nickbattam-tessella: Ignoring the 2^n vs 10^n debate; does this number need to be a string?

hsautessella: I'm sure I changed it for a reason... But it seems to work as a number

_test/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+            exit(~res)"
+        TIMEOUT ${TEST_TIMEOUT_LENGTH}
+    )
+    # The MPI tests require that the `[herbert|horace]_on` and `worker_v2`
+    # scripts be on the path for all workers when Matlab starts up: add the
+    # path of the directory it's copied into to the MATLABPATH environment
+    # variable when tests are run.
+    set_tests_properties("${TEST_NAME}"
+        PROPERTIES
+            ENVIRONMENT "MATLABPATH=${MATLABPATH}:${CMAKE_BINARY_DIR}/horace_config"
```
nickbattam-tessella: does this need the Herbert `local_init` folder adding (c.f. https://github.com/pace-neutrons/Herbert/blob/master/_test/CMakeLists.txt)?

admin/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+# Copy template files into <build_dir>/horace_config
+set(WORKER_CONFIG_DIR "${CMAKE_BINARY_DIR}/horace_config")
+file(MAKE_DIRECTORY ${WORKER_CONFIG_DIR})
+
+set(_path_regex "[a-zA-Z0-9/\\\\_\\-\\.: ]+")
```
nickbattam-tessella: the comment we have in the Herbert file explaining this regex needs to be copied here

_test/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+            exit(~res)"
+        TIMEOUT ${TEST_TIMEOUT_LENGTH}
+    )
+    # The MPI tests require that the `[herbert|horace]_on` and `worker_v2`
+    # scripts be on the path for all workers when Matlab starts up: add the
+    # path of the directory it's copied into to the MATLABPATH environment
+    # variable when tests are run.
+    set_tests_properties("${TEST_NAME}"
+        PROPERTIES
+            ENVIRONMENT "MATLABPATH=${MATLABPATH}:${LOCAL_INIT_DIR}"
```
nickbattam-tessella: need to add logic to handle the Windows `;` vs Linux `:` path separator

admin/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+# Copy template files into <build_dir>/horace_config
+file(MAKE_DIRECTORY ${LOCAL_INIT_DIR})
```
nickbattam-tessella: it's not obvious what this path is or where it came from

_test/CMakeLists.txt
nickbattam-tessella: it's not obvious what this path is or where it came from

CMakeLists.txt
```diff
@@ -15,19 +15,26 @@ if(POLICY CMP0074)
...
-endif()
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH
+    "${CMAKE_SOURCE_DIR}/cmake"
+    "${CMAKE_SOURCE_DIR}/cmake/external"
+)
+
+# Set some useful global variables
+# this will hold herbert/horace_on.m and worker_v2.m
+set(LOCAL_INIT_DIR "${CMAKE_BINARY_DIR}/local_init")
```
nickbattam-tessella: should thie `LOCAL_INIT_DIR` be set in the HERBERT build environment too; it's set locally in `_test\CMakeLists.txt` as `${CMAKE_BINARY_DIR}/local_init"`

hsautessella: Yes it should, I just haven't gotten around to it yet

abuts:
Also depends on #189

PR #119, opened 2020-02-04
Implement packaging step

hsautessella:
This PR adds CMake `install` commands to generate a packaged version of Horace - that also contains Herbert. It also adds the ability to run `make install` or `MSBuild INSTALL.vcxproj` to install a Horace package in the directory specified by `CMAKE_INSTALL_PREFIX`.

This is the Horace equivalent of https://github.com/pace-neutrons/Herbert/pull/44

The directory structure of the package is:
```
Horace-<version>[-<date>]-<os>-<matlab release>[-<git sha>]
    +- Herbert
        +- <contents of herbert_core>
    +- Horace
        +- <contents of horace_core>
    +- herbert_on.m
    +- horace_on.m
    +- worker_v2.m
    +- LICENSE
    +- README..md
```
I'm would welcome feedback on the above structure.


cmake/horace_CPackConfig.cmake
```diff
@@ -0,0 +1,16 @@
+# For the list of CPack variables see https://cmake.org/cmake/help/v3.7/module/CPack.html
+
+set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_SOURCE_DIR}/LICENSE")
+set(CPACK_RESOURCE_FILE_README "${CMAKE_SOURCE_DIR}/README.md")
+set(CPACK_PACKAGE_VERSION_MAJOR "${${PROJECT_NAME}_VERSION_MAJOR}")
+set(CPACK_PACKAGE_VERSION_MINOR "${${PROJECT_NAME}_VERSION_MINOR}")
+set(CPACK_PACKAGE_VERSION_PATCH "${${PROJECT_NAME}_VERSION_PATCH}")
+set(CPACK_PACKAGE_FILE_NAME "${PROJECT_NAME}-${${PROJECT_NAME}_FULL_VERSION}")
```
nickbattam-tessella: the Herbert equivalent uses 
```
set(CPACK_PACKAGE_FILE_NAME "${PROJECT_NAME}-${Herbert_FULL_VERSION}")
```
and you refer to Horace_FULL_VERSION in the `horace_Version.cmake` script

cmake/horace_Version.cmake
```diff
@@ -0,0 +1,49 @@
+set(Horace_FULL_VERSION "${PROJECT_VERSION}")
```
nickbattam-tessella: missing header text (see Herbert file)

cmake/horace_Version.cmake
```diff
@@ -0,0 +1,49 @@
...
+        "\\1"
+        ${ARGV1}
+        ${${ARGV1}}
+    )
+    string(REPLACE " " "." ${ARGV1} ${${ARGV1}})
+    string(REPLACE "\"" "" ${ARGV1} ${${ARGV1}})
+endmacro()
+
+if(UNIX)
+    # Get distribution name from /etc/os-release
```
nickbattam-tessella: Herbert simplified this to `linux`

admin/CMakeLists.txt
```diff
@@ -34,3 +34,27 @@ file(WRITE "${LOCAL_INIT_DIR}/herbert_on.m" "${_herbert_on_script}")
...
 # Copy Herbert's worker template
 configure_file("${Herbert_WORKER_TEMPLATE}" "${LOCAL_INIT_DIR}/worker_v2.m")
+
+# =============================================================================
+# Install commands
+# =============================================================================
+# Function to install a file whose name ends in .template
+function(install_template_file)
+    set(_template_file "${ARGV0}")
+    get_filename_component(_template_base_name "${_template_file}" NAME)
```
nickbattam-tessella: is this different from the Herbert variant to handle file files with paths?

CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ set(LOCAL_INIT_DIR "${CMAKE_BINARY_DIR}/local_init")
 option(BUILD_TESTS "Build the C++ tests" ON)
 
 # Look for packages early so we can exit if they're not found
+include(horace_Version)
```
nickbattam-tessella: is there any reason for horaceVersion to be the first include?  Is there any ordering required on these imports?

cmake/horace_CPackConfig.cmake
hsautessella: Which do you think is best using `${horace_FULL_VERSION}` or `${${PROJECT_NAME}_FULL_VERSION}`.

The second is definitely more general - so when we come to extracting out common CMake functions, we can use the latter. But, at this point there's no practical distinction. Unless you think it's better to be consistent between the two repos? Or we could raise an issue to alter Herbert

admin/CMakeLists.txt
hsautessella: Out of necessity yes. In Herbert all the template files were in the `admin` directory and were passed into the function as relative paths. We now want to pass some absolute paths - which we need to trim down or the function will attempt to install the files to 
`C:\<CMAKE_INSTALL_PRFIX>\.\C:\absolute\path\to\template\file.template`.

cmake/horace_CPackConfig.cmake
nickbattam-tessella: it's more consistent to be using `$PROJECT_NAME` in this file; it is much less readable though, so definitely keep `Horace_` in the other scripts.

I'd suggest small PR in Herbert to make the scripts consistent once Horace is finished.

admin/CMakeLists.txt
nickbattam-tessella: where does the absolute path come from?

nickbattam-tessella: add comment to say this comes from find_herbert

CMakeLists.txt
hsautessella: I don't think so, I'll move it down below the FindMatlab stuff since it uses variables defined in there - but more for our own sanity

hsautessella: Actually - the Matlab version wasn't being added to the packages... whoops

admin/CMakeLists.txt
hsautessella: That's what the `Herbert_` prefix in the variable names being passed to the function is supposed to signify. It's common CMake convention - e.g. we know that all variables starting `Matlab_` come from the Matlab module

nickbattam-tessella: here it's just `_template_file`  though with no clue where it's from.

Following back to where this is called you do have `Herbert_WORKER_TEMPLATE` so what is actually needed is a change to the `horace_FindHerbert.cmake` header to say this will be an absolute path.

nickbattam-tessella: All found paths are absolute; there is nothing to clarify here.

PR #127, opened 2020-02-12
Sqw design

nickbattam-tessella:
Documentation for SQW review + redesign


documentation/add/07-sqw_redesign.md
```diff
@@ -0,0 +1,541 @@
...
+**Notes**
+(1): What data subset of should be displayed at the console or in an text  export file?
+
+#### Data manipulation
+
+- Implement basic arithmetic operations on object data. 
+
+- Cut -- extract a N-dimensional subset of pixels
+
+- Projection -- aggregate pixel data onto an M-dimensional surface (M<N). This should support projections onto planes and spherical shells (TBI) and spiral slices (TBI).
```
g5t:
```suggestion
- Projection -- aggregate pixel data onto an M-dimensional surface (M &ge; 0). This should support projections onto planes and spherical shells (TBI) and spiral slices (TBI).
```
I don't think that we want to restrict to dimensional reducing projections. A reorientation is a projection with `M` &equiv; `N` and 'extra' dimensional projections will have `M` > `N`

rebeccafair:
It may also need to be considered what (if any) structure/symmetry information is required in S(Q,w) for simulation software, from the point of view of converting Q in Horace reciprocal lattice units to Q in simulation reciprocal lattice units. I've expanded on this in https://github.com/pace-neutrons/Pace-Project-Plan/issues/19.

PR #131, opened 2020-02-21
Bug in write nsqw

abuts:
When write_nsqw_to_sqw is executed with 'drop_subzone_headers' option, it fails on the headers, which do not obey the format of the subzone headers (despite this is not actually a subzone headers at all). The change forces code to correctly identify this situation as "not a subzone header"


abuts:
@nickbattam-tessella  it is job for the Toby's user who is doing custom sequence of the cuts, processing them and need to combine all her cuts together. The script I sent to Toby is not the part of Horace, so not reflected here (which would make things much clearer if it is here)

I hope our changes will make this work not a hack but a normal operation :)

The fix is based on incompleted job on parallel symmetrization, which should be done differently. 
(Through get_pixels interface to sequence of cuts, probably)

But the part of current implementation is split the cuts into smaller pieces and process these pieces in paraellel. To idenfify these pieces, their filenames were modified according to the artificial format, to tell that they are the pieces. As this is internal part of the job, the format is internal and was not intended to go anywhere. 

But this part of the implementation appeared useful to quickly satisfy this particular request.

PR #133, opened 2020-02-25
Add MATLAB compiler and library version compatibility information

nickbattam-tessella:


hsautessella:
What's there looks fine to me. 

R2017b uses MSMPI 5.0.12435 and R2018b uses MSMPI 8.0.1.... (the same as R2019)

nickbattam-tessella:
Fixes #133

PR #144, opened 2020-03-05
Add list of disabled tests to _test folder

hsautessella:
This is a list of tests that I've been able to identify as disabled. 

Feel free to add to it.


nickbattam-tessella:
We need to agree the process for keeping this list up-to-date.

I propose that if a PR includes commits that enable or disable individual tests or test files these changes *must* be reflected in an update to this document made on the same branch.
It is the responsibility of PR reviews and the developer to ensure this happens.

As such it makes sense to create separate Herbert and Horace lists in the corresponding repo.

PR #147, opened 2020-03-06
Use Herbert CMake scripts in builds

### LINKED ISSUE #145: Make CMake helpers general and include in build package
https://github.com/pace-neutrons/Herbert/issues/95

hsautessella:
With https://github.com/pace-neutrons/Herbert/pull/96 merged, Horace can now use the CMake scripts in Herbert to perform its builds. This significantly reduces duplication between the two repositories.

Fixes #145


cmake/FindHerbert.cmake
```diff
@@ -37,6 +37,10 @@ set(DIRS_TO_SEARCH
     ENV{ProgramFiles}/Herbert/herbert_core
 )
 
+if(Herbert_ROOT AND NOT EXISTS "${Herbert_ROOT}/herbert_init.m")
```
nickbattam-tessella: what is the purpose of this check? The next command is a `find_path`

hsautessella: See commit message of https://github.com/pace-neutrons/Horace/pull/147/commits/678850436e6ca324ea2cfbd76b6279d6af224804. If `Herbert_ROOT` is passed to CMake on the command line, `find_path` won't do anything because `Herbert_ROOT` is already a cache variable.

hsautessella: Offline discussion: the `find_path` call is to be moved into an `else()` block

nickbattam-tessella: It would remove the ambiguity to move the following `find_path(Herbert_ROOT, ...` into an `else()` block.

If it's defined and valid - the `find_path` will execute and short-circuit
if it's defined and invalid - the new if() check will catch an raise a FATAL_ERROR
if it's undefined - the `find_path` will execute and (hopefully) find it.

hsautessella:
@nickbattam-tessella I'm going to merge this one, so I can get #148 through. Let me know if you spot any issues and I can deal with them in a new PR.

Issue #155, opened 2020-03-13
Re-enable test_gen_sqw_for_powders_tests


hsautessella:
**Outputs:**
- The `test_gen_sqw_for_powders` disabled in commit https://github.com/pace-neutrons/Horace/commit/bea56145a279b33188768a4f400fc4949dfa8b82 should be re-enabled.
- The tests should pass locally and on the build servers (regardless of config)

Part of #111

PR #156, opened 2020-03-13
Fix gen_sqw_for_powders tests

hsautessella:
Makes sure the `.par` file loaded during the test is loaded with the Matlab version of the ASCII loader. This ensures consistency with the reference file.

Fixes https://github.com/pace-neutrons/Horace/issues/155


_test/test_gen_sqw_for_powders/test_combine_cyl.m
```diff
@@ -68,30 +68,30 @@
...
             %add_to_files_cleanList(this,this.spe_file_1,this.spe_file_2);
             add_to_path_cleanList(this,test_functions_path);
-            
-            
+
+
         end
-        
+
         function this=test_combine_pow1(this)
```
nickbattam-tessella: why is this `test_combine_pow1` when it calls cylinder test helper? Should this be `test_combine_cyl1` to match the file name in line 78?

_test/test_gen_sqw_for_powders/test_combine_pow.m
```diff
@@ -30,31 +30,31 @@
...
-            
-            
+
+
+
             % =====================================================================================================================
             % Create spe files:
             this.par_file=fullfile(common_data_dir,'map_4to1_dec09.par');
-            %spe_dir = tmp_dir(); 
+            %spe_dir = tmp_dir();
```
nickbattam-tessella: is this dead code needed?

_test/test_gen_sqw_for_powders/gen_sqw_powder_test_helper.m
```diff
@@ -1,4 +1,4 @@
-function gen_sqw_powder_test (spe_file, par_file, sqw_file, efix, emode)
+function gen_sqw_powder_test_helper (spe_file, par_file, sqw_file, efix, emode)
```
nickbattam-tessella: we're not using the return values of this function (or those of `gen_sqw_cylinder_test_helper`)

Issue #128, opened 2020-02-17
Scripted test execution changes working directory


nickbattam-tessella:
Execution of the tests on Windows through the `build.ps1` changes directory from `Horace/` to `Horace/build`. 

The build step does not change the directory.
The Linux build and test steps leave the terminal in the initial directory.

The same issue is seen in Herbert.

nickbattam-tessella:
The `Invoke-Test` method begins with a call to `Push-Location`. 
Each test is subsequently executed with a call to `Invoke-In-Dir` which `Push-Location` -- `Pop-Location`
The initial `Push-Location` is never popped

PR #164, opened 2020-03-26
Remove unrequired directory push

nickbattam-tessella:
Minor update to build script removing unrequired directory change


nickbattam-tessella:
Fixes #128

Issue #136, opened 2020-02-28
Create Experiment class


nickbattam-tessella:
Create new Experiment class wrapping Sample, Instrument and Detector information

Part of #21

PR #176, opened 2020-04-07
Experiment class

nickbattam-tessella:
Initial implementation of SQW Experiment class.

Requires https://github.com/pace-neutrons/Herbert/pull/125
Part of https://github.com/pace-neutrons/Horace/issues/136


nickbattam-tessella:
Load/Save of .mat manually tested

nickbattam-tessella:
Tests will fail until https://github.com/pace-neutrons/Herbert/pull/125 is merged

nickbattam-tessella:
`init` method can be added when it's needed

nickbattam-tessella:
Failing tests are unrelated to this PR.

nickbattam-tessella:
> `set.property(obj)` should be without underscore `set.property_(obj)`. When user see object, he see `instance.property = something`

There are `set.property(obj, value)` and `set.property_(obj, value)` methods. The `obj.property = value` use-case is tested.

PR #197, opened 2020-04-15
Re-enable test_tobyfit_1

### LINKED ISSUE #186: test_tobyfit_1 is failing
This test passed on the PR that re-enabled the running of the TobyFit tests, but seems to be consistently failing now.

The test consistently passes when I run it locally.

**Outputs:**
- Fix and re-enable the test so that it passes consistently on the build servers

hsautessella:
Re-enables the previously disabled `test_tobyfit_1` test. I could not re-create any failures locally, although there were a load of failures on the build server. I've re-enabled the test and added a seed to the top of the test script. This should allow us to re-create any failures we see on the build servers and implement the appropriate fix.

Fixes #186


hsautessella:
Test failure was a different test `test_tobyfit_let_1`

PR #210, opened 2020-04-28
Update build pipelines to use Herbert sources

### LINKED ISSUE #196: Consider/Update PR build pipeline to use Herbert sources
Horace is built against the build artifacts created by the last successful build of Herbert master (typically the nightly).

This means there is a long lag between any issues arising from changes introduced by Herbert PRs being seen in Horace.

This issue it to consider, and potentially implement, a workflow change to the build system so that Horace PRs are built/tested against the current Herbert master *source*. 
Branch and master builds would continue to be build and tests against the successful Herbert master builds.

Consequences:
- Horace PRs will fail more quickly if there are bugs introduced
- Horace is not build built against a "published" build artifact
- Horace builds will include an extra build (but not test) Herbert step which must be maintained
- As now, any changes to Herbert master will not trigger the PR to rebuild.

hsautessella:
This PR moves us to using and building Herbert sources instead of using Jenkins archived master builds.

- PR and branch builds now use/build Herbert master source
- Nightly builds still use the latest successful Herbert nightly package

I've added `branch` options to the Herbert clone and build scripts so it should be relatively straight forward to add a `HERBERT_BRANCH` parameter on Jenkins, so we can build against an arbitrary Herbert branch e.g. some release branch

Fixes #196


tools/build_config/Jenkinsfile
```diff
@@ -73,25 +73,47 @@ pipeline {
...
+                    "--cmake_flags \"-DHorace_RELEASE_TYPE=\$RELEASE_TYPE\" \
+                    --matlab_release \$MATLAB_VERSION \
+                    --build_tests OFF
+              '''
+            } else {
+              powershell '''
+                ./tools/build_config/clone_herbert_branch.ps1 \
+                    -cmake_flags ""-DHerbert_RELEASE_TYPE=\$env:RELEASE_TYPE"" \
+                    -matlab_release \$env:MATLAB_VERSION \
+                    -build_tests OFF
```
nickbattam-tessella, line 110: there's no need to pass the `-build_tests OFF` argument -- we will never run Herbert tests from the Horace pipeline so that can be added directly to the build commands in the scripts below before the vararg piece.

tools/build_config/clone_herbert_branch.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
nickbattam-tessella, line 1: unlike the powershell script, this one will leave the current directory as `[...]/Herbert`

tools/build_config/Jenkinsfile
```diff
@@ -73,25 +73,47 @@ pipeline {
       }
     }
 
-    stage('Clone-Herbert') {
+    stage('Copy-Herbert') {
```
nickbattam-tessella, line 76: `Get-Herbert`? It's now either cloning or copying.

tools/build_config/clone_herbert_branch.ps1
```diff
@@ -0,0 +1,26 @@
...
+
+param(
+  [string][Alias("B")]$branch = "master",
+  [string]$build_args = ""
+)
+
+. $PSScriptRoot/powershell_helpers.ps1 <# Imports:
+  Write-And-Invoke
+  Invoke-In-Dir
+#>
```
nickbattam-tessella, line 14: that's confusing comment indenting; maybe would make more sense to thave the Imports Write+Invoke and Invoke-In-Dir comment abouve the ./powershell_helpers.ps1 line?

tools/build_config/clone_herbert_branch.ps1
```diff
@@ -0,0 +1,26 @@
...
+
+$HERBERT_URL = "https://github.com/pace-neutrons/Herbert.git"
+
+Write-Output "Cloning and building Herbert branch $branch..."
```
nickbattam-tessella, line 18: should this be a different Write in the two branches: it's not cloning if there's a Herbert folder already there

tools/build_config/clone_herbert_branch.ps1
```diff
@@ -0,0 +1,26 @@
+# Clone and build the given Herbert branch. If no branch is given, default to
+# master
+# Pass a specific branch using the -branch parameter, any other arguments are
+# passed verbatim to the Herbert build script
+
+param(
+  [string][Alias("B")]$branch = "master",
```
nickbattam-tessella, line 7: your API is inconsistent internally and between the bash and powershell scripts;
- the bash script doesn't accept short-args 
- you don't have a short form of the second argument.

tools/build_config/clone_herbert_branch.sh
nickbattam-tessella: is there any reason we don't have an "invoke-in-dir" equivalent function for all the bash scripting?

tools/build_config/clone_herbert_branch.ps1
hsautessella: This print statement is more about telling you want branch is being built. I've removed the cloning bit. What action is actually being performed is made obvious by the commands being printed before they're executed

tools/build_config/clone_herbert_branch.ps1
hsautessella: I've just removed the short argument. It's not used and this is not a public facing script

tools/pwsh/clone_herbert_branch.ps1
```diff
@@ -0,0 +1,26 @@
...
+<# Import:
+    Write-And-Invoke
+    Invoke-In-Dir #>
+. $PSScriptRoot/powershell_helpers.ps1
+
+$HERBERT_URL = "https://github.com/pace-neutrons/Herbert.git"
+
+Write-Output "Building Herbert branch $branch..."
+if (Test-Path -Path "./Herbert") {
+  Invoke-In-Dir -directory "Herbert" -command "git fetch origin"
```
nickbattam-tessella, line 20: Should the "Herbert" directory name be a variable as it's used in four places?

tools/pwsh/clone_herbert_branch.ps1
```diff
@@ -14,13 +14,14 @@ param(
 . $PSScriptRoot/powershell_helpers.ps1
 
 $HERBERT_URL = "https://github.com/pace-neutrons/Herbert.git"
+$HERBERT_DIR = "Herbert"
 
 Write-Output "Building Herbert branch $branch..."
-if (Test-Path -Path "./Herbert") {
-  Invoke-In-Dir -directory "Herbert" -command "git fetch origin"
-  Invoke-In-Dir -directory "Herbert" -command "git checkout origin/$branch"
+if (Test-Path -Path "$HERBERT_DIR") {
```
nickbattam-tessella, line 20: is the change from `-Path "./Herbert"` to `-Path "Herbert"` safe?

hsautessella: Should be, it's still looking in the working directory

tools/pwsh/clone_herbert_branch.ps1
```diff
@@ -16,12 +16,14 @@ param(
...
 if (Test-Path -Path "$HERBERT_DIR") {
   Write-And-Invoke "git -C $HERBERT_DIR fetch origin"
   Write-And-Invoke "git -C $HERBERT_DIR reset --hard origin/$branch"
 } else {
   Write-And-Invoke "git clone $HERBERT_URL --depth 1 --branch $branch $HERBERT_DIR"
 }
-$build_cmd = "$HERBERT_DIR/tools/build_config/build.ps1 -build -build_tests OFF $build_args"
+
+$build_cmd = "$HERBERT_DIR/tools/build_config/build.ps1 -build"
+$build_cmd += " -build_tests OFF $build_args"
```
nickbattam-tessella, line 28: it would be clearer to split this over two lines; make the `build_test OFF` explicit

CMakeLists.txt
```diff
@@ -73,5 +78,6 @@ install(
     DESTINATION "Herbert"
     USE_SOURCE_PERMISSIONS
     PATTERN "*.m~" EXCLUDE
-    PATTERN "*.asv" EXCLUDE
+    PATTERN "*.asv ${EXTERNAL_IGNORE_PATTERN}" EXCLUDE
```
nickbattam-tessella, line 81: accidental pattern duplication

PR #212, opened 2020-04-29
Powder projections

abuts:
simple move of projection scripts to algorithm directory and updated powder tests.


_test/common_functions/simulate_spe_testfunc.m
```diff
@@ -32,7 +32,11 @@
 [~,seed_key] = fileparts(spe_file);
 sk = regexp(seed_key,'(\d+)','tokens');
 sk = [sk{:}];
-seed_key =['key_', [sk{:}]];
+if isempty(sk)
```
nickbattam-tessella, line 35: How does this random number seed management work alongside the Herbert `_test\shared\seed_rng` (available to Herbert and Horace tests) @hsautessella ?

horace_core/algorithms/gen_sqw_cylinder.m
```diff
@@ -1,5 +1,10 @@
-function gen_sqw_cylinder_test_helper (spe_file, par_file, sqw_file, efix, varargin)
+function gen_sqw_cylinder(spe_file, par_file, sqw_file, efix, varargin)
 % Read one or more spe files and a detector parameter file, and create an output sqw file.
+% 
+% WARNING: The sqw objects and files, produced by this routine are not
```
nickbattam-tessella, line 4: this warning isn't clear -- what does "not fully functional" mean? Is that what's covered by the `*** IMPORTANCE NOTES ***` below.

Should the 
```
	% *** TEST ROUTINE
%       This was created rapidly as a fix-up during an experiment. A polished version is	%       This was created rapidly as a fix-up during an experiment. A polished version is
%       marked for addition at a later date.	%       marked for addition at a later date.
```
comment be deleted? This isn't a test routine as more as it's moved into the main codebase and won't be used for testing; it's a partial implementation of the functionality with limitations noted in the header and reinforced by the `WARNING`

nickbattam-tessella: (same comment for `gen_sqw_powder`)

_test/common_functions/simulate_spe_testfunc.m
hsautessella: Doesn't look like this seed ever touches Matlab's `rand` functions. It's only used in `rand_like` (defined in `Herbert/herbert_core/utilities/maths`).

Also, `seed_rng` lives in `test_tobyfit` at the moment. We need to sort out an `init_tests` setting in Horace, so we can add the right directories to the Matlab path when we're running Horace tests

abuts: I think these two are completely independent. I am using one, deployed in gen_sqw

horace_core/algorithms/gen_sqw_cylinder.m
abuts: I have no idea and no desire to find out -- our plan is to fully redesign this code. The advice is to look at the tests and see what is working.

abuts: can be a bit rephrased -- yes.

PR #217, opened 2020-05-01
Update Jenkins SMG with Herbert dependency plan

hsautessella:
Added a section the Jenkins SMG explaining how each pipeline type will deal with Horace's Herbert dependency now that #210 has been merged


hsautessella:
Since this is just a documentation change, I've cancelled the builds to keep the strain off the Windows machine

PR #231, opened 2020-05-20
mpi documentation

abuts:
just playing with tests and trying to complete all unit/system tests. Checking if some parallel combine needs to be disabled on windows.

Depends on 
https://github.com/pace-neutrons/Herbert/pull/173


admin/CMakeLists.txt
```diff
@@ -31,9 +32,15 @@ string(REGEX REPLACE
...
+#
+if(${BUILD_TESTS})
+    configure_file("worker_4tests.m.template" "${LOCAL_INIT_DIR}/worker_4tests.m")
+endif()
+configure_file("worker_v2.m.template" "${LOCAL_INIT_DIR}/worker_v2.m")
+
 
 # Copy Herbert's worker template
-configure_file("${Herbert_WORKER_TEMPLATE}" "${LOCAL_INIT_DIR}/worker_v2.m")
+#configure_file("${Herbert_WORKER_TEMPLATE}" "${LOCAL_INIT_DIR}/worker_v2.m")
```
nickbattam-tessella, line 43: this line and the associated comment can be deleted rather than commented out they're not being used.

horace_core/configuration/@hpc_config/hpc_config.m
```diff
@@ -219,6 +219,8 @@
             end
             if use_mpi
                 pc = parallel_config;
+                pc.saveable = false;
```
nickbattam-tessella, line 222: is this `pc.saveable=false` required? you needed to remove it elsewhere to propagate configuration to the workers.

admin/CMakeLists.txt
hsautessella: Also, if we're not using the `Herbert_WORKER_TEMPLATE` variable any more, we no longer need to find it in `FindHerbert`. So the block here: https://github.com/pace-neutrons/Horace/blob/c395b5344074c962b548bb114e565eb88c8bb990/cmake/FindHerbert.cmake#L56-L61 can be removed

horace_core/configuration/@hpc_config/hpc_config.m
abuts: its safe to do it in local tests which change configuration but not deploy parallel job.

nickbattam-tessella: This is a global change which will affect tests running MPI jobs too

admin/CMakeLists.txt
abuts: done

PR #235, opened 2020-05-21
Add "How to use the SAN area" SMG

### LINKED ISSUE #232: SAN + Jenkins - Proof of concept
**Output**:

- A Jenkinsfile script that can mount the SAN area and copy across a file, whilst running on the ANVIL sl7 nodes

hsautessella:
This PR adds an SMG entry for how to use the SAN area. A short example has been given on how to mount the drive on ANVIL. 

As we develop techniques for dealing with SAN data (i.e. automated data downloading for tests) more should be added to the SMG.

This PR is only meant to address the associated issue.

Fixes #232


documentation/smg/10_how_to_use_the_SAN_area.md
```diff
@@ -0,0 +1,35 @@
...
+that are too large to store in git repositories.
+
+A FedID is required in order to mount the SAN area. In order to mount the SAN
+on PACE's systems, a dedicated user ID with which to mount the SAN area has
+been provided to PACE.
+
+## Mounting the SAN on Jenkins
+
+To mount the SAN when running a Jenkins pipeline, you should use the dedicated
+user ID provided to PACE. The credentials for this are stored in a credentials
```
nickbattam-tessella, line 15: I would be explicit here on the format of that file so anyone looking at it knows exactly what they're seeing 
```
username\n
domain\n
password\n
```
If we were using the username/password facility of the Jenkins credentials plugin that would give us `MY_ID` (`username:password`), `MY_ID_USR` (`username`) and `MY_ID_PSW` (`password`) so everything is explicit.

nickbattam-tessella: `stored in a credentials file` possible needs a little more explanation: 
- this is a credentials file stored in the secure Jenkins Credentials Binding plugin
- https://www.jenkins.io/doc/book/using/using-credentials/

documentation/smg/10_how_to_use_the_SAN_area.md
```diff
@@ -0,0 +1,35 @@
...
+gives an example of how to copy the file `README.txt` from the SAN area in a
+Jenkinsfile. ANVIL requires the use of `dbus-run-session` when mounting the
+drive.
+
+```groovy
+stage('SAN-Mount-And-Copy') {
+  steps {
+    withCredentials([file(credentialsId: 'SAN_credentials_file', variable: 'san_credentials')]) {
+      sh '''
+        echo "gio --version; gio mount smb://isis.cclrc.ac.uk/Shares/PACE_Project_Tool_Source/ < ${san_credentials}; gio copy smb://isis.cclrc.ac.uk/Shares/PACE_Project_Tool_Source/README.txt . -p" > mount_script.sh
```
nickbattam-tessella, line 28: Should the share location be a Jenkinsfile parameter? (stored in the Jenkins job configuration or keystore)? 
- it's used twice in this mini script, probably more times in real code
- it could change
- it probably shouldn't be stored in GitHub

documentation/smg/10_how_to_use_the_SAN_area.md
```diff
@@ -0,0 +1,54 @@
...
+A FedID is required in order to mount the SAN area. In order to mount the SAN
+on PACE's systems, a dedicated user ID with which to mount the SAN area has
+been provided to PACE.
+
+## Mounting the SAN on Jenkins
+
+The path to the SAN is stored as a secret in Jenkins with the ID `SAN_path`.
+To mount the SAN when running a Jenkins pipeline, you should use the dedicated
+user ID provided to PACE. The credentials for this are stored in a credentials file in Jenkins the with ID `SAN_credentials_file`,
+see https://www.jenkins.io/doc/book/using/using-credentials/ for an overview
```
nickbattam-tessella, line 17: Now the path and credentials are stored as jenkins secrets, `see https://www.jenkins.io/doc/book/using/using-credentials/` doesn't fit here as includes an overview of both the credentials file and the secret path.

It makes more sense to move this link after the description of the file format or the the start. That way the section becomes more structured as:
- everything is stored in jenkins secrets
- here a link to how to use jenkins secrets 
- the path is < here >
- the credentials are < here > in < this > format

documentation/smg/10_how_to_use_the_SAN_area.md
```diff
@@ -0,0 +1,54 @@
...
+
+```txt
+username
+domain
+password
+```
+
+One way to mount the drive on Linux is to use `gio`. The code fragment below
+gives an example of how to copy the file `README.txt` from the SAN area in a
+Jenkinsfile. ANVIL requires the use of `dbus-run-session` when mounting the
```
nickbattam-tessella, line 29: "from the SAN area in a Jenkinsfile [using a generated script].". 

Half of the code is here is generating the script.

documentation/smg/10_how_to_use_the_SAN_area.md
```diff
@@ -0,0 +1,54 @@
...
+(alongside scripts recording how the smaller sets were generated) and files
+that are too large to store in git repositories.
+
+A FedID is required in order to mount the SAN area. In order to mount the SAN
+on PACE's systems, a dedicated user ID with which to mount the SAN area has
+been provided to PACE.
+
+## Mounting the SAN on Jenkins
+
+The path to the SAN is stored as a secret in Jenkins with the ID `SAN_path`.
```
nickbattam-tessella, line 14: have you created the `SAN_path` key?

documentation/smg/10_how_to_use_the_SAN_area.md
hsautessella: This felt like the most succinct way to give an example. In reality, we'd have an independent script that contains all the `gio` commands, then we'd call `dbus-session -- sh ` on that. This was only supposed to document a proof-of-concept. 

I'll add a note that you may want to have a separate script that does the mount/copy, rather than generating the script in the Jenkinsfile

documentation/smg/10_how_to_use_the_SAN_area.md
hsautessella: Yes

PR #262, opened 2020-06-11
Create data storage ADRs

### LINKED ISSUE #218: Resolve mix of float and double data in SQW
1a) The SQW file stores pixel data as single-precision float data
1b) The SQW file stores image and all other data as double-precision
2) The C++ code holds and processes pixel data in a float vector
3) MATLAB reads this data into a double array (which doubles the size of the object in memory)
4) Modelling and fitting codes use double-precision coefficients

Float data use reduces the memory and storage requirements and reflect the reality of machine resolution (~3sf).
It is possible, and will be increasingly likely that, data sets will be too large to fit in memory. Custom paging is already in use to only load the needed pieces of the data array (i.e. there is no need to hold the full array in memory).

Interfaces with external systems (Brille, Euphonic, SpinW) will have an expectation on the data types which we must meet (?what is that).

Mixing float and double in calculation will dynamically truncate the double data to a single precision float value (even if that reduces it to float(infinity)

Questions:
- [ ] should we store pixel data as double-precision
- [ ] should we store image data as double-precision
- [ ] should we store model parameters as double-precision
- [ ] should we allow the data type (single/double) to be configurable

nickbattam-tessella:
Document decisions reached discussing #218 

Fixes #218

### DIFF
--- a/documentation/adr/0016-use-double-array-in-memory-for-pixel-data.md
+++ b/documentation/adr/0016-use-double-array-in-memory-for-pixel-data.md
@@ -38,8 +38,9 @@ Access to the data through the PixelData class API will return `double` values.
 
 ## Consequences
 
-- Individual functions that use single precision data for performance optimization (e.g. sorting) will be required to convert data on input and output
-- Rounding errors accumulated during calculations will be 
+- Individual functions that use single precision data for performance optimization (e.g. sorting) will be required to convert data on input and output.
+- Rounding errors accumulated during calculations will typically remain smaller than the signal underlying signal.
+- Calculated data values will be truncated when written to file.
 - The recomputation of pixel coordinates by different routines in `TobyFit` and `gen_sqw` means rounding errors may result in pixels to be binned differently -- these differences should be considered and if possible eliminated.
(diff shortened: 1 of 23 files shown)


documentation/adr/0016-use-double-array-in-memory-for-pixel-data.md
```diff
@@ -0,0 +1,45 @@
...
+Pixel data will stored internally as an array of MATLAB doubles and converted on read or write. 
+
+Access to the data through the PixelData class API will return `double` values.
+
+ 
+
+## Consequences
+
+- Individual functions that use single precision data for performance optimization (e.g. sorting) will be required to convert data on input and output
+- Rounding errors accumulated during calculations will be
```
g5t, line 42:
```
- Rounding errors accumulated during calculations will be ...
```
Accumulated rounding errors will be ignored? Implicitly truncated when writing to disk?

nickbattam-tessella: ... typically smaller than the signal, addressing Toby's concern that for single precision data they could accumulate to be of a similar order.

PR #263, opened 2020-06-15
Add file-backed read access to PixelData API

### LINKED ISSUE #246: PixelData - Add file-backed read access to API
Add support for paged/partial file reads to the core PixelData object

Part of #224

Output
- [ ] Update PixelData to support file backed array read
- [ ] Create consistent API for handling of large and small data
- [ ] Extend Getter APIs to support return values too large to fit in memory

hsautessella:
This PR adds file-backed data access to the `PixelData` class. 

- `PixelData` can now take a file path or `faccess` object as a constructor argument. 
- A "page" of data that fills a given memory allocation is read into memory _when a getter is called_, no data is loaded on construction
- The getters now pull only from what is in the current "page" of data
- An `advance` method will read in the next "page" of data
- `PixelData` is now a handle class
    - this means `sqw` and `PixelData` objects must be explicitly copied (i.e. with the `copy()` function) if they are to be independent
- The new `PixelData.has_more()` method will return true if there is a subsequent page of data in the file that can be loaded

### Follow-up tasks
Issues should be raised for the following:
- Update algorithm usages of PixelData so that they iterate over PixelData with a `while pix.has_more()` loop
- Add methods `get_data_at_index(fields, indices)` and `get_pix_at_index(indices)` that will pull data from the file according to real pixel index, rather than pixel index in the current page #270

Fixes #246

### DIFF
--- a/_test/test_sqw/test_PixelData.m
+++ b/_test/test_sqw/test_PixelData.m
@@ -386,9 +386,11 @@ methods
     function test_page_size_is_set_on_construction_when_given_as_argument(obj)
         mem_alloc = obj.small_page_size_;  % 1Mb
         expected_page_size = floor(mem_alloc/(8*9));  % mem_alloc/(double*num_rows)
-        assertEqual(obj.pix_data_small_page.page_size, expected_page_size);
+        % the first page is loaded on access, so this first assert which accesses
+        % .variance is necessary to set pix.page_size
         assertEqual(size(obj.pix_data_small_page.variance), ...
                     [1, obj.pix_data_small_page.page_size]);
+        assertEqual(obj.pix_data_small_page.page_size, expected_page_size);
     end
 
     function test_calling_getter_returns_data_for_single_page(~)
--- a/_test/test_sqw/test_PixelData.m
+++ b/_test/test_sqw/test_PixelData.m
@@ -346,6 +346,8 @@ methods
     function test_construction_with_file_path_sets_num_pixels_in_file(obj)
         f_accessor = sqw_formats_factory.instance().get_loader(...
                 obj.test_sqw_file_path);
+        assertEqual(obj.pix_data_from_file.num_pixels, f_accessor.npixels);
+    end
 
     function test_construction_with_file_path_sets_size(obj)
         f_accessor = sqw_formats_factory.instance().get_loader(...
--- a/_test/test_sqw/test_PixelData.m
+++ b/_test/test_sqw/test_PixelData.m
@@ -2,7 +2,10 @@ classdef test_PixelData < TestCase
 
 properties
     raw_pix_data = rand(9, 10);
+    test_sqw_file_path = '../test_sqw_file/sqw_1d_1.sqw';
+
     pixel_data_obj;
+    pix_data_from_file;
 end
 
 methods (Access = private)
@@ -19,7 +22,10 @@ methods
     function obj = test_PixelData(~)
         obj = obj@TestCase('test_PixelData');
 
+        % Construct an object from raw data
         obj.pixel_data_obj = PixelData(obj.raw_pix_data);
+        % Construct an object from a file
+        obj.pix_data_from_file = PixelData(obj.test_sqw_file_path);
     end
 
     function test_default_construction_sets_empty_pixel_data(~)
@@ -318,11 +324,30 @@ methods
         assertExceptionThrown(f, 'PIXELDATA:data');
     end
 
-    function test_construction_with_file_path_sets_file_path_on_object(~)
-        file_path = '../test_sqw_file/sqw_1d_1.sqw';
-        pix_data = PixelData(file_path);
-        assertEqual(pix_data.file_path, file_path);
+    function test_construction_with_file_path_sets_file_path_on_object(obj)
+        assertEqual(obj.pix_data_from_file.file_path, obj.test_sqw_file_path);
     end
+
+    function test_construction_with_file_path_populates_data_from_file(obj)
+        assertFalse(isempty(obj.pix_data_from_file));
+        expected_signal_chunk = [0, 0, 0, 0, 0483.5, 4463.0, 1543.0, 0, 0, 0];
+        assertEqual(obj.pix_data_from_file.signal(9825:9834), ...
+                    expected_signal_chunk);
+    end
+
+    function test_construction_with_file_path_sets_num_pixels_in_file(obj)
+        f_accessor = sqw_formats_factory.instance().get_loader(...
+                obj.test_sqw_file_path);
+        num_pix_in_file = f_accessor.npixels;
+        assertEqual(obj.pix_data_from_file.num_pixels, num_pix_in_file);
+    end
+
+    function test_error_on_construction_with_non_existent_file(~)
+        file_path = 'not-a-file';
+        f = @() PixelData(file_path);
+        assertExceptionThrown(f, 'SQW_FILE_IO:runtime_error');
+    end
+
 end
 
 end
--- a/_test/test_sqw/test_PixelData.m
+++ b/_test/test_sqw/test_PixelData.m
@@ -2,11 +2,13 @@ classdef test_PixelData < TestCase
 
 properties
     raw_pix_data = rand(9, 10);
+    small_page_size_ = 1e6;  % 1Mb
     test_sqw_file_path = '../test_sqw_file/sqw_1d_1.sqw';
 
     pixel_data_obj;
     pix_data_from_file;
     pix_data_from_faccess;
+    pix_data_small_page;
 end
 
 methods (Access = private)
@@ -30,6 +32,8 @@ methods
         % Construct an object from a file accessor
         f_accessor = sqw_formats_factory.instance().get_loader(obj.test_sqw_file_path);
         obj.pix_data_from_faccess = PixelData(f_accessor);
+        % Construct an object from file accessor with small page size
+        obj.pix_data_small_page = PixelData(f_accessor, obj.small_page_size_);
     end
 
     % --- Tests for in-memory operations ---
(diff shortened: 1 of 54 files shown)


horace_core/sqw/@PixelData/PixelData.m
```diff
@@ -327,41 +440,110 @@
...
+end
+
+methods (Access = private)
+
+    function obj = init_from_file_accessor_(obj, f_accessor)
+        % Initialise a PixelData object from a file accessor
+        obj.f_accessor_ = f_accessor;
+        obj.file_path_ = fullfile(obj.f_accessor_.filepath, ...
+                                  obj.f_accessor_.filename);
+        pg_size = obj.page_size;
```
nickbattam-tessella, line 518: unnecessary local variable `pg_size`

nickbattam-tessella: Why only load the first page if it hold all the data?

horace_core/sqw/@PixelData/PixelData.m
```diff
@@ -327,41 +440,110 @@
...
+
+    function page_size = get.page_size(obj)
+        % The number of pixels that are held in the current page. If no pixels
+        % are held, this returns the number of pixels that will be loaded into
+        % the next page.
+        %
+        % If "num_pixels in file" < "number of pixels that can fit in memory size",
+        % return num_pixels - this avoids ever reading past end of file
+        if isempty(obj.data)
+            num_bytes_in_val = 8;  % pixel data stored in memory as a double
```
nickbattam-tessella, line 498: would it make sense for this (constant) be defined where the pixel array cache is, or to calculate it from the array (`whos('obj.data').bytes / numel(obj.data)`)?

horace_core/sqw/@PixelData/PixelData.m
```diff
@@ -327,41 +440,110 @@
...
+            num_pix = obj.f_accessor_.npixels;
+        end
+    end
+
+    function file_path = get.file_path(obj)
+        file_path = obj.file_path_;
+    end
+
+    function page_size = get.page_size(obj)
+        % The number of pixels that are held in the current page. If no pixels
```
nickbattam-tessella, line 491: this method/comment are slightly confusing:
- returns the (max) number of pixels that can be in a page (derived from the allocated memory size)
OR
- the number of pixels that ARE in the page

horace_core/sqw/@PixelData/PixelData.m
```diff
@@ -248,9 +309,54 @@
...
+        %
+        %    >> has_more = pix.has_more();
+        %
+        has_more = false;
+        if ~isempty(obj.f_accessor_)
+            has_more = obj.pix_position_ + obj.page_size - 1 < obj.num_pixels;
+        end
+    end
+
+    function advance(obj)
```
nickbattam-tessella, line 328: Nice example

horace_core/sqw/@PixelData/PixelData.m
```diff
@@ -327,41 +440,110 @@
...
+                                  obj.f_accessor_.filename);
+        pg_size = obj.page_size;
+        if obj.num_pixels <= pg_size
+            obj.data = obj.f_accessor_.get_pix(1, pg_size);
+            obj.pix_position_ = obj.num_pixels;
+        end
     end
 
+    function obj = load_page_(obj, pix_idx_start)
+        if pix_idx_start >= obj.num_pixels
```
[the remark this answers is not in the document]
hsautessella: Calling `.advance` won't work because as far as the API is concerned the first page is already there, so calling `.advance` will give you the second page. But I see your point, a user should never see a `PIXELDATA:load_page_`. Although that should never be thrown within `load_first_page_if_data_empty_` unless the SQW file has no pixels - is this possible?

horace_core/sqw/@PixelData/PixelData.m
hsautessella: `obj.page_size` is a dependent variable, so having the local variable saved a function call.

If all the data can be held in one page we may as well load that in. If the data needs multiple pages, we hold off on loading until we know which page needs to be loaded. This will be most useful when we implement `get_data_at_pix_index`

horace_core/sqw/@PixelData/PixelData.m
hsautessella: Yes, perhaps these should be two separate properties `page_size` and `max_page_size`

_test/test_sqw/test_PixelData.m
```diff
@@ -311,9 +333,227 @@ function test_construction_with_float_raises_PIXELDATA_error(~)
...
+        mem_alloc = 11*8*9;  % 11 pixels held in memory at a time
+        pix = PixelData(faccess, mem_alloc);
+        assertEqual(size(pix.signal), [1, pix.page_size]);
+        assertEqual(pix.signal, data(8, 1:11));
+    end
+
+    function test_calling_get_data_returns_data_for_single_page(~)
+        data = rand(9, 30);
+        faccess = FakeFAccess(data);
+        mem_alloc = 11*8*9;
```
nickbattam-tessella, line 408: what are these magic numbers: `8`, `9`, `11`?

_test/test_sqw_file/test_faccess_sqw_prototype.m
```diff
@@ -23,104 +23,104 @@
...
             assertTrue(isa(data_dnd,'data_sqw_dnd'));
             assertEqual(data_dnd.filename,'test_sqw_read_write_v0_t.sqw');
-            
-            data = to.get_data('-ver',1,10);
+
+            data = to.get_data('-ver');
             assertEqual(data.filename,data_dnd.filename)
             assertEqual(data.filepath,data_dnd.filepath)
-            assertEqual(size(data.pix),[9,10]);
+            assertTrue(isa(data.pix, 'PixelData'));
```
nickbattam-tessella, line 121: should this PixelData class contain any data? 
Would an additional `data.pix.num_pixels` tests be useful?

horace_core/sqw/file_io/@dnd_binfile_common/dnd_binfile_common.m
```diff
@@ -455,10 +455,35 @@ function check_obj_initated_properly(obj)
             sqw_obj  = obj.sqw_holder_;
         end
         %
-        function is = is_activated(obj)
-            % Check if the file-accessor is bind with open binary file
+        function is = is_activated(obj, read_or_write)
```
nickbattam-tessella, line 458: should this argument be `mode` and take "single character" `r`, `w`, `rw` alternatives to full words?

horace_core/sqw/file_io/@dnd_binfile_common/dnd_binfile_common.m
```diff
@@ -455,10 +455,35 @@ function check_obj_initated_properly(obj)
...
+            %                 for reading. If 'write' return true if file is
+            %                 open for writing.
+            %
+            full_file_path = fullfile(obj.filepath, obj.filename);
+            [file_id_path, permission] = fopen(obj.file_id_);
+            is = ~isempty(obj.file_closer_) && strcmp(full_file_path, file_id_path);
+
+            if is && nargin == 2
+                if strcmp(read_or_write, 'read')
+                    read_mode_regex = '([ra]b\+?)|(wb\+)';
```
nickbattam-tessella, line 474: the `read_mode_regex` and `write_mode_regex` are constants.

horace_core/sqw/file_io/@dnd_binfile_common/dnd_binfile_common.m
```diff
@@ -474,18 +499,21 @@ function check_obj_initated_properly(obj)
             obj.file_id_ = 0;
         end
         %
-        function obj = activate(obj)
+        function obj = activate(obj, permission)
```
nickbattam-tessella, line 502: no documentation for the `permission` argument... what is that? what is the default?

nickbattam-tessella: if we're only ever opening in a finite set of modes would adding those string to the class as well named constants be helpful so you could call `.activate(READ_ONLY)` ?

horace_core/sqw/file_io/@dnd_binfile_common/dnd_binfile_common.m
```diff
@@ -455,10 +455,35 @@ function check_obj_initated_properly(obj)
...
+            % read_or_write   Char array. If 'read' return true if file is open
+            %                 for reading. If 'write' return true if file is
+            %                 open for writing.
+            %
+            full_file_path = fullfile(obj.filepath, obj.filename);
+            [file_id_path, permission] = fopen(obj.file_id_);
+            is = ~isempty(obj.file_closer_) && strcmp(full_file_path, file_id_path);
+
+            if is && nargin == 2
+                if strcmp(read_or_write, 'read')
```
nickbattam-tessella, line 473: these comparisons should be `strcmpi`

horace_core/sqw/file_io/@dnd_binfile_common/dnd_binfile_common.m
```diff
@@ -474,18 +499,32 @@ function check_obj_initated_properly(obj)
...
+            %
+            % Input
+            % -----
+            %
+            % read_or_write   Char array. If 'read' open the file in read-only
+            %                 mode If 'write' open the file in read/write mode.
+            %                 Default is 'read'.
+            %
+            if nargin == 1
+                permission = 'rb';
```
nickbattam-tessella, line 518: you have `rb` in two places now; this should be a call to `get_fopen_permission_("read")` or adapt that function to handle the default case

horace_core/sqw/file_io/@dnd_binfile_common/private/get_fopen_permission_.m
```diff
@@ -0,0 +1,12 @@
+function permission = get_fopen_permission_(permission)
```
nickbattam-tessella, line 1: no tests for this function

horace_core/sqw/file_io/@sqw_binfile_common/get_pix.m
```diff
@@ -10,6 +10,10 @@
         'get_pix method called from un-initialized loader')
 end
 
+if ~obj.is_activated('read')
+    obj = obj.activate('rb+');
```
nickbattam-tessella: this should be updated to use the same "read" => permission mapping function

horace_core/sqw/file_io/@sqw_binfile_common/put_pix.m
```diff
@@ -22,6 +22,9 @@
         'SQW_BINFILE_COMMON::put_pix: %s',mess);
 end
 
+if ~obj.is_activated('write')
+    obj = obj.activate('rb+');
```
nickbattam-tessella: this should be updated to use the same "write" => permission mapping function

horace_core/sqw/file_io/@dnd_binfile_common/private/get_fopen_permission_.m
hsautessella: This is a private function, so I've tested at a higher level in https://github.com/pace-neutrons/Horace/pull/263/commits/f2d3d0806fa3911bb22859111241eac0b70de43b

nickbattam-tessella: You've tested the get-permission / is_activated loop in those tests - which are good tests - but not that read permission `== rb,` and write permission `== rb+`.

That's probably sufficient for now as testing at the lower level would need a big restructure.

hsautessella: I've stuck a check in for `rb` and `rb+`, I didn't put it in originally because it's slightly messy to do, but it's definitely worth having

PR #273, opened 2020-06-30
Always return PixelData.num_pixels as double

### LINKED ISSUE #272: PixelData.num_pixels sometimes returns uint64
This is causing issues when using `.num_pixels` in calculations. This only happens when the data is file-backed, as the PixelData instance pulls the number from `faccess` which stores it as `uint64`.

Example:
```matlab
signal_sum = sum(fe_sqw.data.pix.get_data('signal'));
while fe_sqw.data.pix.has_more()
    fe_sqw.data.pix.advance();
    signal_sum = signal_sum +  sum(fe_sqw.data.pix.get_data('signal'));
end
mean_signal = signal_sum/fe_sqw.data.pix.num_pixels
```
prints:
> mean_signal = 
    <ins>unit64</ins>
    0

**Outputs:**:
- `PixelData.num_pixels` should always return a double

hsautessella:
This PR ensures that `PixelData.num_pixels` always returns a double. See the corresponding issue (#272) for bug description and reasoning behind the decision.

Fixes #272

### DIFF
--- a/_test/test_sqw/test_PixelData.m
+++ b/_test/test_sqw/test_PixelData.m
@@ -545,6 +545,16 @@ methods
         assertTrue(new_pix.has_more());
     end
 
+    function test_num_pixels_is_a_double_if_faccess_returns_uint(obj)
+        npix = 30;
+        data = rand(9, npix);
+        faccess = FakeFAccess(data);
+        faccess = faccess.set_npixels(uint64(npix));
+
+        pix = PixelData(faccess);
+        assertEqual(class(pix.num_pixels), 'double');
+    end
+
     % -- Helpers --
     function do_pixel_data_loop_with_f(obj, func, data)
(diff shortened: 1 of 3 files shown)


_test/test_sqw/test_PixelData.m
```diff
@@ -545,6 +545,16 @@ function test_constructing_from_PixelData_with_valid_file_inits_faccess(obj)
         assertTrue(new_pix.has_more());
     end
 
+    function test_num_pixels_is_a_double_if_faccess_returns_uint(obj)
```
nickbattam-tessella, line 548: You need tests confirming that num_pix is double in the three use-cases:
- faccess backed (<1 page)
- in-memory backed (e.g. array created object)
- file-backed contained less than one-page of data

PR #276, opened 2020-07-02
Add file backed write access to PixelData API

### LINKED ISSUE #247: PixelData - Add file-backed write access to API
Add support for paged/partial file writes to the core PixelData object

Part of #224

Output

- [ ] Update PixelData to support file backed array write
- [ ] Create consistent API for handling of large and small data
- [ ] Extend Setter APIs to support setting of large values

hsautessella:
This PR adds file-backed writing to the `PixelData` class. It does this by:

- A user loads an sqw file, which loads some pixel data
- the user does something like `my_sqw.data.pix.signal = 1;` 
- we then mark the current page of data "dirty" (the data in the cache does not match the data in the sqw file)
- user executes `pix.advance()`
- on the `advance` call, we write the edited page of data to a tmp file
- if the user then requests the page of data in which they set the `signal` to 1, instead of loading data from the sqw file, it is loaded from the tmp file

I encourage a close look at the new tests and whether there are any cases I've missed.

Fixes #247 

#### Issues to raise:
- SMG on PixelData use and how it operates under-the-hood
- Change the way `advance` works (?). Currently, `advance` loads the next page of data when it's called. Should this be changed to:
    a) clear the cache
    b) set `page_number` to next page 
    c) load data when a setter/getter is called


horace_core/sqw/@PixelData/PixelData.m
```diff
@@ -577,6 +714,68 @@
...
+            fclose(file_id);
+            rethrow(ME);
+        end
+        fclose(file_id);
+    end
+
+    function obj = write_pix_to_file_(obj, file_id)
+        % Write the pixels in the current page to the file corresponding to the
+        % given file ID.
+        % TODO: improve this by not writing all data at once
```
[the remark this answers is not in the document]
hsautessella: Bearing in mind that the page sizes may be quite large, I thought the writing may be more efficient if we write in "blocks". This is what's done in [`put_pix`](https://github.com/pace-neutrons/Horace/blob/ca88999fb95b4e1448ef0febb2826f09caa0893c/horace_core/sqw/file_io/%40sqw_binfile_common/put_pix.m#L171)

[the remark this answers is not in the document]
nickbattam-tessella: It would be worth expanding that TODO to include that extra detail

horace_core/sqw/@PixelData/PixelData.m
```diff
@@ -577,6 +701,81 @@
         page_size = floor(mem_alloc/num_bytes_in_pixel);
     end
 
+    function is = page_is_dirty_(obj, page_number)
```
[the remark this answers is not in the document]
nickbattam-tessella: The majority (all except the clean/dirty read/write switches call `page_is_dirty` or `set_page_dirty` with anything the current page number.
Would it be cleaner to make the pagenumber optional and default to using the current page number? That would remove a lot of `obj.set_page_dirty_(true, obj.page_number_);` calls.

If that API is useful a wrapper `set_current_page_dirty()` function could be created that calls that without needing the caller to pass any arguments.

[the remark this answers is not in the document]
hsautessella: Maybe it's just a Python thing, but I usually write names like this. `if` statements sound more coherent: `if page_is_dirty()` flows better than `if is_page_is_dirty()`.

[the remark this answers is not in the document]
nickbattam-tessella: You're right; this is fine. `is_<verb>` is more of a Java convention.

horace_core/sqw/PixelData/@PixelTmpFileHandler/PixelTmpFileHandler.m
```diff
@@ -0,0 +1,117 @@
...
+        file_id = fopen(tmp_file_path, 'wb');
+        if file_id < 0
+            error('PIXELTMPFIELHANDLER:write_dirty_pix', ...
+                  'Could not open file ''%s'' for writing.\n', tmp_file_path);
+        end
+
+        try
+            obj.write_pix_to_tmp_file_(file_id, pix_data);
+        catch ME
+            fclose(file_id);
```
nickbattam-tessella, line 71: The `fclose(file_id)` can live in an `onCleanup` handler so you don't have to have it in the catch and in outside the block (the same pattern can be used in the `load_dirty_pix` method above

horace_core/sqw/PixelData/@PixelData/PixelData.m
```diff
@@ -554,19 +666,44 @@
...
 
-    function obj = load_first_page_if_data_empty_(obj)
-        % Check if there's any data in the current page and load a page if not
-        if isempty(obj.data_) && ~isempty(obj.f_accessor_)
-            obj = obj.load_page_(1);
+    function obj = load_dirty_page_(obj, page_number)
+        % Load a page of data from a tmp file
+        raw_pix = obj.tmp_io_handler.load_page(page_number);
+        npix_cols = obj.PIXEL_BLOCK_COLS_;
+        obj.data_ = reshape(raw_pix, [npix_cols, numel(raw_pix)/npix_cols]);
```
nickbattam-tessella, line 680: why is the reshape here, not in the io_handler? 

In the current API, `write_page` takes a `nx9` array argument and load_page returns a `1x9n` vector. That difference is likely to lead to errors

horace_core/sqw/PixelData/@PixelTmpFileHandler/PixelTmpFileHandler.m
```diff
@@ -0,0 +1,110 @@
...
+                      'Could not write to file ''%s'':\n%s', ...
+                      tmp_file_path, ferror(file_id));
+            end
+        end
+    end
+
+    function tmp_file_path = generate_dirty_pix_file_path_(obj, page_number)
+        % Generate the file path to the tmp directory for this object instance
+        tmp_file_path = fullfile(obj.dirty_pix_dir_, ...
+                                 sprintf('%09d.tmp', page_number));
```
nickbattam-tessella, line 105: since you've extracted the temp dir name pattern as a constant it would be consistent to extract the filename pattern too

horace_core/sqw/PixelData/@PixelData/PixelData.m
```diff
@@ -554,19 +666,44 @@
...
+    function obj = set_page_dirty_(obj, is_dirty, page_number)
+        % Mark the given page as "dirty" i.e. the data in the cache does not
+        % match the data in the original SQW file
+        %
+        % Input
+        % -----
+        % is_dirty     Logical specifying if the page is dirty
+        % page_number  The page number to mark as dirty (default is current page)
+        %
+        if nargin == 2
```
nickbattam-tessella, line 703: The `nargin==2` check catches an omitted page number. What happens if `nargin==1`, i.e. the `is_dirty` flag is omitted?

horace_core/sqw/PixelData/@PixelData/PixelData.m
hsautessella: The reshape is in the `PixelData` class because only the `PixelData` instance knows how many columns it has. The IO handler doesn't necessarily need to know that information, it just reads/writes bytes, and the PixelData instance is responsible for knowing how to interpret those bytes. `write_page` can take any shape data

horace_core/sqw/PixelData/@PixelData/PixelData.m
hsautessella: You get a `Not enough input arguments.` error

horace_core/sqw/PixelData/@PixelTmpFileHandler/PixelTmpFileHandler.m
```diff
@@ -0,0 +1,112 @@
...
+        dirty_pix_dir_name = sprintf(obj.DIRTY_PIX_DIR_BASE_NAME_, obj.pix_id_);
+        obj.dirty_pix_dir_ = fullfile(tempdir(), dirty_pix_dir_name);
+    end
+
+    function raw_pix = load_page(obj, page_number, ncols)
+        % Load a page of data from the tmp file with the given page number
+        %
+        % Input
+        % -----
+        % page_number   The number of the page to read data from
```
nickbattam-tessella, line 34: missing `ncols` argument description. Should it error or default to 1 if omitted?

horace_core/sqw/PixelData/@PixelTmpFileHandler/PixelTmpFileHandler.m
```diff
@@ -0,0 +1,112 @@
+classdef PixelTmpFileHandler
+
+properties (Constant, Access=private)
+    DIRTY_PIX_DIR_BASE_NAME_ = 'sqw_pix%05d';
```
nickbattam-tessella, line 4: do these need to be `DIRTY_`? This temp file IO class just deals with temporary files.

PR #278, opened 2020-07-06
Handle copying PixelData instances with tmp files

### LINKED ISSUE #277: Handle copying PixelData instances with tmp files
#276 added file-backed write access to the `PixelData` class. This involved creating `tmp` files to hold pages of edited data. This creates an issue when we copy a `PixelData` instance, since both instances will now point at the same tmp files, if the copy edits a page, this edit will also affect the original's data.

The simplest solution is to copy all the tmp files into a new directory and have the new, copied, PixelData instance point at the copied data. But, there may be something more efficient we can do.

There's also a question as to what should happen with the `sqw` file backing the PixelData object. We will probably need to make a copy of that as well.

The behaviour should be:

1) changes to the copy do no affect the original
2) changes to the original after copying do not affect the copy
3) the copy contains changes made to the original before the copy operation

hsautessella:
This PR implements a non-default `copy` method on `PixelData` that copies temporary files used to save edits to pages of data. This ensures `PixelData` copies are entirely independent of one another.

This PR does _not_ implement a similar thing for copies of `.sqw` files. Creating a copy of an `sqw` object will copy the `PixelData` instance as described above, however the two `PixelData` instances will both still point at the same `.sqw` file; changes to the `.sqw` file will affect both `PixelData` instances. This is a more complicated issue than the one this PR solves. 

Fixes #277

### DIFF
--- a/horace_core/sqw/PixelData/@PixelData/PixelData.m
+++ b/horace_core/sqw/PixelData/@PixelData/PixelData.m
@@ -236,7 +236,7 @@ methods
         end
         % In memory construction
         if isa(arg, 'PixelData')  % TODO make sure this works with file-backed
-            if ~isempty(arg.file_path) && exist(arg.file_path, 'file')
+            if arg.is_file_backed_() && exist(arg.file_path, 'file')
                 % if the file exists we can create a file-backed instance
                 obj = PixelData(arg.file_path, arg.page_memory_size_);
                 obj.page_number_ = arg.page_number_;
@@ -343,7 +343,7 @@ methods
         if ~isa(fields, 'cell')
             fields = {fields};
         end
-        obj = obj.load_first_page_if_data_empty_();
+        obj = obj.load_current_page_if_data_empty_();
         try
             field_indices = cell2mat(obj.FIELD_INDEX_MAP_.values(fields));
         catch ME
@@ -376,7 +376,7 @@ methods
         % -------
         %   pixels      PixelData object containing a subset of pixels
         %
-        obj = obj.load_first_page_if_data_empty_();
+        obj = obj.load_current_page_if_data_empty_();
         pixels = PixelData(obj.data(:, pix_indices));
     end
 
@@ -443,7 +443,7 @@ methods
 
     % --- Getters / Setters ---
     function pixel_data = get.data(obj)
-        obj = obj.load_first_page_if_data_empty_();
+        obj = obj.load_current_page_if_data_empty_();
         pixel_data = obj.data_;
     end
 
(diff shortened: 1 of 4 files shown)


horace_core/sqw/PixelData/@PixelTmpFileHandler/PixelTmpFileHandler.m
```diff
@@ -75,6 +74,26 @@
         obj.write_float_data_(file_id, raw_pix);
     end
 
+    function copy_tmp_folder(obj, new_pix_id)
```
nickbattam-tessella, line 77: `destination_` or `target_pix_id` rather than 'new'?

horace_core/sqw/PixelData/@PixelData/PixelData.m
```diff
@@ -236,10 +236,12 @@
         end
         % In memory construction
         if isa(arg, 'PixelData')  % TODO make sure this works with file-backed
-            if ~isempty(arg.file_path) && exist(arg.file_path, 'file')
+            if arg.is_file_backed_() && exist(arg.file_path, 'file')
                 % if the file exists we can create a file-backed instance
                 obj = PixelData(arg.file_path, arg.page_memory_size_);
                 obj.page_number_ = arg.page_number_;
+                obj.page_dirty_ = arg.page_dirty_;
+                arg.tmp_io_handler_.copy_tmp_folder(obj.object_id_);
```
nickbattam-tessella, line 244: Would it be clearer as `arg.tmp_io_handler_.copy_folder(obj.object_id_)`? It's quite hard to read this link is snake case already and repetition of `_tmp` makes it harder -- the tmp_io_handler only does things with tmp_dirs and tmp_files.

horace_core/sqw/PixelData/@PixelData/PixelData.m
```diff
@@ -1,4 +1,4 @@
-classdef PixelData < matlab.mixin.Copyable
+classdef PixelData < handle
```
nickbattam-tessella, line 1: are there any other consequences swapping from the `Copyable` mixin to being a simple handle class?

hsautessella: Not that I can think of, we'll see in testing. The `Copyable` parent class implements a `copy` method that does a shallow copy of all attributes, that's all it does. I've implemented our own `copy` method which instantiates a _new_ object and then shallow copies across the necessary properties.

A consequence of this is that when we add new properties, we'll need to think about whether they need to be copied in the copy-constructor, or whether the normal constructor can handle it.

I decided this way was better since you cannot override the `copy` method implemented by `Copyable`. So I couldn't easily add the `arg.tmp_io_handler_.copy_folder(obj.object_id_);` step. There may have been a way using [`copyElement`](https://uk.mathworks.com/help/matlab/ref/matlab.mixin.copyable-class.html), but it felt simpler to do what I did here.

nickbattam-tessella: It may be worth adding that note as a comment either in the `copy()` method or in the code block where new properties would be added.

horace_core/sqw/PixelData/@PixelTmpFileHandler/PixelTmpFileHandler.m
```diff
@@ -89,30 +108,44 @@ function delete_tmp_files(obj)
...
-                      tmp_file_path, ferror(file_id));
+        SIZE_OF_FLOAT = 4;
+        chunk_idx_length = hor_config().mem_chunk_size/SIZE_OF_FLOAT;
+        data_size = SIZE_OF_FLOAT*numel(pix_data);  % in bytes
+        num_chunks = ceil(data_size/hor_config().mem_chunk_size);
+
+        for i = 1:num_chunks
+            start_idx = (i - 1)*chunk_idx_length + 1;
+            end_idx = min(i*chunk_idx_length, numel(pix_data));
+            try
```
nickbattam-tessella, line 119: the `try ... catch` could be outside the `for` loop to keep the loop chunk smaller. In either case it will abort the loop as soon an an error is thrown, you're not recording the chunk index in any of the log messages.

PR #280, opened 2020-07-07
Added description of the internal release in isis

abuts:
Small addition -- I think, self-explanatory.


nickbattam-tessella:
This is the wrong document to contain this information -- this is about build pipelines not release locations. https://github.com/pace-neutrons/Horace/blob/251-release-locations/documentation/smg/11_ISIS_release.md was created as a template for this information.

abuts:
I have added the right document to another branch, but think both should exist despite some overlap as reflect the issue from different points of view and make things much clearer.

nickbattam-tessella:
I disagree -- we don't want information in two places. 

There is a lot of detail here that is specific to how/where to do with the releases once they've been created in the build pipelines. That information about paths, groups and scripts should only exist in one document: the ISIS_release guide.

PR #192, opened 2020-04-09
Remove references to libisis in comments

nickbattam-tessella:
Code changes are in Herbert PR https://github.com/pace-neutrons/Herbert/pull/132

Fixes #183

PR #291, opened 2020-07-15
transform sqw in parallel

abuts:
These changes add unit test to confirm the user's issue + test for issue itself.

The solution is actually in Herbert [PR#193](https://github.com/pace-neutrons/Herbert/pull/193), ticket [#192](https://github.com/pace-neutrons/Herbert/issues/192)

Should not be merged before changes in Herbert


horace_core/sqw/@rundatah/calc_sqw.m
```diff
@@ -1,4 +1,4 @@
-function [w,grid_size,urange,detchn] = calc_sqw(obj,grid_size_in,urange_in,varargin)
+function [w,grid_size,urange,detdcn] = calc_sqw(obj,grid_size_in,urange_in,varargin)
```
nickbattam-tessella, line 1: the `detdcn` return value isn't documented in the method header

PR #295, opened 2020-07-17
Build mex libraries directly into DLL directory

### LINKED ISSUE #294: Herbert: Build mex files directly into DLL directory
https://github.com/pace-neutrons/Herbert/issues/200

Making this change in Herbert will do the same in Horace, since the CMake file that needs updating is shared.

hsautessella:
Update to make things consistent with https://github.com/pace-neutrons/Herbert/pull/201.

This is a less disruptive change than in Herbert and should be merged first.

Fixes #294

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,8 +23,10 @@ list(APPEND CMAKE_MODULE_PATH
 )
 
 # Set some useful global variables
-# this will hold herbert/horace_on.m and worker_v2.m
+# This will hold herbert/horace_on.m and worker_v2.m
 set(LOCAL_INIT_DIR "${CMAKE_BINARY_DIR}/local_init")
+# This sets the destination for mex build artifacts (used in PACE_AddMex)
+set(Horace_DLL_DIRECTORY "${CMAKE_SOURCE_DIR}/horace_core/DLL")
 
 # Set our options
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -23,8 +23,10 @@ list(APPEND CMAKE_MODULE_PATH
 )
 
 # Set some useful global variables
-# this will hold herbert/horace_on.m and worker_v2.m
+# This will hold herbert/horace_on.m and worker_v2.m
 set(LOCAL_INIT_DIR "${CMAKE_BINARY_DIR}/local_init")
+# This sets the destination for mex build artifacts (used in PACE_AddMex)
+set(Horace_DLL_DIRECTORY "${CMAKE_SOURCE_DIR}/horace_core/DLL")
```
nickbattam-tessella, line 29: is this still writing into Debug/Release while Herbert has been changed to use a common folder for both

hsautessella: No, if you compare the top-level CMakeLists file in Herbert and this one, the `*_DLL_DIRECTORY` points to the same (relative) place. The `PACE_AddMex` function tells the compiler not to create Release/Debug folders within `*_DLL_DIRECTORY` and that file is shared between Herbert and Horace

PR #299, opened 2020-07-20
PixelData - Implement file backed unary operations

### LINKED ISSUE #248: PixelData - Implement file-backed unary operations
Update SQW unary operations to handle large data

Part of #224 

Output

- [ ] SQW unary operations support large PixelData API
- [ ] Consistent API for handling of large and small data
- [ ] Consistent DND / SQW API

hsautessella:
This PR updates the `unary_op_manager` to operate on file-backed `PixelData` instances.

Fixes #248


horace_core/sqw/@sqw/private/unary_op_manager.m
```diff
@@ -14,10 +14,8 @@
 w = w1;
 for i=1:numel(w1)
     if is_sqw_type(w1(i))
```
nickbattam-tessella, line 16: does the `else` here need updating too?

_test/test_sqw/test_PixelData_operations.m
```diff
@@ -24,6 +33,9 @@
 
         file_info = dir(obj.test_sqw_file_path);
         obj.page_size = file_info.bytes/6;  % Gives us 6 pages
+
+        obj.pix_in_memory = sqw_test_obj.data.pix;
+        obj.pix_with_pages = PixelData(obj.test_sqw_file_path, obj.page_size);
```
nickbattam-tessella, line 38: this assumes there will always be room for the test data plus 1/6th of the test data in memory: is that true?

_test/test_sqw/test_PixelData_operations.m
```diff
@@ -87,31 +95,172 @@ function test_compute_bin_data_correct_output_5_pages_mex_1_thread(obj)
...
+        iter = 0;
+        while true
+            start_idx = (iter*npix_in_page) + 1;
+            end_idx = min(start_idx + npix_in_page - 1, pix.num_pixels);
+
+            original_signal = data(obj.SIGNAL_IDX, start_idx:end_idx);
+            original_variance = data(obj.VARIANCE_IDX, start_idx:end_idx);
+
+            expected_data = data;
+            expected_data(obj.SIGNAL_IDX, start_idx:end_idx) = ...
```
nickbattam-tessella, line 139: A comment here explaining where the signal and variance calculations are taken from would make this explicit

_test/test_sqw/test_PixelData_operations.m
```diff
@@ -87,31 +95,172 @@ function test_compute_bin_data_correct_output_5_pages_mex_1_thread(obj)
...
+            data = obj.get_random_data_in_range(obj.NUM_COLS_IN_PIX_BLOCK, ...
+                                                num_pix, data_range);
+            pix = obj.get_pix_with_fake_faccess(data, npix_in_page);
+            pix = pix.do_unary_op(unary_op);
+
+            transformed_data = obj.concatenate_pixel_pages(pix);
+
+            pix_in_mem = PixelData(data);
+            pix_in_mem = pix_in_mem.do_unary_op(unary_op);
+            expected_transformed_data = pix_in_mem.data;
```
nickbattam-tessella, line 207: rather than `transformed_data` and `expected_transformed_data` these are `in_memory` and `file_backed` data

_test/test_sqw/test_PixelData_operations.m
```diff
@@ -87,31 +95,172 @@ function test_compute_bin_data_correct_output_5_pages_mex_1_thread(obj)
...
+
+            assertEqual(transformed_data, expected_transformed_data, ...
+                        sprintf('Failure calculating ''%s''', char(unary_op)), ...
```
nickbattam-tessella, line 210: "failure calculating"?

_test/test_sqw/test_PixelData_operations.m
```diff
@@ -87,31 +95,172 @@ function test_compute_bin_data_correct_output_5_pages_mex_1_thread(obj)
...
+                        obj.FLOAT_TOLERANCE);
+        end
     end
 
+    function test_do_unary_op_returns_correct_output_with_cosine_1_page(obj)
```
nickbattam-tessella, line 215: This is the same test as the gt_1_page test above but just looking at a single page worth of data (which happens to be all of the data). One of these is unnecessary.
The all-functions routine has tested that multi-page and single page execution of all operations return the same data, so you only need to demonstrate one page (a few data points) works correctly for every (or just cos) operation.

_test/test_sqw/test_PixelData_operations.m
```diff
@@ -87,31 +95,172 @@ function test_compute_bin_data_correct_output_5_pages_mex_1_thread(obj)
...
+    end
+
+    % -- Helpers --
+    function pix = get_pix_with_fake_faccess(obj, data, npix_in_page)
+        faccess = FakeFAccess(data);
+        mem_alloc = npix_in_page*obj.NUM_BYTES_IN_VALUE*obj.NUM_COLS_IN_PIX_BLOCK;
+        pix = PixelData(faccess, mem_alloc);
+    end
+
+    function data = concatenate_pixel_pages(obj, pix)
```
nickbattam-tessella, line 240: does this function need testing? Its correct behaviour is critical to the all-operator test above.

_test/test_sqw/test_PixelData_operations.m
hsautessella: For the test data it's using, yes. It's only a few Mb. Many of the tests rely on us being able to hold the data in memory (e.g. `test_compute_bin_data_correct_output_in_memory_mex_1_thread`), so using data that doesn't fit in memory does not make sense for this test suite.

horace_core/sqw/@sqw/private/unary_op_manager.m
hsautessella: I don't think so, the `else` is for dnd data, so no pixels

PR #303, opened 2020-07-22
tilde in path

abuts:
This is the Horace part of the ticket, allowing users to have tilde in path

### DIFF
--- a/horace_core/algorithms/write_nsqw_to_sqw.m
+++ b/horace_core/algorithms/write_nsqw_to_sqw.m
@@ -201,7 +201,6 @@ if hor_log_level>-1
     disp(' ')
     disp('Reading and accumulating binning information of input file(s)...')
 end
-[]
 
 if combine_in_parallel
     %TODO:  check config for appropriate ways of combining the tmp and what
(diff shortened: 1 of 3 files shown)


horace_core/algorithms/write_nsqw_to_sqw.m
```diff
@@ -248,6 +248,22 @@
...
 clear nopix
 
+[outfile,file_exist] = resolve_path(outfile);
+if ~file_exist % file not exist or path to file can not be resolved
+    file_path = fileparts(outfile);
+    if ~isempty(file_path) && ~(exist(file_path,'dir') == 7)
+        error('WRITE_NSQW_TO_SQW:invalid_argument',....
+            'Can not find folder to write file %s',outfile);
+    end
+    fh = fopen(outfile,'w');
```
nickbattam-tessella, line 258: this identical block of code appears twice: resolvePath, openfile, closefile, delete.
Would it be clearer to call this as a "check file is writable" helper function?

abuts: one is in private sqw, another one in private algorithms. And making it public and placing in helpers....
do not know. May be worth thinking but the code is not that extensive...
Will have a look.

PR #305, opened 2020-07-22
Add tests for SQW binary operations

### LINKED ISSUE #304: Add tests for SQW binary operations
Add a suite of tests for SQW binary operations as part of #249.

hsautessella:
Adds a suite of tests for SQW binary operations. 

The tests written are not meant to test that the individual binary operation implementations are correct e.g. `plus`, `minus`. The point of the tests is to ensure that performing binary operations with SQW objects behaves as it should, i.e. correct type returned, no errors raised, errors raised where they should be.

Fixes #304

### DIFF
--- a/_test/test_sqw/test_binary_ops.m
+++ b/_test/test_sqw/test_binary_ops.m
@@ -1,7 +1,9 @@
 classdef test_binary_ops < TestCase
-
+%% TEST_BINARY_OPS tests for behaviour of binary operations between SQW and
+% and other type of objects e.g. scalars, dnd and other SQW objects
+%
 properties (Constant)
-    DOUBLE_REL_TOLERANCE = 5e-7;
+    DOUBLE_REL_TOLERANCE = 10e-6;
 end
 
 properties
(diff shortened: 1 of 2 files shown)


_test/test_sqw/test_binary_ops.m
```diff
@@ -0,0 +1,123 @@
...
+        f = @() int64(ones(size(obj.sqw_obj.data.npix))) + obj.sqw_obj;
+        assertExceptionThrown(f, 'SQW:binary_op_manager_single');
+    end
+
+    function test_SQW_error_if_first_input_is_char(obj)
+        f = @() 'some char' + obj.sqw_obj;
+        assertExceptionThrown(f, 'SQW:binary_op_manager_single');
+    end
+
+    function test_SQW_error_if_second_input_not_supported_type(obj)
```
nickbattam-tessella, line 37: what does 'not supported type' mean here?

horace_core/sqw/@sqw/private/binary_op_manager_single.m
```diff
@@ -10,6 +10,13 @@
 %           >> nd = dimensions(obj)
 %   (3) have private function that returns class name
 %           >> name = classname     % no argument - gets called by its association with the class
+%
+allowed_types = {'double', 'd0d', 'd1d', 'd2d', 'd3d', 'd4d', 'sqw', 'sigvar'};
```
nickbattam-tessella, line 14: you test that `int64` is prohibited, but not that `int32`. Is `float` supported as `double` is?

hsautessella: I've added cases for other non-double numeric types. The existing code only allowed `double` not any other numeric types (including `float`), so I've left as is. If we need to support `float` as well that should be trivial.

PR #311, opened 2020-07-28
Prevent re-writing of dirty PixelData pages

### LINKED ISSUE #310: PixelData re-writing unchanged dirty pages
As an example, when performing a unary operation we do:
- load page
- `unary_op(page)`
- write page
and repeat until finish.

Then we recompute bin data:
- load page
- sum signal and variance
- _write dirty page!?_
- load next page
repeat until finish

When a page has been marked "dirty" during the unary operation loop, it remains "dirty" in the `recompute_bin_data` loop. This means when we call `advance` in the `recompute_bin_data` loop, the page is written again even though it hasn't been edited, causing a significant performance drop.

hsautessella:
When dirty PixelData pages were read in, even if those pages were not edited, the pages were being re-written to disk.

A new dirty_page_edited_ flag has been added to PixelData to prevent the pages being written again if they weren't edited.

Fixes #310

### DIFF
--- a/horace_core/sqw/PixelData/@PixelData/PixelData.m
+++ b/horace_core/sqw/PixelData/@PixelData/PixelData.m
@@ -446,7 +446,7 @@ methods
         %  This function does nothing if pixels are not file-backed.
         %
         if obj.is_file_backed_() && obj.page_number_ ~= 1
-            if obj.page_is_dirty_(obj.page_number_)
+            if obj.page_is_dirty_(obj.page_number_) && obj.dirty_page_edited_
                 obj.write_dirty_page_();
             end
             obj.page_number_ = 1;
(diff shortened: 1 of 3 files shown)


horace_core/sqw/PixelData/@PixelData/PixelData.m
```diff
@@ -448,6 +450,7 @@ function delete(obj)
                 obj.write_dirty_page_();
```
nickbattam-tessella, line 450: does this `obj.write_dirty_page_`, `if obj.page_is_dirty_()` check need the additional `dirty_page_edited` check (c.f. `advance()`)?

horace_core/sqw/PixelData/@PixelData/PixelData.m
```diff
@@ -734,6 +738,7 @@ function delete(obj)
             page_number = obj.page_number_;
         end
         obj.page_dirty_(page_number) = is_dirty;
+        obj.dirty_page_edited_ = true;
```
nickbattam-tessella, line 741: this will be set to true even if called with `is_dirty=false`

PR #318, opened 2020-08-11
Fix mex error when applying unary ops to multi-dimensional SQW

### LINKED ISSUE #308: Crash when applying unary ops to large SQW object
Applying a unary operation to the iron data set used in the training course (~5Gb) causes Matlab to crash. This only happens when using mex routines, the Matlab routine works fine. 

Most likely this crash started occurring after pixel data became file-backed. There are unit tests for small data in `test_PixelData_operations` that perform unary operations over multiple pages of data using mex functions, these appear to pass.

To recreate: 
```matlab
s = sqw('iron_data.sqw');
s2 = cos(s);
```
Matlab will crash (I think due to a segfault in `compute_pix_sums_c.mex`)

hsautessella:
Fix segfault when running `recompute_bin_data`on an SQW object with more than 1 dimension. 

Also, a quick fix to avoid editing the `PixelData` instance when calling `PixelData.do_unary_op` with a return argument.

Fixes #308

### DIFF
--- a/_test/test_sqw/test_PixelData_operations.m
+++ b/_test/test_sqw/test_PixelData_operations.m
@@ -194,6 +194,13 @@ methods
         end
     end
 
+    function test_do_unary_op_with_nargout_1_doesnt_affect_called_instance(obj)
+        data = rand(9, 10);
+        pix = PixelData(data);
+        sin_pix = pix.do_unary_op(@sin);
+        assertEqual(pix.data, data);
+    end
+
     function test_paged_data_returns_same_unary_op_result_as_all_in_memory(obj)
         % the unary operation and the range the data it acts on should take
         unary_ops = {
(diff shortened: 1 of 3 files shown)


abuts:
looks fine to me and fix is clear. Are you doing page in bytes or words? if bytes, there is discrepancy with hor_config as buffer there is expressed in words (9 pixels values)

hsautessella:
I'm not clear what you mean. `PixelData.page_size` returns the number of pixels in the current page. When you instantiate a `PixelData` object the second constructor argument is a memory allocation, that's given in bytes.

abuts:
I think, the constructor can be issued without the second argument, and if it is not there, the page size should be taken from hor_config (modified with `page_size`)
if the page size is present, its values is compared with hor_config value, and if it is different, checked against number of instances of PixData is performed. If other pixData exist, with different page size, the construction is refused, if not, new configuration page_size should be set-up

abuts:
I do not see a point to have more then one page_size per session and if this is the case, the correct place for page size is configuration.

PR #324, opened 2020-08-19
Implement "append" in PixelData class

### LINKED ISSUE #323: Implement "append" in PixelData class
This method will allow you to append pixels to an existing `PixelData` object. This will be a public facing method, but will also be useful to us in completing #322.

hsautessella:
This PR adds method `append` to the `PixelData` class. This allows one to append the contents of a `PixelData` object to another `PixelData` object. 

The `PixelData` that you're appending must be all in memory and have equal or fewer pixels than are allowed in a page. This means, to append a file-backed PixelData object, you need to do the looping yourself.

Fixes #323


horace_core/sqw/PixelData/@PixelData/append.m
```diff
@@ -0,0 +1,62 @@
...
+    if obj.page_is_dirty_(obj.page_number_) && obj.dirty_page_edited_
+        obj.write_dirty_page_();
+    end
+    obj.load_page_(obj.get_num_pages_());
+end
+
+if isempty(obj)
+    obj.data_ = pix.data;
+    obj.set_page_dirty_(true);
+else
```
nickbattam-tessella, line 38: does the non-empty case need to be a nested `if` rather than `if ... elsif ... else` ?

that would make the three cases we are handling more explicit:
- current object is empty
- current page is full (=> save and create a new page)
- current page is partially full (=> fill current page, save, put remaining pixels in new page)

horace_core/sqw/PixelData/@PixelData/append.m
```diff
@@ -0,0 +1,62 @@
...
+
+        if num_to_allocate_to_pg ~= pix.num_pixels
+            obj.write_dirty_page_();
+
+            obj.data_ = pix.data(:, (num_to_allocate_to_pg + 1):end);
+            obj.page_number_ = obj.page_number_ + 1;
+        end
+
+    elseif obj.page_size == obj.max_page_size_
+        obj.set_page_dirty_(true);
```
nickbattam-tessella, line 52: `obj.set_page_dirty_(true);` appears in all three branches of the `if` -- is there anything preventing that being moved to before the start of the `if` (is order important)

nickbattam-tessella (changes requested):
Missing tests for appending an empty pixdata (the `if isempty(pix)` codepath

Although it's said in the comments, we should add an exception (and test) to catch an attempt to appending a file-backed PixelData object

Are the changes to the way num-pixels are calculated/handled tested by the existing tests?

horace_core/sqw/PixelData/@PixelData/append.m
hsautessella: I've cleaned up the logic a bit - but I don't think it can be brought outside. Order is important here, in some branches we edit an existing page (so it must be marked dirty) then we create a new page, which must also be marked dirty. If the original page is full, then we don't edit it so it need not be marked dirty.

_test/test_sqw/test_PixelData.m
```diff
@@ -885,6 +885,155 @@ function test_unedited_dirty_pages_are_not_rewritten(obj)
...
+        assertExceptionThrown(f, 'PIXELDATA:append');
+    end
+
+    function test_append_does_not_edit_calling_instance_if_nargout_eq_1(obj)
+        data = rand(9, 30);
+        npix_in_page = 11;
+        pix = obj.get_pix_with_fake_faccess(data, npix_in_page);
+        pix_to_append = PixelData(rand(9, 5));
+
+        out_pix = pix.append(pix_to_append);
```
nickbattam-tessella, line 1023: there is no `assertEqual` to check the appended pixels *are* included in the new PixelData return value

_test/test_sqw/test_PixelData.m
```diff
@@ -917,11 +1078,26 @@ function do_pixel_data_loop_with_f(obj, func, data)
 methods (Static)
 
     function advance_pix(pix, niters)
+        % Advance the pixel pages by 'niters'
```
nickbattam-tessella, line 1081: `niters`? That doesn't give the caller any idea what they're iterating. `n[um_]pages` would be clearer.

nickbattam-tessella (approved):

_test/test_sqw/test_PixelData.m
nickbattam-tessella: assigning that as a variable in the call may make what the test is doing clearer too 
```
num_pages_to_advance = <maths>
advance_pix(pix_data, num_pages_to_advance)
```

PR #336, opened 2020-08-27
Add PixelData page size config option

### LINKED ISSUE #320: Add PixelData page size config option
A config option should be added that allows the user to set the page size for `PixelData` objects. Some care will need to be taken as this setting could result in having two separate `PixelData` objects with different page sizes. The differing page sizes will likely interfere with binary operations between the two objects.

**Output:**
- A `hor_config` option `pixel_data_page_size` (or similar)
- Some guarantees in behaviour when `PixelData` objects with different page sizes interact. We could:
    - not allow changing the page size setting if a file-backed `PixelData` object exists
    - change the page size of existing `PixelData` objects when the setting is changed
    - adapt binary operations between `PixelData` objects such that they're not dependent on page size

hsautessella:
Add `pixel_page_size` setting to `hor_config`. This allows users to set a page size for `PixelData` objects. The default page size has been kept as 3GB.

Errors will be raised if the page size is too big (larger than can fit in memory) or too small (less than one pixel). A warning will also be shown if the page size is less than 100 MB, which we deem too small.

Fixes #320


horace_core/configuration/@hor_config/hor_config.m
```diff
@@ -186,6 +197,28 @@
             end
             config_store.instance().store_config(this,'mem_chunk_size',val);
         end
+
+        function this = set.pixel_page_size(this, val)
+            MIN_RECOMMENDED_PG_SIZE = 100e6;
+            if val < PixelData.DATA_POINT_SIZE*PixelData.DEFAULT_NUM_PIX_FIELDS
```
nickbattam-tessella, line 203: you compare against `PixelData.DATA_POINT_SIZE*PixelData.DEFAULT_NUM_PIX_FIELDS` but the error message says `72`

horace_core/configuration/@hor_config/hor_config.m
```diff
@@ -186,6 +197,28 @@
...
+                error('HOR_CONFIG:set_pixel_page_size', ...
+                      ['Error setting pixel page size. Cannot set page '...
+                       'size less than 72 bytes, as this is less than one pixel.']);
+            elseif val < MIN_RECOMMENDED_PG_SIZE
+                warning('HOR_CONFIG:set_pixel_page_size', ...
```
nickbattam-tessella, line 208: is this warning useful?

horace_core/sqw/PixelData/@PixelData/PixelData.m
```diff
@@ -238,6 +242,7 @@
         %               in-memory data. (Optional)
         %
         obj.object_id_ = polyval(randi([0, 9], 1, 5), 10);
+        obj.page_memory_size_ = get(hor_config, 'pixel_page_size');
```
nickbattam-tessella, line 245: this should be combined with the existing page-size code block which should be moved here from the 'file-backed' section of the constructor below:
```
if exist('mem_alloc', 'var')
    obj.page_memory_size_ = mem_alloc;
else
    obj.page_memory_size_ = get(hor_config, 'pixel_page_size');
end
```

nickbattam-tessella: There is no validation on the page size when passed as an argument. This should have the same validation you perform on the config option 'memalloc >= onerow, memalloc <= maxArraySize`

nickbattam-tessella (approved):

PR #338, opened 2020-09-01
Implement masking for file-backed sqw objects

### LINKED ISSUE #330: Implement masking for file-backed SQW objects
https://github.com/pace-neutrons/Horace/pull/329/ implemented a `mask` method for pixels. The `sqw` object's `mask` method should be updated to call `PixelData.mask`.

It may be more efficient if the new `urange` is calculated within `PixelData.mask`, so we avoid looping through pixel pages twice.

hsautessella:
This PR updates `sqw.mask` so that it works when the object contains a file-backed `PixelData` instance.

A clear optimization to what is implemented here would be to calculate the `urange` of the SQW object whilst performing `PixelData.mask`. As it is here, the masking constitutes one loop over the file-backed data and the `recompute_urange` function performs another loop over the data. Performing everything in one loop would be more efficient.

Fixes #330


horace_core/sqw/@sqw/private/recompute_urange.m
```diff
@@ -34,10 +34,30 @@
...
-urange(2,1:3)=max(u_q,[],2)';
-urange(1,4)=min(w.data.pix.dE) + (pix0(4)-u0(4));
-urange(2,4)=max(w.data.pix.dE) + (pix0(4)-u0(4));
 
+min_uq = realmax*ones(size(w.data.pix.q_coordinates));
+max_uq = -realmax*ones(size(w.data.pix.q_coordinates));
+min_dE = realmax;
+max_dE = -realmax;
+while true  % loop through pages of pixel data
+    u_q = (u_to_rlu\pix_to_rlu)*(w.data.pix.q_coordinates) ...
```
nickbattam-tessella, line 43: the constants here `u_to_rlu\pix_to_rlu`, `u_to_rlu\(pix0(1:3)-u0(1:3))` do not need to be recalculated on each loop, you're just calculating: `u_q = A*w.data.pix.q_coordinates + B`

_test/test_sqw/test_mask.m
```diff
@@ -0,0 +1,142 @@
...
+
+    function obj = test_mask(~)
+        obj = obj@TestCase('test_mask');
+
+        addpath(fullfile(obj.this_dir, 'utils'));
+
+        % Swallow any warnings for when pixel page size set too small
+        obj.old_warn_state = warning('OFF', 'PIXELDATA:validate_mem_alloc');
+
+        obj.sqw_2d = sqw(obj.sqw_2d_file_path);
```
nickbattam-tessella, line 26: does masking work correctly in higher dimensions? This is a 2d test case.

_test/test_sqw/test_PixelData_operations.m
```diff
@@ -58,6 +65,11 @@
                                           obj.page_size_2d);
     end
 
+    function delete(obj)
+        rmpath(fullfile(obj.this_dir, 'utils'));
```
nickbattam-tessella, line 69: can this delete throw an exception?

_test/test_sqw/test_mask.m
hsautessella: The masking routine should be dimension agnostic; if it works in 2D it will work in higher dimensions. 

At the `sqw` level as long as the `mask_array` argument to `mask` has the same number of dimensions as the image data it will work - an error is thrown otherwise. At the `PixelData` level, the `npix` argument is flattened using `repelem` as the array of pixels is flat, so any image dimensionality information is discarded.

nickbattam-tessella: "The masking routine should be dimension agnostic" the purpose of the tests is to show it _is_ not that is _should be_. It's possible to call repmat incorrectly, especially in higher dimensions.

Even if there are `PixelData` level tests that prove this works, an additional test at `SQW` level would confirm all the wiring is correct between the layers and give a code example of higher-dimension calls.

_test/test_sqw/test_mask.m
```diff
@@ -23,15 +31,28 @@
...
+        obj.sqw_2d_paged = obj.get_paged_sqw(obj.sqw_2d_file_path);
 
         obj.idxs_to_mask = [2, 46, 91, 93, 94, 107, 123, 166];
         obj.mask_array_2d = ones(size(obj.sqw_2d.data.npix), 'logical');
         obj.mask_array_2d(obj.idxs_to_mask) = 0;
 
         obj.masked_2d = mask(obj.sqw_2d, obj.mask_array_2d);
         obj.masked_2d_paged = mask(obj.sqw_2d_paged, obj.mask_array_2d);
+
+        % 3D case setup
```
nickbattam-tessella, line 45: is there any reason this isn't split into two separate cases: `test_mask_filters_correct_pixels_with_2d_sqw_object` and `test_mask_filters_correct_pixels_with_3d_sqw_object`?

hsautessella: Not sure what you mean. Do you mean two separate test classes?

nickbattam-tessella: Ignore this. Misread the collapsed code.

nickbattam-tessella (approved):

PR #344, opened 2020-09-07
325 multifit tests refactoring

### LINKED ISSUE #325: Multifit tests use the now obsolete multifit and/or fit syntax
Many of the multifit tests in Horace and Herbert use the legacy - now obsolete - calling syntax. These need to be converted to use the object-oriented multifit syntax.

Many multifit tests also use the legacy - now obsolete - fit methods. Users should be directed to use multifit instead rather than to the legacy function.

tgperring:
Fixes #325


nickbattam-tessella (approved):

abuts (approved):
ok, but a lot of disabled tests are not mentioned in disbled.rm file. Probaly worth separate ticket -- Toby would not have time for that

Issue #345, opened 2020-09-08
Update permute_pix_array for file-backed pixel data


hsautessella:
Update: Delete `permute_pix_array` function and `sqw.to_slice()`

~The `permute_pix_array` function needs to be updated such that it does not assume that all pixel data is held in memory.~

~An example of how to loop through pixel data is given in: https://github.com/pace-neutrons/Horace/blob/master/horace_core/sqw/PixelData/%40PixelData/private/compute_bin_data_mex_.m.~

Part of #250 

**Outputs:**
- [ ] `permute_pix_array` deleted

nickbattam-tessella:
This functionality is in slice, and the function is not required any more.

cmarooney-stfc:
Closed following PR merge to fix

PR #350, opened 2020-09-10
delete permute_pix_array and to_slice, remove mention in doc #345

cmarooney-stfc:
As per ticket #345 remove permute_pix_array and to_slice, also delete mention of permute_pix_array from add docs


nickbattam-tessella (approved):
The advantages of deleting unused, untested functions is they need very few changes.

nickbattam-tessella (changes requested):
The 02_sqw_current_implementation document still includes a reference to `to_slice` so that should be deleted if `permute_pix_array` is deleted. 

I wonder if that document should remain as (slightly unstructured) documentation of the original implementation rather than being updated to track the new implementation? 
If so, leaving these reference in perhaps striking them through would be more in the spirit of the document?

cmarooney-stfc:
Strikethrough now pushed

cmarooney-stfc:
Closed after merge with approval

PR #364, opened 2020-09-16
Add ability to build PR label branch in Jenkins

### LINKED ISSUE #363: Update PR-* pipelines to merge base PR branch
At the moment the PR-* pipelines always merge the pull request branch with master. This means, when we're building PRs to be merged into branches other than master, we're building the wrong thing.

This is fairly straightforward to do, but requires updating the pipelines in the GUI. The key `$.pull_request.base.ref` needs to be added to the Generic Webhook Trigger to extract the base branch name from the pull-request payload. 

The checkout settings then need to be updated to use the extracted base branch.

I've added a [test pipeline](https://anvil.softeng-support.ac.uk:444/jenkins/job/PACE-neutrons/job/Horace/job/PR-Scientific-Linux-7-TESTER-2018b/configure) that includes the updated settings.

hsautessella:
Adds the ability to build against a Herbert branch specified in a pull-request label.

When a PR build is triggered, the webhook's JSON is parsed for the pull request's labels. If the one of the labels follows the pattern "Herbert_<some_git_branch>" then "some_git_branch" will be pulled from Herbert and used in the build.

This PR and changes to the `PR-*` pipelines adding the extra webhook trigger parameter to capture the pulle request labels, fixes #363.

### DIFF
--- a/tools/build_config/Jenkinsfile
+++ b/tools/build_config/Jenkinsfile
@@ -75,6 +75,16 @@ def get_default_herbert_branch(String job_name) {
   }
 }
 
+def get_herbert_ref_from_labels(String labels, String herbert_branch) {
+  /* Parse the string of Git issue labels for a label that matches Herbert_*.
+     If a match is found, build using the Herbert branch of that name. */
+  def match = (label_names =~ "Herbert_([a-zA-Z0-9_-]+)")
+  try {
+    return match[0][1]
+  } catch (IndexOutOfBoundsException) {
+    return env.HERBERT_BRANCH_NAME
+  }
+}
 
 properties([
   parameters([
@@ -252,26 +262,28 @@ pipeline {
             }
 
           } else {
+            def herbert_ref = get_herbert_ref_from_labels(env.PR_LABELS,
+                                                          env.HERBERT_BRANCH_NAME)
             // Get Herbert from GitHub and call the build script
             if (isUnix()) {
-              sh '''
+              sh """
                 module load cmake/\$CMAKE_VERSION &&
                 module load matlab/\$MATLAB_VERSION &&
                 module load gcc/\$GCC_VERSION &&
                 ./tools/bash/clone_herbert_branch.sh\
-                  --branch \$HERBERT_BRANCH_NAME\
+                  --branch ${herbert_ref}\
                   --build_args \
                     \"--cmake_flags \\\"-DHerbert_RELEASE_TYPE=\$RELEASE_TYPE\\\" \
                     --matlab_release \$MATLAB_VERSION\"
-              '''
+              """
             } else {
-              powershell '''
+              powershell """
                 ./tools/pwsh/clone_herbert_branch.ps1\
-                  -branch \$env:HERBERT_BRANCH_NAME\
+                  -branch ${herbert_ref}\
                   -build_args\
                     \"-cmake_flags \"\"-DHerbert_RELEASE_TYPE=\$env:RELEASE_TYPE\"\" \
                     -matlab_release \$env:MATLAB_VERSION\"
-              '''
+              """
             }
           }
         }


tools/build_config/Jenkinsfile
```diff
@@ -75,6 +75,16 @@ def get_default_herbert_branch(String job_name) {
...
 }
 
+def get_herbert_ref_from_labels(String labels, String herbert_branch) {
+  /* Parse the string of Git issue labels for a label that matches Herbert_*.
+     If a match is found, build using the Herbert branch of that name. */
+  def match = (labels =~ "Herbert_([a-zA-Z0-9_-]+)")
+  try {
+    return match[0][1]
+  } catch (IndexOutOfBoundsException) {
+    return env.HERBERT_BRANCH_NAME
```
nickbattam-tessella, line 85: the default here is `herbert_branch` not the `env` variable.

tools/bash/clone_herbert_branch.sh
```diff
@@ -25,15 +25,21 @@ if [ "${herbert_branch}" = "" ]; then
...
+    popd || exit 1
+
 else
     git_clone_cmd="git clone ${HERBERT_URL} --depth 1"
     git_clone_cmd+=" --branch \"${herbert_branch}\" ${HERBERT_DIR}"
     echo_and_run "${git_clone_cmd}"
 fi
 
-echo -e "\nBuilding Herbert revision $(git -C ${HERBERT_DIR} rev-parse HEAD)..."
+pushd "${HERBERT_DIR}" || exit 1
```
nickbattam-tessella, line 39: Do we need an equivalent to the powershell `Invoke-and-run` in the bash script now to remove repeats of the pushd/action/popd pattern?

If the action between the pushd and popd fails we should fail to the original directory.

tools/build_config/Jenkinsfile
```diff
@@ -75,6 +75,22 @@ def get_default_herbert_branch(String job_name) {
   }
 }
 
+def get_herbert_ref_from_labels(String labels, String herbert_branch) {
+  /* Parse the string of Git issue labels for a label that matches Herbert_*.
+     If a match is found, build using the Herbert branch of that name. */
+  def match = (labels =~ "Herbert_([a-zA-Z0-9_-]+)")
+  try {
+    match[1]
+    return("Error: Found two Herbert branch labels on the pull request: " +
```
nickbattam-tessella, line 84: This message isn't correct -- you've found more than one label.
Using `match[x][1]` in the error message is risky. That could raise another `IndexOutOfBounds` exception which would cascade down into the (expected) catch block.

tools/build_config/Jenkinsfile
```diff
@@ -75,6 +75,22 @@ def get_default_herbert_branch(String job_name) {
...
+           "'${match[0][1]}' and '${match[1][1]}'")
+  } catch (IndexOutOfBoundsException e1) {
+    try {
```
nickbattam-tessella, line 87: a comment here noting that this is an expected exception because there should be only one label would be helpful; it's not often we want an exception to be thrown.

tools/build_config/Jenkinsfile
```diff
@@ -252,26 +272,31 @@ pipeline {
             }
 
           } else {
+            def herbert_ref = get_herbert_ref_from_labels(env.PR_LABELS,
```
nickbattam-tessella, line 275: the documentation for this function needs to note that it will return an error string if multiple labels are found.

nickbattam-tessella (changes requested):

tools/pwsh/clone_herbert_branch.ps1
```diff
@@ -17,13 +17,16 @@ $HERBERT_URL = "https://github.com/pace-neutrons/Herbert.git"
...
-  Write-And-Invoke "git -C $HERBERT_DIR fetch --all --tags"
-  Write-And-Invoke "git -C $HERBERT_DIR reset --hard ""$branch"""
+  Invoke-In-Dir -directory $HERBERT_DIR -command "git fetch --all --tags"
+  Invoke-In-Dir -directory $HERBERT_DIR -command "git reset --hard ""$branch"""
 } else {
   Write-And-Invoke "git clone $HERBERT_URL --depth 1 --branch ""$branch"" $HERBERT_DIR"
 }
 
-Write-Output "`nBuilding Herbert revision $(git -C $HERBERT_DIR rev-parse HEAD)..."
+Push-Location "$HERBERT_DIR"
```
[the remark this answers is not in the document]
hsautessella: `Invoke-In-Dir` also prints whatever it's invoking. Seeing as we're trying to print something, we'd end up printing it twice.

tools/bash/clone_herbert_branch.sh
hsautessella: Did you mean `Invoke-In-Dir`? There's no bash equivalent of `try {} finally {}` as far as I know. Will `pushd; <code> || popd; popd` work? 

These are developer scripts, so I'm not sure if it's worth doing or not. This script is only used in build pipelines, so leaving us in a separate directory is not the end of the world - the subsequent `sh '...'` commands spawn new shell sessions from the workspace root anyway

tools/pwsh/clone_herbert_branch.ps1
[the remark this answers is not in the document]
nickbattam-tessella: The choice we have is between having repetition of the "move-into-directory and execute an action" code, or seeing in the log:
```
> echo "some message {$subst}"
some message with substituted content
```
The repeated message is less likely to introduce future issues

tools/bash/clone_herbert_branch.sh
nickbattam-tessella: I did.

The problem with it leaving you in the wrong directory is more of an issue when running a local build (this used to happen regularly). 

Maybe `pushd && action && popd || popd` or simply `pushd && action; popd`

tools/bash/clone_herbert_branch.sh
```diff
@@ -25,20 +25,16 @@ if [ "${herbert_branch}" = "" ]; then
 fi
 
 if [[ -d "${HERBERT_DIR}" ]]; then
-    pushd "${HERBERT_DIR}" || exit 1
-    echo_and_run "git fetch -all --tags" &&
-    echo_and_run "git reset --hard \"${herbert_branch}\""
-    popd || exit 1
-
+    run_in_dir "git fetch --all --tags && git reset --hard \"${herbert_branch}\"" \
```
nickbattam-tessella, line 28: the line break here is confusing with so much whitespace and escaped quotes in the command. Defining the command to execute as a variable would simply the call:
```
git_cmds = "git fetch --all --tags && git reset --hard \"${herbert_branch}\"
run_in_dir  "${git_cmds }" "${HERBERT_DIR}"
```

tools/pwsh/powershell_helpers.ps1
```diff
@@ -52,10 +52,14 @@ function Invoke-In-Dir {
         C:\Users\Public\Documents
         C:\Users\Public\
 #>
-  param([string]$directory, [string]$command)
+  param([string]$directory, [string]$command, [switch]$noPrint)
```
nickbattam-tessella, line 55: I'm not sure how much value the `noPrint` parameter adds - it adds complexity for marginal gains. 
Whether it's needed or not, it is missing from the function documentation.

tools/bash/clone_herbert_branch.sh
```diff
@@ -25,20 +25,16 @@ if [ "${herbert_branch}" = "" ]; then
...
 else
     git_clone_cmd="git clone ${HERBERT_URL} --depth 1"
     git_clone_cmd+=" --branch \"${herbert_branch}\" ${HERBERT_DIR}"
     echo_and_run "${git_clone_cmd}"
 fi
 
-pushd "${HERBERT_DIR}" || exit 1
-echo -e "\nBuilding Herbert revision $(git rev-parse HEAD)..."
-popd || exit 1
+run_in_dir "echo -e \"\nBuilding Herbert revision \$(git rev-parse HEAD)...\"" \
```
nickbattam-tessella, line 36: similarly here, defining the command as a variable would make the `run_in_dir` call clearer

documentation/smg/08_setting_up_jenkins_ci.md
```diff
@@ -187,11 +187,13 @@ have completed*.
...
   - `pull_request.base.ref`: The name of the base branch of the pull request.
-  - `pull_request.labels[*].name`: The names of labels attached to the pull
-    request ticket. This will be in the form `["label1", "label2"]. These can
-    be used to trigger different behaviors depending on labels attached to a
-    pull request. In particular, these are used to specify a particular version
-    of Herbert to use in the build.
+  - `pull_request.labels[*].name`:
+    The names of labels attached to the pull request ticket.
+    This will be in the form `["label1", "label2"].
+    These can be used to trigger different behaviors depending on labels
```
nickbattam-tessella, line 193: This comment is misleading. We use the labels for exactly one thing -- the point you make on the next line.

These two lines could be reduced to something more like:
`The inclusion of a label matching the pattern "Herbert_<tagOrBranch>" specifies a particular tag or branch of Herbert to using in the build, otherwise Herbert master is used.`

tools/bash/bash_helpers.sh
```diff
@@ -5,6 +5,10 @@ function echo_and_run {
   eval "$1"
 }
 
+function run_in_dir {
+  (cd "$2" && eval "$1")
```
nickbattam-tessella, line 9: commands being executed via `run_in_dir` were previously executed with `echo_and_run` so we've lost the console cmd output.

tools/bash/clone_herbert_branch.sh
hsautessella: This will work, but I don't know if it's any clearer
```bash
print_herb_rev_cmd="echo -e \"\nBuilding Herbert revision"
run_in_dir "${print_herb_rev_cmd} \$(git rev-parse HEAD)...\"" "${HERBERT_DIR}"
```

The `git rev-parse` needs to be executed in the `run_in_dir` function, otherwise it's going to execute it within the Horace repository.

nickbattam-tessella: The other option is to 
```
echo -e "Build Herbert revision \c" \
  && run_in_dir "$(git rev-parse HEAD)..." "${HERBERT_DIR}"
```

tools/pwsh/powershell_helpers.ps1
```diff
@@ -52,7 +52,7 @@ function Invoke-In-Dir {
...
```
nickbattam-tessella, line 55: The was left after the rest of the `noPrint` flag code was removed

tools/build_config/Jenkinsfile
```diff
@@ -281,13 +281,18 @@ pipeline {
             }
 
           } else {
+            if (env.PR_LABELS && "${env.PR_LABELS}" != "[]") {
```
nickbattam-tessella, line 284: The second clause in unnecessary; an empty list of labels is as important as a populated one for logging. Since the default (empty) case gives a non-empty string that distinguishes it from the non-labels-passed (i.e non-PR) pipeline case trapped by the first clause.

nickbattam-tessella (approved):

hsautessella:
Failed tests are `gen_sqw` parpool tests on Windows. The tests are timing out, I think this is due to the heavy load the Windows machine is under at the moment - I can't see how this PR could have caused that. Happy for me to merge @nickbattam-tessella?

nickbattam-tessella:
Happy to merge; nothing we've changed in the build pipelines can have affected those tests.

PR #395, opened 2020-10-07
Do not default to Visual Studio 2017 in build.ps1

### LINKED ISSUE #356: Do not default to Visual Studio 2017 in build.ps1
The PowerShell build script will default to requests CMake to build with Visual Studio 2017, even if VS 2017 is not installed.

We should not pass a generator (`-G` flag) to CMake if the `-vs_version` parameter is not passed to the build script. We should let CMake choose the compiler in this case.

There is a `New-CMake-Generator-Command` function in the `build.ps1` script, this needs to return an empty string if no `-vs_version` parameter is given to the script.

- [ ] Herbert https://github.com/pace-neutrons/Herbert/issues/225
- [ ] Horace

hsautessella:
Allow CMake to pick the default compiler, this will be the most up-to-date Visual Studio on the system. The script still allows one to
specify a specific Visual Studio version

This was causing issues when attempting to use the build script to re-build targets originally build with Visual Studio 2019. The script
would default to attempting to rebuild with 2017, which would upset CMake and cause an error.

Fixes #356

### DIFF
--- a/tools/build_config/build.ps1
+++ b/tools/build_config/build.ps1
@@ -41,7 +41,8 @@ param (
 
   # The version of Visual Studio to build with. Other Windows compilers are
   # not supported by this script. {2015, 2017, 2019}
-  # [default: 0 (use latest installed version)]
+  # [default: use latest installed version or, if rebuilding, the version used
+  # in the previous build]
   [int][ValidateSet(2015, 2017, 2019)]
   [Alias("VS")]
   $vs_version = 0,
--- a/tools/build_config/build.ps1
+++ b/tools/build_config/build.ps1
@@ -40,10 +40,11 @@ param (
   [switch][Alias("h")]$help,
 
   # The version of Visual Studio to build with. Other Windows compilers are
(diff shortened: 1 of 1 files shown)


tools/build_config/build.ps1
```diff
@@ -40,10 +40,11 @@ param (
   [switch][Alias("h")]$help,
 
   # The version of Visual Studio to build with. Other Windows compilers are
-  # not supported by this script. {2015, 2017, 2019} [default: 2017]
+  # not supported by this script. {2015, 2017, 2019}
+  # [default: 0 (use latest installed version)]
   [int][ValidateSet(2015, 2017, 2019)]
   [Alias("VS")]
-  $vs_version = 2017,
+  $vs_version = 0,
```
nickbattam-tessella, line 47: As you have mentioned `0` in the "default" comment it needs to be added to the `ValidateSet` list, otherwise change the comment to just say "default: use latest installed version" (which may not be one of the valid set....)

nickbattam-tessella (approved):

Issue #389, opened 2020-09-30
Update put_pix for PixelData with more than one page


hsautessella:
The [`put_pix`](https://github.com/pace-neutrons/Horace/blob/master/horace_core/sqw/file_io/%40sqw_binfile_common/put_pix.m) function is used to write pixel data into a `.sqw` file. This function needs to be updated for the case where the SQW object's pixel data is paged.

There's already a weak attempt to implement paging in one branch of the function (my fault), however there is no `move_to_first_page` call and no testing. The other branches of the function will also need updating.

The `put_pix` function also allows the caller to specify a range of pixels to write. There is a conversation that needs to be had about whether that range should index into the current page of pixels, or into the full pixel array, and what the implications of both options are. 

refs #250 

**Outputs:**
- The `put_pix` function should save pixels within an SQW file correctly if there are one or more pages of data in the `PixelData` object.
- Tests guaranteeing the above

nickbattam-tessella:
The discussion point at the end of this ticket has been moved to a new ticket: #420

PR #403, opened 2020-10-09
pagify put_pix

cmarooney-stfc:
First pass by me at getting put_pix to work for file-backed paging sqw object. Responds to issue #389 .

Only code change here is to add move_to_first_page() for the existing code where Harry has already added advance(). This is for the case where the number of pixels to write < block size and write_all is true. Cases where write_all is false  or the pixel range is limited have still to be done. Pushed this PR (i) so that the existing code is completed (ii) so that the testing code also here can be reviewed.

Also added are (iii) test_pagedPixedData.m which tests the code so far. It reads (creates sqw object) signal data with visually simple data (for ease of visual comparison in the debugger) with a small page size, and writes (saves) it. Note that put_pix implements save(sqw-obj,'file'). The test reads this data back in (again creates sqw object) with large page size and then compares it with another sqw object created from the same file but using the large page size.

And (iv) an sqw data file (deterministic.sqw) with the signal data manually populated by sequences of 1:337 (the number of pixels in the file is 100337). This allows a developer to see the data and easily match it with a correct version.


_test/test_sqw/test_pagedPixelData.m
```diff
@@ -0,0 +1,95 @@
...
+        % Construct an object from a file accessor
+        f_accessor = sqw_formats_factory.instance().get_loader(obj.test_sqw_file_path);
+        obj.pix_data_from_faccess = PixelData(f_accessor);
+        % Construct an object from file accessor with small page size
+        obj.pix_data_small_page = PixelData(f_accessor, obj.small_page_size_);
+    end
+    
+    function test_put_pixel_paged(obj)
+        hc = hor_config();
+        hc.pixel_page_size = 100000;
```
hsautessella, line 58:
```suggestion
        hc.pixel_page_size = 1e5;
```
I usually prefer standard form in these cases, spares the reader from counting zeros.

_test/test_sqw/test_pagedPixelData.m
```diff
@@ -0,0 +1,95 @@
...
+    function delete(obj)
+        warning(obj.old_warn_state);
+    end
+
+    % --- Tests for in-memory operations ---
+    function test_def(~)
+        pix_data = PixelData();
+        num_cols = 9;
+        assertEqual(pix_data.data, zeros(num_cols, 0));
+    end
```
hsautessella, line 91: There's a very similar test to this in `test_PixelData`, we can probably delete this one.

_test/test_sqw/test_pagedPixelData.m
```diff
@@ -0,0 +1,95 @@
+classdef test_pagedPixelData < TestCase
```
hsautessella, line 1: I think these tests belong in `test_sqw_file/`. The intention of this PR's tests is to verify that when we call `save(sqw_obj, ...)`, if `sqw_obj` contains paged pixels, that we save all those pixels. We're testing `save` performs the correct actions on the `PixelData` object, not testing the `PixelData` functions.

I would've said the most sensible place for the test you've written would be in `test_sqw_file_read_write.`, but unfortunately that's an old-style test (non-class based) and I'd be hesitant to add to that. My recommendation would be to create a new test class (something like `test_read_then_write`) in `test_sqw_file/` and put your new tests in there.

_test/test_sqw/test_pagedPixelData.m
```diff
@@ -0,0 +1,95 @@
...
+
+        % Construct an object from raw data
+        obj.pixel_data_obj = PixelData(obj.raw_pix_data);
+        % Construct an object from a file
+        obj.pix_data_from_file = PixelData(obj.test_sqw_file_path);
+        % Construct an object from a file accessor
+        f_accessor = sqw_formats_factory.instance().get_loader(obj.test_sqw_file_path);
+        obj.pix_data_from_faccess = PixelData(f_accessor);
+        % Construct an object from file accessor with small page size
+        obj.pix_data_small_page = PixelData(f_accessor, obj.small_page_size_);
```
hsautessella, line 53: Are any of these attributes used anywhere?

hsautessella (changes requested):

_test/test_sqw/test_pagedPixelData.m
cmarooney-stfc: Very good point - will do

_test/test_sqw/test_pagedPixelData.m
cmarooney-stfc: This was copy/pasted from the original test_PixelData and was left in by mistake - will delete

_test/test_sqw/test_pagedPixelData.m
cmarooney-stfc: ok will do

_test/test_sqw/test_pagedPixelData.m
cmarooney-stfc: Again, copy/pasted from test_PixelData. In this case I wondered if there was a standard pro forma for constructed objects in this area, and they should be available for adding future tests. But evidently not and I can delete.

horace_core/sqw/file_io/@sqw_binfile_common/put_pix.m
```diff
@@ -167,6 +167,7 @@
     npix_to_write = npix_hi-npix_lo+1;
```
nickbattam-tessella, line 167: The if/else here is switching on block not page-size so the else block may need paging too,

Two specific issues:
- are the start and end pixels whole dataset index or in-page pixel indexes, and does that match the expected arguments for `get_pixels()`
- if the chunk of pixels being written is larger than a single page this block/page loop will need to be rewritten to handle looping both

cmarooney-stfc: As noted in the opening comment: "Cases where write_all is false or the pixel range is limited have still to be done. " But thanks for the specifics. Note from Harry in #389 "There is a conversation that needs to be had about whether that range should index into the current page of pixels, or into the full pixel array, and what the implications of both options are." I have pursued this in the debugger but am still unclear about the use-case.

hsautessella:
Happily `put_pix` did not need any updating for the case where `mem_chunk_size` (or `block_size`) was less than the pixel block size. Now `PixelData.get_pixels` uses absolute indexing, the code works as it is

nickbattam-tessella (approved):

PR #409, opened 2020-10-15
Write ADR for pixel data absolute/relative indexing

### LINKED ISSUE #408: Add ADR for absolute/paged indexing design decision for pixel data
Add an ADR detailing design decisions for absolute and relative (paged) indexing for pixel data.

In essence, the decision is to have `pix.get_*` methods use absolute indexing, whilst the attributes e.g. `pix.signal`, `pix.variance` will use relative indexing.

**Outputs:**
- An ADR recording design decisions for indexing into the `PixelData` object

hsautessella:
Adds an ARD record for the decisions made r.e. absolute/relative indexing into an SQW object's pixel array.

Fixes #408

### DIFF
--- a/documentation/adr/0016-use-double-array-in-memory-for-pixel-data.md
+++ b/documentation/adr/0016-use-double-array-in-memory-for-pixel-data.md
@@ -13,7 +13,8 @@ Accepted
 
 Pixel data will be stored on disk as single precision (float32) data.
 
-MATLAB's default numeric data type is `double` (float64). Mixing `single` and `double` data in calculations typically produces as `single` result.
+MATLAB's default numeric data type is `double` (float64).
+Mixing `single` and `double` data in calculations produces a `single` result.
 
 Options are to:
 
--- a/documentation/adr/0016-use-double-array-in-memory-for-pixel-data.md
+++ b/documentation/adr/0016-use-double-array-in-memory-for-pixel-data.md
@@ -1,4 +1,5 @@
-[<-previous](0015-store-pixel-data-in-single-precision.md) | next->
+[<-previous](0015-store-pixel-data-in-single-precision.md) |
+[next->](0017-separate-absolute-and-relative-indexing-APIs-in-pixel-array.md)
 
 # 16 - Use double array for in-memory Pixel data
 
@@ -26,15 +27,11 @@ Significant performance improvements can be seen in some algorithms when perform
 
 Pixel binning is performed in both the MATLAB and C++ code. These routines do not create a consistent binning resulting from rounding errors in calculated data.
 
-
-
 ## Decision
 
-Pixel data will stored internally as an array of MATLAB doubles and converted on read or write. 
+Pixel data will stored internally as an array of MATLAB doubles and converted on read or write.
 
 Access to the data through the PixelData class API will return `double` values.
-
- 
 
 ## Consequences
 
@@ -43,4 +40,3 @@ Access to the data through the PixelData class API will return `double` values.
 - Calculated data values will be truncated when written to file.
 - The recomputation of pixel coordinates by different routines in `TobyFit` and `gen_sqw` means rounding errors may result in pixels to be binned differently -- these differences should be considered and if possible eliminated.
 - There is no requirement to maintain `single` and `double` precision implementations of most toolkit functions.
-
--- a/documentation/adr/0016-use-double-array-in-memory-for-pixel-data.md
+++ b/documentation/adr/0016-use-double-array-in-memory-for-pixel-data.md
@@ -25,7 +25,8 @@ Long chains of function calls are executed within some algorithms (e.g. TobyFit)
 
 Significant performance improvements can be seen in some algorithms when performed on `single` precision rather than `double` data..
 
-Pixel binning is performed in both the MATLAB and C++ code. These routines do not create a consistent binning resulting from rounding errors in calculated data.
+Pixel binning can be performed in both MATLAB and C++ code.
+The two routines can result in inconsistent binning due to rounding errors.
 
 ## Decision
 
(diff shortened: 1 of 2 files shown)


nickbattam-tessella (approved):

documentation/adr/0017-separate-absolute-and-relative-indexing-APIs-in-pixel-array.md
```diff
@@ -0,0 +1,70 @@
...
+    Similarly, to retrieve a range of data from particular pixel array fields:
+
+    ```matlab
+    pixels.get_data(100:200, {'signal', 'variance'})
+    ```
+
+2. **Perform relative indexing using attribute:**
+
+    Obtaining pixel data using an attribute will return just the data for the
+    currently cached page.
```
g5t, line 52: The [`PixelData` help text](https://github.com/pace-neutrons/Horace/blob/efb681809a12a590609eccec62dd1f94632f1612/horace_core/sqw/PixelData/%40PixelData/PixelData.m#L19) says
> No pixel data will be loaded from the file on construction.

Will the relative indexing silently load a page of data if the `PixelData` object's cache is empty? Or would it be better to issue an error?

hsautessella: Data will be silently loaded when a getter is called

documentation/adr/0017-separate-absolute-and-relative-indexing-APIs-in-pixel-array.md
```diff
@@ -0,0 +1,70 @@
...
+## Consequences
+
+- Users and developers will need to be made aware of the distinction between
+the two methods of indexing.
+This should be done through documentation.
+
+- When using attributes (e.g. `.signal`),
+it should usually be accompanied by a `while has_more()` loop,
+looping through pages.
+The `get_` methods however, will usually not require a `while` loop.
```
cmarooney-stfc, line 70: ... at the point where they are called. Of course these has_more/advance' constructions will still be needed inside the implementations of the get_ methods.

Maybe stating the obvious, but once we have said get_ methods do not require while loops, it opens the door to misunderstanding of what the implementation requires

documentation/adr/0017-separate-absolute-and-relative-indexing-APIs-in-pixel-array.md
```diff
@@ -0,0 +1,70 @@
...
+
+    Obtaining pixel data using an attribute will return just the data for the
+    currently cached page.
+    Hence, indexing into these attributes will be relative.
+    For example, the following will retrieve the signal values of pixels 10-20
+    in the currently cached page:
+
+    ```matlab
+    pixels.signal(10:20)
+    ```
```
cmarooney-stfc, line 59: There is nothing here on the need for put_ methods - writing data into the structures and onto permanent disk. At least worth stating that this will need to be addressed.

documentation/adr/0016-use-double-array-in-memory-for-pixel-data.md
```diff
@@ -26,21 +27,16 @@ Significant performance improvements can be seen in some algorithms when perform
 
 Pixel binning is performed in both the MATLAB and C++ code. These routines do not create a consistent binning resulting from rounding errors in calculated data.
```
cmarooney-stfc, line 29: As it stands, I am in doubt whether this is a misprint for "resulting IN rounding errors". Could we rephrase to make it clearer that the inconsistent binning is a result of rounding errors?

documentation/adr/0016-use-double-array-in-memory-for-pixel-data.md
```diff
@@ -26,21 +27,16 @@ Significant performance improvements can be seen in some algorithms when perform
```
cmarooney-stfc, line 27: at line 16, "as" is not needed.

documentation/adr/0017-separate-absolute-and-relative-indexing-APIs-in-pixel-array.md
cmarooney-stfc: Checking I have understood correctly. Harry has said that data will be silently loaded if a get_ method is called. Greg has asked if data will be silently loaded if the attribute accessed directly by index range e.g signal(1:100). Guessing that this will not load a page but produce an error if signal etc have not been loaded. These seem to be two different scenarios, maybe need separate texts.

hsautessella: Sorry, I probably wasn't clear enough. By "getter" I meant a `get_` method or an attribute (the attributes are implemented in Matlab as `get.<attribute>` functions, so they are technically getters as well).

Constructing a `PixelData` object from a file, then calling `pix.signal(1:100)` will get you the signal at indices 1 to 100 from page 1. But the data is only loaded from file on the `pix.signal` call.

documentation/adr/0017-separate-absolute-and-relative-indexing-APIs-in-pixel-array.md
hsautessella: There are no public `put_` methods on the `PixelData` objects that write to files. I don't think there's a requirement to be able to write pixel data to file without the rest of the accompanying SQW object. It's the responsibility of the IO functions to write the `PixelData` object's underlying data into the SQW file.

As for `put_` (or `set_`?) methods that sets data within objects, I think you're probably right, we should mention those. At the moment there are no `set_` equivalents of `get_pixels` or `get_data` - we haven't needed them. Data can be set using the attributes (relative indexing), e.g. `pix.signal(1:10) = 0`.

PR #412, opened 2020-10-16
Make PixelData.get_pixels retrieve data using absolute index

### LINKED ISSUE #270: Add pixel index getters to PixelData API
Add methods `get_data_at_index(fields, indices)` and `get_pix_at_index(indices)` that will pull data from the file according to real pixel index in the file, rather than pixel index in the current page.

hsautessella:
This PR changes the behaviour of `PixelData.get_pixels` such that it indexes into the full, file-backed, pixel array. The previous behaviour was to index the currently cached page of pixels (which may or may not have been the full array). These changes are in line with the ADR written in https://github.com/pace-neutrons/Horace/issues/409.

To aid in the new implementation of `get_pixels` I've added the `PixelData.move_to_page` method. This allows you to move to an arbitrary page of the pixel data, rather than having to call `.advance()` n times.

I've also changed the behaviour of `.advance` so that it does not immediately load the next page of pixels. The pixels are not loaded until data is requested from that page. This makes it cheap to loop through pixel pages if you don't necessarily need data from every page.

Fixes #270

### DIFF
--- a/_test/test_sqw/test_PixelData.m
+++ b/_test/test_sqw/test_PixelData.m
@@ -1171,35 +1171,22 @@ methods
         assertEqual(pix_chunk.data, concatenate_pixel_pages(pix));
     end
 
-    % function test_get_abs_pix_range_throws_if_end_idx_lt_start_idx(~)
-    %     pix = PixelData();
-    %     f = @() pix.get_abs_pix_range(10, 9);
-    %     assertExceptionThrown(f, 'PIXELDATA:get_abs_pix_range');
-    % end
-
-    % function test_get_abs_pix_range_throws_invalid_arg_if_args_not_int_gt_0(~)
-    %     pix = PixelData();
-    %     f = @() pix.get_abs_pix_range(1.5, 10);
-    %     assertExceptionThrown(f, 'MATLAB:InputParser:ArgumentFailedValidation');
-    % end
-
-    function test_get_abs_pix_range_pages_pixels_if_page_size_is_exceeded(obj)
+    function test_get_abs_pix_range_reorders_output_according_to_indices(obj)
         num_pix = 30;
         data = rand(PixelData.DEFAULT_NUM_PIX_FIELDS, num_pix);
         npix_in_page = 11;
-
-        config_cleanup = set_temporary_config_options(...
-            hor_config(), 'pixel_page_size', obj.BYTES_IN_PIXEL*npix_in_page);
-
         pix = obj.get_pix_with_fake_faccess(data, npix_in_page);
-        pix_chunk = pix.get_abs_pix_range(2:num_pix - 1);
 
-        assertTrue(pix_chunk.page_size < pix_chunk.num_pixels);
+        rand_order = randperm(num_pix);
+        pix_out = pix.get_abs_pix_range(rand_order);
 
-        raw_original_pix = concatenate_pixel_pages(pix);
-        raw_pix_chunk = concatenate_pixel_pages(pix_chunk);
-        assertElementsAlmostEqual(...
-            raw_pix_chunk, raw_original_pix(:, 2:num_pix - 1), 'relative', 1e-7);
+        assertEqual(pix_out.data, data(:, rand_order));
+    end
+
+    function test_get_abs_pix_range_throws_invalid_arg_if_indices_not_vector(~)
+        pix = PixelData();
+        f = @() pix.get_abs_pix_range(ones(2, 2));
+        assertExceptionThrown(f, 'MATLAB:InputParser:ArgumentFailedValidation');
     end
 
     % -- Helpers --
--- a/_test/test_sqw/test_PixelData.m
+++ b/_test/test_sqw/test_PixelData.m
@@ -1200,6 +1200,40 @@ methods
         assertExceptionThrown(f, 'PIXELDATA:get_abs_pix_range');
     end
 
+    function test_get_abs_pix_range_throws_if_an_idx_lt_1_with_paged_pix(obj)
+        num_pix = 30;
+        data = rand(PixelData.DEFAULT_NUM_PIX_FIELDS, num_pix);
+        npix_in_page = 11;
+        pix = obj.get_pix_with_fake_faccess(data, npix_in_page);
+
+        idx_array = -1:20;
+        f = @() pix.get_abs_pix_range(idx_array);
+        assertExceptionThrown(f, 'MATLAB:InputParser:ArgumentFailedValidation');
+    end
+
+    function test_paged_pix_get_abs_pix_range_can_be_called_with_a_logical(obj)
+        num_pix = 30;
+        data = rand(PixelData.DEFAULT_NUM_PIX_FIELDS, num_pix);
+        npix_in_page = 11;
+        pix = obj.get_pix_with_fake_faccess(data, npix_in_page);
+
+        logical_array = logical(randi([0, 1], [1, 10]));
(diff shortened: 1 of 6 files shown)


_test/test_sqw/test_PixelData.m
```diff
@@ -1113,6 +1114,208 @@ function test_PIXELDATA_raised_if_mem_alloc_argument_is_not_scalar(~)
...
+        start_idx = 9;
+        end_idx = 23;
+
+        pix = obj.get_pix_with_fake_faccess(data, npix_in_page);
+        pix_chunk = pix.get_pixels(start_idx:end_idx);
+
+        assertEqual(pix_chunk.data, data(:, start_idx:end_idx));
+    end
+
+    function test_get_pixels_retrieves_correct_data_at_page_boundary(obj)
```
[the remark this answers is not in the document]
nickbattam-tessella: I think this is logically a different test to the `1:num_pixels` test below

_test/test_sqw/test_PixelData.m
```diff
@@ -1113,6 +1114,208 @@ function test_PIXELDATA_raised_if_mem_alloc_argument_is_not_scalar(~)
...
+        data = rand(PixelData.DEFAULT_NUM_PIX_FIELDS, num_pix);
+        npix_in_page = 11;
+
+        pix = obj.get_pix_with_fake_faccess(data, npix_in_page);
+        pix_chunk = pix.get_pixels(1:num_pix);
+
+        assertEqual(pix_chunk.data, concatenate_pixel_pages(pix));
+    end
+
+    function test_get_pixels_reorders_output_according_to_indices(obj)
```
nickbattam-tessella, line 1222: there's a lot of MATLAB magic wrapped up in this test; `randperm`, out-of-order indexing in the `data(:, rand_order)` in the assert.

Is there a way of recasting the test to what is going on is more explicit than:
```
matlab-magic
assertEqual(actual, data(matlab-magic))
```

_test/test_sqw/test_PixelData.m
```diff
@@ -1113,6 +1114,208 @@ function test_PIXELDATA_raised_if_mem_alloc_argument_is_not_scalar(~)
...
+        data = rand(PixelData.DEFAULT_NUM_PIX_FIELDS, num_pix);
+        npix_in_page = 11;
+        pix = obj.get_pix_with_fake_faccess(data, npix_in_page);
+
+        idx_array = 25:35;
+        f = @() pix.get_pixels(idx_array);
+        assertExceptionThrown(f, 'PIXELDATA:get_pixels');
+    end
+
+    function test_get_pixels_throws_if_an_idx_lt_1_with_paged_pix(obj)
```
nickbattam-tessella, line 1251: why is this "with_paged_pix"? `pix.get_pixels(-1)` shouldn't work for an in-memory object either.

_test/test_sqw/test_PixelData.m
```diff
@@ -1113,6 +1114,208 @@ function test_PIXELDATA_raised_if_mem_alloc_argument_is_not_scalar(~)
...
+    end
+
+    function test_get_pixels_throws_if_indices_not_positive_int(~)
+        pix = PixelData();
+        idx_array = 1:0.1:5;
+        f = @() pix.get_pixels(idx_array);
+        assertExceptionThrown(f, 'MATLAB:InputParser:ArgumentFailedValidation');
+    end
+
+    function test_paged_pix_get_pixels_can_be_called_with_a_logical(obj)
```
nickbattam-tessella, line 1269: are there additional test cases for an expected failure (or perhaps just use first n-pixels if `len(logical) > len(data)`

_test/test_sqw/test_PixelData.m
```diff
@@ -1113,6 +1114,208 @@ function test_PIXELDATA_raised_if_mem_alloc_argument_is_not_scalar(~)
...
+        npix_in_page = 11;
+        pix = obj.get_pix_with_fake_faccess(data, npix_in_page);
+
+        idx_array = cat(2, randperm(num_pix), randperm(num_pix));
+
+        pix_chunk = pix.get_pixels(idx_array);
+        assertEqual(pix_chunk.data, data(:, idx_array));
+    end
+
+    function test_pg_size_is_correct_after_advance(obj)
```
nickbattam-tessella, line 1303: is the test here that "page size reports size of partially full final page" rather than "correct after advance"

horace_core/sqw/PixelData/@PixelData/get_pixels.m
```diff
@@ -0,0 +1,90 @@
+function pix_out = get_pixels(obj, pix_indices)
```
nickbattam-tessella, line 1: For clarity should we rename these arguments `abs_pix_indices`, or those in relative-index calls `page_pix_indieces`

horace_core/sqw/PixelData/@PixelData/move_to_page.m
```diff
@@ -0,0 +1,39 @@
...
+
+    if page_number > obj.get_num_pages_
+        error('PIXELDATA:move_to_page', ...
+              'Cannot advance to page %i only %i pages of data found.', ...
+              page_number, obj.get_num_pages_);
+    end
+end
+
+
+function is = is_scalar_positive_int(number)
```
nickbattam-tessella, line 37: should the index validators more into a utility folder? we have two in this PR, these may duplicate argument validation we have already implemented and are likely to use again as we move to using `inputParser` for all routines.

while some specialist validation may be required in some functions, the `is_logical_or_vector_of_ints` and `is_positive_integer` are likely to be repeated

_test/test_sqw/test_PixelData.m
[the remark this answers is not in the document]
hsautessella: The test above this one (`test_get_pixels_retrieves_data_at_absolute_index`) tests crossing a page boundary

horace_core/sqw/PixelData/@PixelData/move_to_page.m
nickbattam-tessella: that would better be done as a TechDebt story after this PR is completed as it requires a review of other inputParsers and changes to code unrelated to PixelData paging

hsautessella: My feeling would be the function would be better placed in Herbert anyway - there are some similar utility functions there.

horace_core/sqw/PixelData/@PixelData/get_pixels.m
```diff
@@ -0,0 +1,101 @@
...
+%                    indices, in the given order, in the returned PixelData
+%                    object.
+%                    If a vector of logicals, keep pixels where the logical
+%                    array index is 1 and remove pixels where it's 0.
+%
+% Output:
+% -------
+%   pix_out        Another PixelData object containing only the pixels
+%                  specified in the abs_pix_indices argument.
+%                  This PixelData object will hold all its pixels in memory and
```
nickbattam-tessella, line 29: This is only true as a consequence of the default constructor and will be changing.

Unless this information is absolutely critical here I suggest removing it as it will be incorrect when a change is made to another piece of code.

horace_core/sqw/PixelData/@PixelData/move_to_page.m
hsautessella: https://github.com/pace-neutrons/Horace/issues/414 opened to address this

nickbattam-tessella (approved):

PR #415, opened 2020-10-19
Update PixelData.get data to use absolute indexing

### LINKED ISSUE #413: Update PixelData.get_data to use absolute indexing
The `PixelData.get_data` method should index absolutely into the pixel array, rather than indexing into the current page. The method should be updated in line with the behaviour described in ADR written in #409.

A similar behaviour was implemented for `PixelData.get_pixels` in #412.

**Outputs:**
- The `PixelData.get_data` method should return data from the absolute index into the pixel array

hsautessella:
This PR updates `PixelData.get_data` to use absolute indexing rather than relative (to page) indexing.

This brings `get_data`'s API up-to-date with `get_pixels` and the API described in the ADR added in https://github.com/pace-neutrons/Horace/pull/409.

Fixes #413

### DIFF
--- a/horace_core/sqw/PixelData/@PixelData/get_data.m
+++ b/horace_core/sqw/PixelData/@PixelData/get_data.m
@@ -1,4 +1,4 @@
-function data = get_data(obj, pix_fields, abs_pix_indices)
+function data = get_data(obj, pix_fields, varargin)
 % Retrive data for a field, or fields, for the given pixel indices in
 % the current page. If no pixel indices are given, all pixels in the
 % current page are returned.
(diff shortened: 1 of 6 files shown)


horace_core/sqw/PixelData/@PixelData/get_data.m
```diff
@@ -0,0 +1,130 @@
...
+%   pix_fields       The name of a field, or a cell array of field names
+%   abs_pix_indices  The pixel indices to retrieve, if not given, get full range
+%
+[pix_fields, abs_pix_indices] = parse_args(obj, pix_fields, varargin{:});
+field_indices = cell2mat(obj.FIELD_INDEX_MAP_.values(pix_fields));
+
+if obj.is_file_backed_()
+
+    base_pg_size = obj.max_page_size_;
+    if abs_pix_indices == -1
```
nickbattam-tessella, line 28: you're using `-1` as a magic number across this file for "no pixel indexes". A named constant would be clearer.

nickbattam-tessella (approved):

PR #416, opened 2020-10-20
Paged algorithm API ADR

### LINKED ISSUE #410: Add ADR for paged file algorithm updates
The introduction of support for large data files (via a paged PixelData object) requires updates to all SQW algorithms.

A number of algorithms have both algorithm(sqw) and algorithm(filename) APIs, others only exist in algorithm(sqw) form.

Support for paged data will require each of these to be updated to work correctly, but this migration will be performed in an incremental fashion. A strategy is required to ensure that
i) the yet to be updated algorithms continue to work
ii) the updated algorithms work identically on paged/unpaged data
iii) the updated algorithms work identically when called with SQW object and filename

nickbattam-tessella:
Document the meeting and decisions reached regarding the implementation of a standard API for algorithms and how that will work with paging.

Resolves #410

### DIFF
--- a/documentation/adr/0018-algorithm-migration-to-paged-files.md
+++ b/documentation/adr/0018-algorithm-migration-to-paged-files.md
@@ -1,4 +1,4 @@
-[<-previous](0015-store-pixel-data-in-single-precision.md) | next->
+[<-previous](0017-separate-absolute-and-relative-indexing-APIs-in-pixel-array.md) | next->
 
 # 18 - Algorithm migration to paged implementations
 
@@ -16,8 +16,9 @@ The codebase includes two different `cut()` methods:
 
 The on-disk implementation is significantly slower and users are aware of this.
 If a 'too large' file is operated on in-memory, an OutOfMemory Error will be thrown and the process will fail.
-The typical use-case for cut is that a initial cut is made from the complete dataset in a file into an object which is small and fits in memory.
-Subsequent cuts are in-memory object to in-memory object.
+A typical use-case for `cut` consists of taking an initial cut from a large dataset held in a file
+and storing this cut in memory as an `sqw` or `dnd` object.
+Subsequent cuts are in-memory object to the in-memory object.
 
 As a consequence:
  - users are required to develop (and test) separate scripts for each API
@@ -29,10 +30,9 @@ and the attempts were abandoned.
 
 One of driving forces for the current PACE project is the support of large datasets,
 which are expected to be increasingly common and of increasing size.
-The new SQW object hides all choice between in-memory and on-disk operation through
-the use of a paged `PixelData` object.
+The new SQW object has been designed around a paged `PixelData` object which holds data in memory, up to a configurable page size, and pages additional data from disk as it is required. Under this model:
 
- - provides identical API for small and large datasets on small and large computers
+ - an identical API is available for large and small datasets on small and large computers
  - the switch between in-memory and on-disk (paged) operation is automatic as dataset grows
  - the page size is a configuration parameter that is intended to be user configurable for the machines and datasets they’re using:
     - a large page size will offer better performance if the whole dataset fits in a single page, 
@@ -41,16 +41,17 @@ the use of a paged `PixelData` object.
 - implementation of a single `cut()` method (or function) that operates on an SQW object (either `this` or one passed as an argument) operating through the public API and agnostic to the size of the datasets.
 
 There are several algorithms that need to be updated to support large datasets.
-Any updates must be made in such a way that algorithms function as they currently do on "small" datasets
-until a point at which all algorithms are updated.
-During the migration it is also necessary that scripts using a subset of the updated algorithms
-can be created to perform end-to-end testing.
-These should not require code changes or alternate APIs.
+Any updates must be made in such a way that algorithms function
+as they currently do on "small" datasets.
+
+During the migration scripts which make use of a
+subset of the updated algorithms may be required to perform end-to-end testing with paged data.
+These should not require code changes or the use of alternate APIs.
 
 
 ## Decision
 
-- A new API for SQW object will be created.
+- A new `page_size` argument for will be added to the SQW object will be created.
 The SQW class will to default to in memory operation (i.e. no paging)
 if a page size is not passed explicitly to the constructor
 until all code has been updated to support paged operation
@@ -95,15 +96,19 @@ the performance drop will happen at the point the dataset exceeds page size.
 This switch will present as a sudden, unexpected performance drop.
 Display of a console message will aid users in understanding the change.
 - Paging will not be generally enabled.
-All SQW object will continue to work with existing (unmodified) algorithms
-- Paging can be switched “on” simply supporting developer activity
+All SQW object will continue to work with existing (unmodified) algorithms.
+- Paging can be switched “on” simply supporting developer activity.
 - If paging is enabled, i.e. the filename API used, and the SQW file is less that page sized,
-the operation will be performed in memory giving improved performance
+the operation will be performed in memory giving improved performance.
 - A single implementation of each algorithm will be created that will work for paged and in-memory data.
-- The paged SQW file created by the filename API must not be returned from the algorithm call
-to prevent "page leakage" and unexpected failures elsewhere in the application
+- The paged SQW file created internally by an algorithm's filename API must not be returned from the algorithm call.
+Doing so would "leak" paged data objects and result in unexpected failures elsewhere in the application.
 - Existing algorithms will continue to work on **unpaged data**.
-The behaviour if executed on **paged** data is undefined (should this throw an error?)
+The behaviour if they are executed on **paged** data is undefined.
+
+This situation will only arise if a paged `sqw` object is explicitly created by a user
+and passed in, or as a result of page "leakage" resulting from
+an error in the implementation of an algorithm.
 - Once all algorithms are updated, the default behaviour of SQW will be changed
 from unpaged to paged. 
 Any code in the wrapper methods removing paged intermediate objects may be dropped.
--- a/documentation/adr/0018-algorithm-migration-to-paged-files.md
+++ b/documentation/adr/0018-algorithm-migration-to-paged-files.md
@@ -51,7 +51,7 @@ These should not require code changes or the use of alternate APIs.
 
 ## Decision
 
-- A new `page_size` argument for will be added to the SQW object.
+- A new `pixel_page_size` argument for will be added to the SQW object.
 The SQW class will to default to in memory operation (i.e. no paging)
 if a page size is not passed explicitly to the constructor
 until all code has been updated to support paged operation,
@@ -61,7 +61,7 @@ until all code has been updated to support paged operation,
 s = sqw(filename)
 
 % paged with specified page size
-s = sqw(filename, page_size)
+s = sqw(filename, 'pixel_page_size', page_size)
 ```
 - a warning message will be output when operations are performed
   on dataset that requires paging to enable a user to understand the performance drop,
--- a/documentation/adr/0018-algorithm-migration-to-paged-files.md
+++ b/documentation/adr/0018-algorithm-migration-to-paged-files.md
@@ -51,10 +51,10 @@ These should not require code changes or the use of alternate APIs.
 
 ## Decision
 
-- A new `page_size` argument for will be added to the SQW object will be created.
+- A new `page_size` argument for will be added to the SQW object.
 The SQW class will to default to in memory operation (i.e. no paging)
 if a page size is not passed explicitly to the constructor
-until all code has been updated to support paged operation
+until all code has been updated to support paged operation,
 
 ```matlab
 % unpaged: default page size is 0/inf
@@ -64,7 +64,7 @@ s = sqw(filename)
 s = sqw(filename, page_size)
 ```
 - a warning message will be output when operations are performed
-  on dataset that require paging to enable a user to understand the performance drop
+  on dataset that requires paging to enable a user to understand the performance drop,
 
 - all algorithms will be created following a common API pattern:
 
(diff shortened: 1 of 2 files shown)


documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+The SQW class will to default to in memory operation (i.e. no paging)
+if a page size is not passed explicitly to the constructor
+until all code has been updated to support paged operation
+
+```matlab
+% unpaged: default page size is 0/inf
+s = sqw(filename)
+
+% paged with specified page size
+s = sqw(filename, page_size)
```
hsautessella, line 63: I think we should make this a named argument
```suggestion
s = sqw(filename, 'page_size', page_size)
```

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+ - users are required to develop (and test) separate scripts for each API
+ - users make a conscious choice between “in memory (fast)” and “on disk (slow)” operation.
+
+Mantid have (multiple times) attempted to present seamless in-memory/on-disk operation to their users (for a slightly different use case).
+Each time this either failed for technical reasons or because users did not embrace it,
+and the attempts were abandoned.
+
+One of driving forces for the current PACE project is the support of large datasets,
+which are expected to be increasingly common and of increasing size.
+The new SQW object hides all choice between in-memory and on-disk operation through
```
hsautessella, line 32: It doesn't yet. At the moment the pixel page size is large enough such that everything is always in memory. Is this an aspiration?

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+the use of a paged `PixelData` object.
+
+ - provides identical API for small and large datasets on small and large computers
```
hsautessella, line 35: These bullet points come out of nowhere a bit. What provides identical API ...?

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+ - the switch between in-memory and on-disk (paged) operation is automatic as dataset grows
+ - the page size is a configuration parameter that is intended to be user configurable for the machines and datasets they’re using:
+    - a large page size will offer better performance if the whole dataset fits in a single page, 
+    - a moderate page size appears optimal for datasets too large for system RAM
+    - it is hard / impossible / undesirable to calculate optimal page size on a per-machine/per-file basis
+- implementation of a single `cut()` method (or function) that operates on an SQW object (either `this` or one passed as an argument) operating through the public API and agnostic to the size of the datasets.
+
+There are several algorithms that need to be updated to support large datasets.
+Any update must be made in such a way that algorithms that the whole of Horace continues to function as it currently does on "small" datasets
```
hsautessella, line 44:
```suggestion
Any updates must be made in such a way that algorithms function as they currently do on "small" datasets
```

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+until a point at which all algorithms are updated.
```
hsautessella, line 45: Even once all algorithms are updated we still want the in-memory operations to be as efficient as they are now.
I'm not sure if this line is necessary, I'm not clear what changes in this regard when all algorithms are updated. When all algorithms are updated we can begin passing around file-backed objects, but the behaviour for small datasets should still be the same.

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+During the migration it is also necessary that scripts using a subset of the updated algorithms
+can be created to perform end-to-end testing.
```
hsautessella, line 47:
```suggestion
are created to perform end-to-end testing.
```

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+These should not require code changes or alternate APIs.
+
+
+## Decision
+
+- A new API for SQW object will be created.
```
hsautessella, line 53: A new API for SQW objects? Isn't it just a new constructor argument (`page_size`)?

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+% paged with specified page size
+s = sqw(filename, page_size)
+```
+- a warning message will be output when operations are performed
+on dataset that require paging to enable a user to understand the performance drop
+- all algorithms will be created with a common API pattern.
+The two APIs will share a common implementation,
+the filename variant creating a (paged) SQW object
+	- `algorithm(filename, outfilename)`
+perform operation on a new SQW object that  will be explicitly created as paged
```
hsautessella, line 71:
```suggestion
perform operation on a new SQW object that will be explicitly created as paged
```

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+	- `algorithm(sqw, [outfilename])`
+perform operation on existing SQW object (may be paged or unpaged)
```
hsautessella, line 73: The language in this section seems a little confused.

There will be three ways to call an algorithm operating on an SQW object:

1. Using a filename `sqw_obj = algorithm(file_name)`: this will create a paged SQW object from the given file. The algorithm will then operate on the paged SQW object. The paged SQW object will never be visible to the user. A smaller all-in-memory SQW object will be outputted.

2. Using an SQW object `algorithm(sqw_obj)`: the input SQW object will be fully in-memory and so will the output.

3. Using a filename and outputting to a file `algorithm(file_name, out_file_name)`: Equivalent to (1) except the resulting SQW object will be saved to file instead of returned as an object. This provides a method to perform an algorithm that maps a large SQW file to another large SQW file.

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+cut(infile, outfile) % disk => disk
+```
+
+
+## Consequences
+
+- The internal switch between in-memory and disk operations means
+the performance drop will happen at the point the dataset exceeds page size.
+This switch will present as a sudden, unexpected performance drop.
+Display of a console message will aid users in understanding the change.
```
hsautessella, line 96: This is true, but also the fact that users will be inputting a file name rather than an SQW object, clues them in that there may be a performance drop.

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+- Paging will not be generally enabled.
+All SQW object will continue to work with existing (unmodified) algorithms
+- Paging can be switched “on” simply supporting developer activity
+- If paging is enabled, i.e. the filename API used, and the SQW file is less that page sized,
+the operation will be performed in memory giving improved performance
+- A single implementation of each algorithm will be create that will work for paged data,
+this will then work for in-memory or file backed use-cases
```
hsautessella, line 103:
```suggestion
- A single implementation of each algorithm will be created that will work for paged and in-memory data.
```

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+- The paged SQW file created by the filename API must not be returned from the algorithm call
+to prevent "page leakage" and unexpected failures elsewhere in the application
+- Existing algorithms will continue to work on **unpaged data**.
+The behaviour if executed on **paged** data is undefined (should this throw an error?)
```
hsautessella, line 107: Not sure what you mean by this. An algorithm that has not been updated for paged data should never be using paged data. Paged `PixelData` objects are not publicly exposed, so we shouldn't run into any problems like this.

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+- Once all algorithms are updated the default behaviour of SQW will be updated
+from unpaged to paged and custom code in the wrapper methods
+removing paged intermediate objects may be dropped
```
hsautessella, line 110:
```suggestion
- Once all algorithms are updated, the default behaviour of SQW will be changed
from unpaged to paged. 
Any code in the wrapper methods removing paged intermediate objects may be dropped.
```

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
...
+- Memory errors will be raised if a cut is larger than available RAM is made
+and no `outfilename` is specified
```
hsautessella, line 112:
```suggestion
- Memory errors will be raised if a cut is larger than available RAM and no `outfilename` is specified
```

hsautessella (changes requested):
This is a good first draft.

I think we should emphasize the fact that the dual API allows use to incrementally update SQW algorithms before we make paged PixelData objects publicly accessible. If we were to make SQW paged by default, then we would need to update every SQW method and algorithm before we'd have a functioning SQW object.

Hiding the paged SQW objects from users means we can maintain a working product whilst we update the rest of the codebase to support file-backed operation.

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,115 @@
+[<-previous](0015-store-pixel-data-in-single-precision.md) | next->
```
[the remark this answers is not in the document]
hsautessella: Also has the change been made in `0017-` to point to this ADR?

documentation/adr/0018-algorithm-migration-to-paged-files.md
nickbattam-tessella: I added that as a consequence of the other strategy.

If it does happen, for example a developer calling with a paged object, or an bug in a paged algorithm implementation that leaks the paged file, we need to raise that be behaviour is undefined.

Raising an error would be a nicer solution, but would require code changes everywhere.

documentation/adr/0018-algorithm-migration-to-paged-files.md
nickbattam-tessella: Once the whole of Horace has been updated to support large data it will be possible to call algorithms with paged `sqw` objects. The use of the filename vs object API isn't what triggers this performance drop, the existence of multiple pages of data does.

documentation/adr/0018-algorithm-migration-to-paged-files.md
nickbattam-tessella: Good point, I will update the way this is expressed.

You point (2) however is incorrect: if a paged SQW object is passed (e.g. by a developer)  the algorithm will operate on in as paged data. returning a paged object.

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,120 @@
...
+and storing this cut in memory as an `sqw` or `dnd` object.
+Subsequent cuts are in-memory object to the in-memory object.
+
+As a consequence:
+ - users are required to develop (and test) separate scripts for each API
+ - users make a conscious choice between “in memory (fast)” and “on disk (slow)” operations.
+
+Mantid have (multiple times) attempted to present seamless in-memory/on-disk operation to their users (for a slightly different use case).
+Each time this either failed for technical reasons or because users did not embrace it,
+and the attempts were abandoned.
```
abuts, line 29: Small modification:
Mantid have successfully implemented paging operations in two situations: Paged instrument workspace, corresponding to workspace, generated for single experiment and MDWorkspace, functionally equivalent to Horace SQW object. Unfortunately, the success was only "technical" and in fact was the failure, at least for ISIS, as the code is not used by the target scientific community and, in the first case, has been fully removed from the code base due to parlous performance of file based operations with respect to memory operations. In the second case the reasons for failure are more subtle but analysis of the users complains points to:
 - negative user experience due to substantial performance drop at implicit change from memory-based to file-based  operations
 - steep learning curve, as the users, concerning about their scientific task are requested to specify deeply technical information substantially affecting performance (e.g. page size, and/or thread count) from the beginning. 
 - unclear and unusual interface, including the need to perform the requested operations in two steps, i.e. creating object and doing the operations over then.  Despite the later sound very minor, the number of users, who finds this way of operating the package confusing is surprising. 

Horace workflow, from other side, remains natural an popular way of analyzing the results of experiments, so should be kept as much as reasonable.  In addition to that, the idea that if one invokes file operation it may run slowly, and if it is memory operation it should be fast is very natural to any our users.

abuts (changes requested):

documentation/adr/0018-algorithm-migration-to-paged-files.md
abuts: you are forgetting the need to run algorithms in parallel and keep the same API. This means that interface should be
`result = algorithm(memory_object)`
`result = algorithm(memory_object,with_mpi)`
`result = algorithm(file_name)`
`result = algorithm(file_name,with_mpi)`
`result = algorithm(file_name,output_to_file)`
`result = algorithm(file_name,output_to_file,with_mpi)` % here result may me file-name 
file-based object in two last cases can be result but should it be....

This is also acceptable common inferface.

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,120 @@
...
+
+% paged with specified page size
+s = sqw(filename, page_size)
+```
+- a warning message will be output when operations are performed
+  on dataset that requires paging to enable a user to understand the performance drop,
+
+- all algorithms will be created following a common API pattern:
+
+  1. `obj = algorithm(filename, ...)`: internally create a paged SQW object from the given file and execute the algorithm on the paged object. The SQW will returned will be a smaller, in-memory object,
```
cmarooney-stfc, line 71: Extra "will" before "returned" should be deleted

documentation/adr/0018-algorithm-migration-to-paged-files.md
```diff
@@ -0,0 +1,120 @@
...
+
+The example APIs for `cut` would be:
+```matlab
+% Pure in-memory operation (small file only)
+s = sqw(infile) 
+slice = cut(s)
+
+% Paged operation (in-memory for small file, paged for large file)
+s = sqw(infile, page_size)
+slice = cut(x) 		 % object => object
```
cmarooney-stfc, line 84: Argument should be s

PR #432, opened 2020-11-02
Refactor of @sqw\sqw to @sqw_old\sqw_old to enable new class creation

nickbattam-tessella:
Automated refactor of sqw -> sqw_old. 

PR will test the unit tests still execute correctly before proceeding with DND rename..


cmarooney-stfc (commented):
This is quite an impressive global edit. The useful thing is that for methods with "do_stuff_to_sqw(..." type methods, the naming is included in the renamed class directory @sqw_old. So just waiting now for the build/test to finish....

PR #433, opened 2020-11-05
changes necessary to make release 3.5

abuts:
hopefully no changes will be necessary except the version


hsautessella:
Is there an argument that this should be version 4.0.0 since the new `PixelData` object is an API break?

abuts:
would disagree. PixData are internal changes nobody cares about. 4 will be when sqw object is a new type object. This would be significant and opens at least the path to compile it.

hsautessella:
OK, that's fine. As long as users aren't using pixel data directly I'm happy with incrementing the minor version 👍

PR #436, opened 2020-11-06
Add pixel page size option sqw constructor

### LINKED ISSUE #435: Add pixel page size option to sqw constructor
https://github.com/pace-neutrons/Horace/pull/416/ states the need for an additional argument in the `sqw` constructor that set the page size for file-backed pixel data. This new parameter should be passed down to the `PixelData` class constructor to create a file-backed `PixelData` object with the `sqw` instance.

**Outputs:**

- A new named argument added to the `sqw` class constructor, something similar to `pixel_page_size`.
- Use of this argument should instantiate an `sqw` instance whose pixels are contained within a paged `PixelData` object

hsautessella:
~**This PR is marked as draft since some tests may be running incorrectly. `PixelData` no longer reads the `hor_config.pixel_page_size` value to set the page size. As such, some tests that use this method to set the pixel data page size will no longer be paging data, but will likely still pass. I've got some more of these tests to track down.**~

This PR adds a `pix_pg_size` named argument to the `sqw` constructor. This allows a user of the `sqw` class to specify the page size of the underlying `PixelData` object.

By default, the `PixelData` object no longer reads `hor_config.pixel_page_size` to get the page size. The `mem_alloc` argument must be explicitly passed to the `PixelData` constructor to set the page size (the default `mem_alloc` being `realmax`).

Fixes #435


_test/test_sqw/test_noisify.m
```diff
@@ -57,37 +32,39 @@ function delete(obj)
...
         % a deterministic set of numbers 1:999 repeated. Use factor to make
         % them in range 0:1
-        factor =1/999;
+        factor = 1/999;
         % We make an sqw object with the above page size
-        sqw_obj1 = sqw(obj.test_sqw_file_full_path);
-        
+        sqw_obj1 = sqw(obj.test_sqw_file_full_path, 'pix_pg_size', pixel_page_size);
+
+        % ensure we're actually paging data
```
nickbattam-tessella, line 49: do we need a second non-paged test?

_test/test_sqw/test_PixelData.m
```diff
@@ -1091,20 +1093,6 @@ function test_has_more_is_true_after_appending_page_to_non_file_backed(obj)
         assertFalse(pix.has_more());
     end
 
-    function test_max_page_size_set_by_pixel_page_size_config_option(obj)
```
nickbattam-tessella, line 1094: this test needs to be replaced with a pair of "page size is set by constructor argument" and "page size is infinite be default".

A test demonstrating the new patterns of
```
page_size = get(hor_config, 'pixel_page_size');
s = sqw('filename', 'pixel-page-size', page_size)
```
would be a useful addition to help developers and test the config file contains a sensible, non-infinite value there.

horace_core/sqw/PixelData/@PixelData/PixelData.m
```diff
@@ -103,8 +103,9 @@
 end
 
 properties (Constant)
-    DEFAULT_NUM_PIX_FIELDS = 9;
     DATA_POINT_SIZE = 8;  % num bytes in a float
+    DEFAULT_NUM_PIX_FIELDS = 9;
+    DEFAULT_PAGE_SIZE = realmax;  % this gives no paging by default
```
nickbattam-tessella, line 108: We were using `floor(realmax)` previously. Is the existence of the fractional part important?

horace_core/sqw/file_io/@sqw_binfile_common/get_data.m
```diff
@@ -1,4 +1,4 @@
-function [data,obj] = get_data (obj,varargin)
+function [data,obj] = get_data(obj,varargin)
```
nickbattam-tessella, line 1: new page-size argument is missing from the function header

horace_core/sqw/@sqw/sqw.m
```diff
@@ -118,38 +118,10 @@
...
-        w = class(w,class_type);
-        return
-    else
-        error(mess)
-    end
-
+elseif narg == 1 && is_string(args{1})
+    w = from_file_name(args{1}, dnd_type, class_type);
+elseif narg > 1 && is_string(args{1})
+    w = from_file_name(args{1}, dnd_type, class_type, varargin{2:end});
```
nickbattam-tessella, line 124: the new page-size argument is missing from the function header

horace_core/sqw/@sqw/sqw.m
```diff
@@ -168,3 +140,46 @@
 
 end
 
+end  % function
+
+
+% -----------------------------------------------------------------------------
+function w = from_file_name(file_name, dnd_type, class_type, varargin)
+    if ~dnd_type
```
nickbattam-tessella, line 148: There are three different argument parsers handling the varargininput to the get_data, get_sqw and PixelData constructor calls. 
It may be possible to simplify these are as part of the SQW refactor work, different approaches to argument handling makes the code harder to follow.

nickbattam-tessella (changes requested):

nickbattam-tessella (changes requested):

_test/test_sqw/test_noisify.m
```diff
@@ -73,14 +67,51 @@ function test_how_noisify_works(obj)
...
+        noisy_obj = noisify(sqw_obj, noise_factor);
+
+        original_signal = sqw_obj.data.pix.signal;
+        noisy_signal = noisy_obj.data.pix.signal;
+        signal_diff = original_signal - noisy_signal;
+
+        % Fit the signal differences and check that the expected mu and sigma
+        % for the normal distribution fall within the 95% confidence interval
+        % using paramci
+        pd = fitdist(signal_diff(:), 'normal');
```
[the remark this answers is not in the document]
hsautessella: Ah, I forgot! Thanks

nickbattam-tessella (approved):

_test/test_sqw/test_PixelData.m
```diff
@@ -69,8 +63,9 @@
     end
 
     function delete(obj)
-        rmpath(fullfile(obj.this_dir, 'utils'));
-        set(hor_config, obj.old_config);
+        if ismember(fullfile(obj.this_dir, 'utils'), split(path, {':', ';'}))
```
abuts, line 66: should be 
`split(path,pathsep)`
`pathsep` -- system dependent  path separator

abuts (changes requested):

_test/test_sqw/test_PixelData.m
abuts: hor_config 
needs description of page_size after     
%   mem_chunk_size    - Maximum length of buffer array to accumulate pixels
%                       from an input file.
something like:
% page_size  -- property to define page size in experimental file based mode

horace_core/sqw/PixelData/@PixelData/PixelData.m
cmarooney-stfc: Also I would expect that the page size will be a large integer. So floor emphasizes the integer. But if looking for a max real, this will be 10^38 or 10^308 depending on the default type for realmax. It means unfeasibly large for a memory integer but do we want that or do we want a realistic upper bound on the amount of memory available, say via intmax with a suitably large integer type?

horace_core/sqw/file_io/@sqw_binfile_common/get_data.m
cmarooney-stfc: @nickbattam-tessella it looks as if it is in varargin and processed by parse_args as a keyword argument

cmarooney-stfc (commented):
@hsautessella A few comments. When you've dealt with Nick's comments, should be ok to go

horace_core/sqw/PixelData/@PixelData/PixelData.m
hsautessella: `floor(realmax) == realmax`, so there's no functional difference. 

I'm hesitant to use an integer type because of Matlab's type propagation: `int64_var + double_var = int46_var`. Meaning that whenever we do an operation with `DEFAULT_PAGE_SIZE` the result will be an integer type, which can result in some interesting bugs.

_test/test_sqw/test_PixelData.m
hsautessella: First test already exists, I've added the second.

hsautessella:
@nickbattam-tessella @cmarooney-stfc are we happy for this to go in now?

PR #445, opened 2020-11-13
Create docs_discussion.md (Do not merge)

oerc0122:


oerc0122:
Just for ease, link to formatted version: https://github.com/oerc0122/Horace/blob/patch-2/documentation/smg/docs_discussion.md

abuts (approved):
This is excellent documents for discussion.  I have my opinion on how to proceed with this, but it will be solution. The only remark is that discussion doc should be in different folder (ADR?) Tessella guys (@hsautessella ; @nickbattam-tessella) would tell better

documentation/smg/docs_discussion.md
```diff
@@ -0,0 +1,203 @@
...
+* Keep docs up-to-date when they are updated
+* No effort to update docs
+
+##### Cons:
+* May lead to damage if docs not verified properly (mitigated by multiple docs versions)
+* Difficult to check how branches would appear
+* Requires manually updating version number in Jenkins
+
+#### Build/Deploy on Horace release
+Tie docs deploy into the existing deploy pipeline such that on each major Horace release docs are released to the right tag
```
hsautessella, line 82: We use the deploy pipeline to create draft releases which developers can then manually test (see [the deployment SMG](https://github.com/pace-neutrons/Horace/blob/master/documentation/smg/12_deploying_releases_to_GitHub.md)). So deploying the docs in the same pipeline could result in the documentation for a release appearing before the actual release package.

I'm not sure what the correct solution here is, but we could set up a [GitHub webhook that pings our build servers when a release is published](https://developer.github.com/webhooks/event-payloads/#release), our servers can then deploy the docs. Although it maybe simpler for the release docs to go up before the packages.

documentation/smg/docs_discussion.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### Dev, Version
+Only have version number and dev version users look up their version
+
+##### Pros:
+* Users of old versions have no confusion on release update
+* Dev changes visible
+
+##### Cons:
+* No short cut to latest version (need to know your version)
```
hsautessella, line 125: What do you mean by latest version here? Latest release, latest git revision or master?

documentation/smg/docs_discussion.md
```diff
@@ -0,0 +1,203 @@
...
+* Discourages users staying up to date?
+
+#### Per-branch docs
```
hsautessella, line 128: I feel per-branch docs are probably overkill, which I see you've noted below. My feeling is a set of docs for each release and a set of docs for master would be sufficient.

documentation/smg/docs_discussion.md
```diff
@@ -0,0 +1,203 @@
...
+
+##### Pros:
+* Easy to force an update through
+
+##### Cons:
+* Easy to force an update through
+
+### Alternatives
+#### PR required
+Disable pushing to main
```
hsautessella, line 194: If the docs are in a separate repo or Horace, we should always be using PRs. If not only for preventing damage to master (as you say), but also for visibility of changes for the rest of the team.

hsautessella (commented):
Plenty of good ideas here, I assume the purpose of this document is only for discussion? When we've made a decision this document can be summarized (along with the decisions we've made) and put in an ADR (Architectural Decison Record) (as Alex suggested). See [our other ADRs](https://github.com/pace-neutrons/Horace/blob/master/documentation/adr/0001-record-architecture-decisions.md). 

I don't know what everyone else's opinions are, but I'm leaning more toward keeping the documentation in the Horace repository.

1. This allows one to update the documentation at the same time as the code, so they need never be out of sync.
2. If we ever want to run tests on code snippets in the documentation, we should be running those snippets against every pull request. This is much easier if those documentation code snippets are in with the source.
3. We won't need to set up and maintain additional Jenkins pipelines that build `horace-docs` pull requests. It feels much simpler to add a "doc tests" stage to our existing Jenkins pipelines building Horace pull requests.

Having a version of the docs for each release is essential I think, and then one more version for master. The "master" docs can be updated when we run our nightly builds.

PR #446, opened 2020-11-13
Create 14_documentation.md

oerc0122:
Initial commit of bare outline of documentation

### DIFF
--- a/documentation/smg/14_documentation.md
+++ b/documentation/smg/14_documentation.md
@@ -1,7 +1,7 @@
 # Documentation
 
 This document will outline the process surrounding the documentation currently housed in the `horace-docs` repository. 
-It assumes a basic knowledge of reStructured text (reST) format (https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html)
+It assumes a basic knowledge of reStructuredText (reST) format (https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html)
 especially the subset of the sphinx-doc toolset (https://www.sphinx-doc.org/en/master/)
 
 ## Documentation style guide
@@ -20,23 +20,30 @@ On Linux (Optional):
 * Make
 
 ### Building Locally
-Once you have cloned the documentation repository (https://github.com/pace-neutrons/horace-docs)
+First, clone the documentation repository (https://github.com/pace-neutrons/horace-docs)
 
     git clone https://github.com/pace-neutrons/horace-docs.git
 
-#### Windows
-With a command terminal in the root folder of the horace-docs project run:
+* Windows
+  
+  With a command terminal in the root folder of the horace-docs project run:
     
-    make.bat html
+      make.bat html
 
-#### Linux
-With `make` installed simply run:
+* Linux (With Make)
 
-    make html
+  With `make` installed simply run:
+
+      make html
     
-If you do not have `make` installed, run:
+* Linux (No Make)
 
-    sphinx-build -M html "docs" "build" 
+  If you do not have `make` installed, run:
+
+      sphinx-build -M html "docs" "build"
+
+
+Built documents will be put in the `/build/html` folder, the main file is the `index.rst` file.
 
 ## Adding new documents
 ### Create new `.rst`
@@ -61,15 +68,18 @@ The first job to add a new document is to create it.
 
 The main file in the docs folder is `index.rst` which consists of two major parts. 
 
-#### The TOCTree
+#### The TOC Tree
 
-The TOC tree describes all child rst files to be loaded for the purpose of references. It also provides the the details for the links sidebar. 
+The TOC (table of contents) tree describes all child reST files to be loaded for the purpose of references.
+It also provides the the details for the links sidebar . 
 
 To include a document and include it in the side bar, simply add the new filename (without file extension) to the toctree. 
 
 To include a document for reference, but not in the sidebar add the element to the TOCTree via:
 
     [NULL] <file-name>
+
+More info can be found https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html?highlight=toctree#table-of-contents
 
 #### Main contents list
 
--- a/documentation/smg/14_documentation.md
+++ b/documentation/smg/14_documentation.md
@@ -36,9 +36,7 @@ To include a document and include it in the side bar, simply add the new filenam
 
 To include a document for reference, but not in the sidebar add the element to the TOCTree via:
 
-    ```
     [NULL] <file-name>
-    ```
 
 #### Main contents list
 
--- a/documentation/smg/14_documentation.md
+++ b/documentation/smg/14_documentation.md
@@ -1,6 +1,6 @@
 # Documentation
 
-This document will outline the process surrounding the documentation currently housed in the `horace-docs` repository. It assumes a basic knowledge of Restructured Text (ReST) format (https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html) especially the subset of the sphinx-doc toolset (https://www.sphinx-doc.org/en/master/)
+This document will outline the process surrounding the documentation currently housed in the `horace-docs` repository. It assumes a basic knowledge of reStructured text (reST) format (https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html) especially the subset of the sphinx-doc toolset (https://www.sphinx-doc.org/en/master/)
 
 ## Documentation style guide
 
@@ -11,14 +11,14 @@ This document will outline the process surrounding the documentation currently h
 The first job to add a new document is to create it. 
 
 * The file should be added into the `docs` folder, preferably with the `.rst` suffix. 
-   * For ease the filename of the document should be the title with spaces replaced by underscores. 
+   * For ease of reference and consistency, the filename of the document should be the title with spaces replaced by underscores. 
 * It should be written using the sphinx-doc style, including interfile references through:
 
     ```
     :ref:`link-label <link_file:link_location>`
     ```
     
- * External references are written in the usual ReST format:
+ * External references are written in the usual reST format:
 
     ```
     `link-label <https:\\link-to-elsewhere>`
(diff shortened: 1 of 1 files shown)


documentation/smg/14_documentation.md
```diff
@@ -0,0 +1,45 @@
+# Documentation
+
+This document will outline the process surrounding the documentation currently housed in the `horace-docs` repository. It assumes a basic knowledge of Restructured Text (ReST) format (https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html) especially the subset of the sphinx-doc toolset (https://www.sphinx-doc.org/en/master/)
```
cmarooney-stfc, line 3: Worried that ReST will be confusing for those thinking about the REST/SOAP API alternatives. The linked document refers consistently to reStructured text with no abbreviation. Suggesting that we keep the capitalization of the full name (i.e. Sem-camel-case) and use the rst of the suffix (asis or all-caps, at your preference) as the abbreviation.

documentation/smg/14_documentation.md
```diff
@@ -0,0 +1,45 @@
...
+## Documentation style guide
+
+
+## Adding new documents
+### Create new `.rst`
+
+The first job to add a new document is to create it. 
+
+* The file should be added into the `docs` folder, preferably with the `.rst` suffix. 
+   * For ease the filename of the document should be the title with spaces replaced by underscores.
```
cmarooney-stfc, line 14: Presuming that the ease is something like "for ease of reference ... " or something; would make it clearer why we do this.

documentation/smg/14_documentation.md
```diff
@@ -0,0 +1,45 @@
...
+
+The main file in the docs folder is `index.rst` which consists of two major parts. 
+
+#### The TOCTree
+
+The TOC tree describes all child rst files to be loaded for the purpose of references. It also provides the the details for the links sidebar. 
+
+To include a document and include it in the side bar, simply add the new filename (without file extension) to the toctree. 
+
+To include a document for reference, but not in the sidebar
```
cmarooney-stfc, line 37: Presuming this line "To include... not in sidebar" refers to the next section for the main index, and I think it should preferably go after the heading. If you think it important that it should go before the heading, then please add an ellipsis ... to indicate that the meaning is continued in the following text.

documentation/smg/14_documentation.md
cmarooney-stfc: Now looking at Sphinx and they do use reST (but note the capitalization.) It may still be worth alerting the reader to the ambiguity (" ... this is reST, the restructured text markup, not REST the API.")

documentation/smg/14_documentation.md
oerc0122: Sorry, that was a trailing bit that I had left to fill in when I had decided on a syntax.

documentation/smg/14_documentation.md
oerc0122: Updated throughout

documentation/smg/14_documentation.md
oerc0122: Updated

documentation/smg/14_documentation.md
cmarooney-stfc: From talking to Nick, all this process was decided and documented in ADR 2 of pace devs. It could be useful to record that here.

documentation/smg/14_documentation.md
```diff
@@ -0,0 +1,47 @@
+# Documentation
+
+This document will outline the process surrounding the documentation currently housed in the `horace-docs` repository. It assumes a basic knowledge of reStructured text (reST) format (https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html) especially the subset of the sphinx-doc toolset (https://www.sphinx-doc.org/en/master/)
```
hsautessella, line 3: Could we split long lines like this onto multiple lines? It makes it easier to comment on/review specific sentences and also means that if we make a small change to part of the paragraph (e.g. fixing a typo) we don't get a diff for the full paragraph.

A good set of rules to follow is "[semantic line breaks](https://sembr.org/)" when it's easy to do so. The gist being: line length limit of 80 characters, always split lines at the end of a sentence and split long sentences on clauses.

abuts (changes requested):
A useful introduction to rst is provided, but I would expect finding here what scripts are used for deploying build on Jenkins, can they be used to compile help locally and what settings are needed for that (Sphinks on the path or just Python?), and how build is triggered, what hooks are there and what may be needed...  Also how to write and run tests for Matlab scripts embedded into the rst docs (may be as reference to the future tickets). 

I see TBD section, which probably means the job is incomplete. Should it be before it merged?

hsautessella (commented):
What's written so far looks good to me, just one comment on style.

oerc0122:
Basically, that stuff is undecided as yet. The pipeline and deployment is currently on a different MR to discuss how we're actually going to do them (#445 )

I will add a bit about local compilation and dependencies, you're very right on that front.

abuts (approved):
That's much better. A reference on the problem of running embedded tests in Matlab would be useful.

oerc0122:
I'm not quite sure what you mean by that.

documentation/smg/14_documentation.md
```diff
@@ -0,0 +1,80 @@
+# Documentation
+
+This document will outline the process surrounding the documentation currently housed in the `horace-docs` repository. 
+It assumes a basic knowledge of reStructured text (reST) format (https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html)
```
nickbattam-tessella, line 4: From the website it is consistently referred to as "reStructuredText" rather than "reStructured text"

documentation/smg/14_documentation.md
```diff
@@ -0,0 +1,80 @@
...
+The documentation requires the following to be installed:
+* Python (3.5+)
+* Sphinx
+* Sphinx RTD theme
+
+On Linux (Optional):
+* Make
+
+### Building Locally
+Once you have cloned the documentation repository (https://github.com/pace-neutrons/horace-docs)
```
nickbattam-tessella, line 23: There's something off with the tenses here: "Once you have cloned [...]" and then give the command _to_ clone

documentation/smg/14_documentation.md
```diff
@@ -0,0 +1,80 @@
...
+With a command terminal in the root folder of the horace-docs project run:
+    
+    make.bat html
+
+#### Linux
+With `make` installed simply run:
+
+    make html
+    
+If you do not have `make` installed, run:
```
nickbattam-tessella, line 37: The placement of the section headings makes it unclear if this is Linux only or Windows and Linux

documentation/smg/14_documentation.md
```diff
@@ -0,0 +1,80 @@
...
+
+    ```
+    `link-label <https://link-to-elsewhere>`
+    ```
+
+### Update index
+
+The main file in the docs folder is `index.rst` which consists of two major parts. 
+
+#### The TOCTree
```
nickbattam-tessella, line 64: `TOCTree` or `TOC Tree` ?

documentation/smg/14_documentation.md
```diff
@@ -0,0 +1,80 @@
...
+
+To include a document and include it in the side bar, simply add the new filename (without file extension) to the toctree. 
+
+To include a document for reference, but not in the sidebar add the element to the TOCTree via:
+
+    [NULL] <file-name>
+
+#### Main contents list
+
+The contents list is a manual table of contents linking to main documentation which will appear on the main page. Any contributions which wish to be listed in the main table of contents should appear here.
```
nickbattam-tessella, line 76: I'm unclear what this section contains. The index file in horace-docs appears to only consist of toctree elements

oerc0122: https://github.com/pace-neutrons/horace-docs/blame/main/docs/index.rst#L78

nickbattam-tessella: Might it be worth including an illustrative code fragment?

abuts:
> 
> 
> I'm not quite sure what you mean by that.

@oerc0122  -- All scripts embedded in the documentation should be runnable.  The only way to ensure this, is to test-run them automatically.  Python rst files have natural ability to do that. Can we try to enable this for Horace? The existing scripts, of course, have to be modified, but as soon as one test is run, the others is the matter of time.

oerc0122:
> > I'm not quite sure what you mean by that.
> 
> @oerc0122 -- All scripts embedded in the documentation should be runnable. The only way to ensure this, is to test-run them automatically. Python rst files have natural ability to do that. Can we try to enable this for Horace? The existing scripts, of course, have to be modified, but as soon as one test is run, the others is the matter of time.

Technically possible, but a bit of a faff, unless I'm misunderstanding what you want, we'd need to have some sort of flag/directive indicating that a block was a piece of Horace input code to be tested and then a script to extract and run these through matlab (with an appropriate build of Horace). 

If you mean the files starting with "Script_", that's slightly easier as they only contain executable code. Still a bit of a faff, but more doable.

abuts:
> Technically possible
Mantid currently does that, (e.g. https://docs.mantidproject.org/nightly/algorithms/GetEi-v2.html but this is for every algorithm) and all sample scripts in examples are runnable and automatically tested. As I understand, the marking for Python scripts is the standard for Sphinx or its extension and all what we need to do is to wrap python script to run Matlab under Python (or Python extracts code and runs Matlab over it)

oerc0122:
> > Technically possible
> 
> Mantid currently does that, (e.g. https://docs.mantidproject.org/nightly/algorithms/GetEi-v2.html but this works for every algorithm) and all sample scripts in examples are runnable and automatically tested. As I understand, the marking for Python scripts is the standard for Sphinx or its extension and all what we need to do is to wrap python script to run Matlab under Python (or Python extracts code and runs Matlab over it)

Their docs also seem to suggest there's a mantid Python API, which I don't know if horace has.

It seems it would be possible through doctest (https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html) but would require a fair amount of setup, rather than just checking they work by hand for now until that wrapper is in place.

oerc0122:
Still WIP until some of these things are set up as part of #454 but ready to be merged as a placeholder document when that is finalised.

nickbattam-tessella (approved):

PR #453, opened 2020-11-19
Add SQW Binary ops

nickbattam-tessella:
Migrate the binary ops to SQW.

Tests for SQW .op. DND operations disabled pending DND class.
All code migrated from sqw_old with minimal changes.


horace_core/sqw/@sqw/sqw.m
```diff
@@ -15,6 +15,12 @@
     end
 
     methods
+        wout = recompute_bin_data(w);
```
hsautessella, line 18: This is a private method on master.

hsautessella (commented):
Looks sensible to me. One or two things that could be tweaked, although I know the code was not originally yours. I'll leave it up to you whether it's more sensible to just leave things as they are whilst you're copying things over.

cmarooney-stfc (approved):
OK!

PR #466, opened 2020-11-27
D2D constructor

nickbattam-tessella:
First DnD constructor. 

Some functionality currently in the class will probably migrate up to the DnDBase or SQWDndBase classes once more constructors are created

As with the SQW constructor, support code is largely copy/paste from the existing classes with only minimal change to keep the scope of this task under control.


horace_core/sqw/@DnDBase/DnDBase.m
```diff
@@ -1,24 +1,217 @@
...
+        ulen
+        ulabel
+        iax
+        iint
+        pax
+        p
+        dax
+        s
+        e
+        npix
```
hsautessella, line 29: Adding comments above these properties will fill out the "Property Summary" section on the Matlab help. E.g. [`PixelData` ](https://github.com/pace-neutrons/Horace/blob/master/horace_core/sqw/PixelData/%40PixelData/PixelData.m#L80-L102):
image

horace_core/sqw/@DnDBase/copy.m
```diff
@@ -0,0 +1,7 @@
+function wout = copy(w)
+% Copy this DnD object or an array of DnD objects
+%
+%   >> wout = copy(w);
+
+    wout = w;
```
hsautessella, line 6: Indent here is inconsistent with most of the other Horace functions, which are not indented.

horace_core/sqw/@DnDBase/equal_to_tol_internal.m
```diff
@@ -0,0 +1,31 @@
+function [ok, mess] = equal_to_tol_internal(w1, w2, name_a, name_b, varargin)
```
hsautessella, line 1: If this is internal, should it be in a `private` folder, or does that mess up the "protected"-ness?

hsautessella: It's also common (in Horace) to suffix private functions with an underscore, although I realise this isn't done consistently

horace_core/sqw/@DnDBase/equal_to_tol_internal.m
```diff
@@ -0,0 +1,31 @@
...
+opt = struct('reorder', true, 'fraction', 1);
+flagnames = {};
+cntl.keys_at_end = false;
+cntl.keys_once = false; % so name_a and name_b can be overridden
+[args, opt, ~, ~, ok, mess] = parse_arguments(varargin, opt, flagnames, cntl);
+if ~ok
+    error(mess);
+end
+if ~islognumscalar(opt.reorder)
+    error('''reorder'' must be a logical scalar (or 0 or 1)')
```
hsautessella, line 14: Is it worth putting an error ID in here?

horace_core/sqw/@sqw/equal_to_tol_internal.m
```diff
@@ -0,0 +1,96 @@
+function [ok, mess] = equal_to_tol_internal(w1, w2, name_a, name_b, varargin)
```
hsautessella, line 1: The first 30-40 lines of this `equal_to_tol_internal` look very similar to the version in `DnDBase`. Is there a way to remove the duplication?

_test/test_sqw_class/test_dnd_constructor.m
```diff
@@ -0,0 +1,108 @@
...
+        assertTrue(isa(dnd_obj, 'd2d'));
+        assertTrue(isa(dnd_obj, 'DnDBase'));;
+        assertTrue(isa(dnd_obj, 'SQWDnDBase'));
+    end
+
+    function test_default_constructor_returns_empty_instance(obj)
+        dnd_obj = d2d();
+
+        assertTrue(isa(dnd_obj, 'd2d'));
+        assertEqual(numel(dnd_obj.pax), 2);
```
hsautessella, line 31: The name of this test suggests that we should have an `isempty` call in here somewhere. I see that `isempty` isn't implemented on `DnDBase`, should we be checking `isempty(dnd_obj.s) && isempty(dnd_obj.e)` is true?

horace_core/sqw/@DnDBase/equal_to_tol_internal.m
nickbattam-tessella: The function must be `protected` in order that `equal_to_tol` in the base class can access it.

horace_core/sqw/@sqw/equal_to_tol_internal.m
nickbattam-tessella: Probably, I've left if here intentionally for this first dnd implementation. Once the final form of the common dnd code is clear it is easier to assess whether moving everything up to the SQWDnDBase is helpful.

Most of the repeated code is argparsing.

_test/test_sqw_class/test_dnd_constructor.m
nickbattam-tessella: This mirrors the equivalent SQW test: 
```    
function test_default_constructor_returns_empty_instance(obj)
        sqw_obj = sqw();

        assertTrue(isa(sqw_obj, 'sqw'));
        assertEqual(sqw_obj.main_header, struct([]));
        assertEqual(sqw_obj.header, struct([]));
        assertEqual(sqw_obj.detpar, struct([]));
        assertEqual(sqw_obj.data.pix, PixelData());
        assertEqual(numel(sqw_obj.data.pax), 0);
 end
```

There is no isempty method on either class, I'm not sure there any end user use-case for such a function either.

_test/test_sqw_class/test_dnd_copy.m
```diff
@@ -0,0 +1,41 @@
...
+            d2d_copy = copy(d2d_obj);
+
+            assertEqualToTol(d2d_copy, d2d_obj);
+        end
+
+        function test_copy_returns_distinct_object(obj)
+            d2d_obj = d2d(obj.test_dnd_2d_fullpath);
+            d2d_copy = copy(d2d_obj);
+
+            d2d_copy.s(1:10) = 1e1000;
```
hsautessella, line 30: Not really important, but `1e1000` evaluates to `inf`. Would it be more explicit to just use `inf` here?

hsautessella (approved):
Looks good to me 👍 

Approved with 1 or 2 very minor things.

Might it be cleaner to have the `d2d` tests in a `test_d2d` or (the existing) `test_dnd` folder? It feels strange that they are in `test_sqw_class/`.

_test/test_sqw_class/test_dnd_constructor.m
```diff
@@ -0,0 +1,143 @@
...
+        dnd_obj = d2d();
+        actual_props = fieldnames(dnd_obj);
+
+        assertEqual(numel(actual_props), numel(expected_props));
+        for idx = 1:numel(actual_props)
+            assertTrue( ...
+                any(strcmp(expected_props, actual_props(idx))), ...
+                sprintf('Unrecognised DnD property "%s"', actual_props{idx}));
+        end
+    end
```
cmarooney-stfc, line 66: Not sure this does what it says. It is checking that the actual properties you create are in the list of expected properties. It does not check that all the expected properties are in the object.

_test/test_sqw_class/test_dnd_constructor.m
```diff
@@ -0,0 +1,143 @@
...
+
+  function test_copy_constructor_clones_object(obj)
+        dnd_obj = d2d(obj.test_dnd_2d_fullpath);
+        dnd_copy = d2d(dnd_obj);
+
+        assertTrue(isa(dnd_obj, 'd2d'));
+        assertEqualToTol(dnd_copy, dnd_obj);
+    end
```
cmarooney-stfc, line 89: Being paranoid, should one check that dnd_copy is not the same object as dnd_obj?

_test/test_sqw_class/test_dnd_constructor.m
```diff
@@ -0,0 +1,143 @@
...
+        assertFalse(equal_to_tol(dnd_copy, dnd_obj));
+    end
+
+
+    function test_filename_constructor_returns_populated_class(obj)
+        d2d_obj = d2d(obj.test_dnd_2d_fullpath);
+
+        expected_ulen = [2.101896, 1.486265, 2.101896, 1.0000];
+        expected_u_to_rlu = [1, 0, 1, 0; 1, 0, -1, 0; 0, 1, 0, 0; 0, 0, 0, 1];
+
```
cmarooney-stfc, line 115: Being picky, should the four rows of the matrix be on separate lines. Would be a lot easier to parse the contents

_test/test_sqw_class/test_dnd_constructor.m
```diff
@@ -0,0 +1,143 @@
...
+        assertEqualToTol(d2d_obj.ulen, expected_ulen, 'tol', 1e-5);
+        assertEqual(d2d_obj.u_to_rlu, expected_u_to_rlu, 'tol', 1e-5);
+    end
+
+    function test_sqw_constuctor_raises_error_if_not_2d(obj)
+        sqw_obj = sqw(obj.test_sqw_1d_fullpath);
+        f = @() d2d(sqw_obj);
+
+        assertExceptionThrown(f, 'D2D:d2d');
+    end
```
cmarooney-stfc, line 130: and does it not raise an error if you use the 2d file?

_test/test_sqw_class/test_dnd_copy.m
```diff
@@ -0,0 +1,41 @@
+classdef test_dnd_copy < TestCase
+
+    properties
+        sqw_file_1d_name = 'dnd_2d.sqw';
+        sqw_files_path = '../test_sqw_file/';
+
```
cmarooney-stfc, line 6: why are these paths being named sqw_? It is giving the impression they are paths to sqw objects. Is dnd_ preferable?

_test/test_sqw_class/test_dnd_copy.m
```diff
@@ -0,0 +1,41 @@
...
+            d2d_copy.s(1:10) = 1e1000;
+            d2d_copy.e = [];
+            d2d_copy.ulabel = {'test'};
+
+            % changed data is not mirrored in initial
+            assertFalse(equal_to_tol(d2d_copy.s, d2d_obj.s));
+            assertFalse(equal_to_tol(d2d_copy.e, d2d_obj.e));
+            assertFalse(equal_to_tol(d2d_copy.ulabel, d2d_obj.ulabel));
+        end
```
cmarooney-stfc, line 38: or more simply assertFalse(d2d_copy == d2d_obj)? Assumes d2d is handle class.

_test/test_sqw_class/test_equal_to_tol.m
```diff
@@ -41,7 +45,61 @@ function delete(obj)
...
+            elseif isstring(dnd_copy.(field_name))
+                dnd_copy.(field_name) = 'test_value';
+            else
+                dnd_copy.(field_name) = [];
+            end
+            
+            [ok, mess] = equal_to_tol(obj.dnd_2d, dnd_copy);
+            assertFalse(ok, ['Expected ', field_name, ' to be not equal']);
+        end
+    end
```
cmarooney-stfc, line 97: Probably not a performance issue as it is only a small test, but you are retesting the entire object for inequality every time you change one field. Hence after the first changed field, the others may actually be equal but you will still get ok==false

cmarooney-stfc (changes requested):
Partial review, down to but not including DnDBase.m

_test/test_sqw_class/test_dnd_constructor.m
nickbattam-tessella: There is a check before the for loop that the number of properties match; in combination these two check confirm the object properties are the expected ones:
```assertEqual(numel(actual_props), numel(expected_props));```

File: _test/test_sqw_class/test_dnd_copy.m (hunk 13 above)
Comment (nickbattam-tessella) [reply _test/test_sqw_class/test_dnd_copy.m:38]: d2d is not a handle class at the moment.

`==` isn't a defined operator. 

Simply testing that `d2d_copy != d2d_obj` wouldn't confirm that all three attributes we have changed - that would return false if any of them had changed. This was a failure mode in the SQW class for a while.

File: _test/test_sqw_class/test_dnd_copy.m (hunk 12 above)
Comment (nickbattam-tessella) [reply _test/test_sqw_class/test_dnd_copy.m:6]: cut+paste error, fixed now.

File: _test/test_sqw_class/test_dnd_constructor.m (hunk 11 above)
Comment (nickbattam-tessella) [reply _test/test_sqw_class/test_dnd_constructor.m:130]: that is an implicit outcome of `test_sqw_constuctor_creates_obj_with_sqw_data` which I will rename as to emphasise that it is a 2d sqw file.

File: _test/test_sqw_class/test_equal_to_tol.m (hunk 14 above)
Comment (nickbattam-tessella) [reply _test/test_sqw_class/test_equal_to_tol.m:97]: That's a good catch, the dnd copy should be inside the loop

Comment (nickbattam-tessella) [reply _test/test_sqw_class/test_equal_to_tol.m:97]: will fix SQW case too

File: horace_core/sqw/@d2d/d2d.m [hunk 15]
```diff
@@ -0,0 +1,105 @@
...
+    methods(Access = protected)
+        wout = unary_op_manager(obj, operation_handle);
+        wout = binary_op_manager_single(w1,w2,binary_op);
+    end
+
+    methods(Static, Access = private)
+
+        function args = parse_args(varargin)
+            % Parse a single argument passed to the SQW constructor
+            %
```
cmarooney-stfc, line 43: It's a DnD/D2D constructor, not SQW

horace_core/sqw/@d2d/d2d.m
```diff
@@ -0,0 +1,105 @@
...
+                        ['D2D cannot be constructed from an instance of this object "' class(input) '"']);
+                end
+            elseif is_string(parser.Results.input)
+                args.filename = input;
+            elseif isstruct(input) && ~isempty(input)
+                args.data_struct = input;
+            else
+                % create struct holding default instance
+                args.data_struct = data_sqw_dnd(d2d.NUM_DIMS);
+            end
```
cmarooney-stfc, line 72: Checking if this is the data_sqw_dnd in the old implementation or revised by you

cmarooney-stfc (changes requested):
Partial review down to but excluding @sqw/equal_to_tol_internal

horace_core/sqw/@d2d/d2d.m
nickbattam-tessella: this is the old implmentation

horace_core/sqw/@DnDBase/DnDBase.m
```diff
@@ -1,24 +1,221 @@
...
+        function val = get.dax(obj)
+            val = [];
+            if ~isempty(obj.data)
+                val = obj.data.dax;
+            end
+        end
+        function obj = set.dax(obj, dax)
+            obj.data.dax = dax;
+        end
+
```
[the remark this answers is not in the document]
nickbattam-tessella: Yes. At the moment this data is simply a struct, so everything is accessible

When we move out of storing things in the data_sqw_dnd object we can look at adding access checks

horace_core/sqw/@DnDBase/copy.m
```diff
@@ -0,0 +1,7 @@
+function wout = copy(w)
+% Copy this DnD object or an array of DnD objects
+%
+%   >> wout = copy(w);
+
+wout = w;
+end
```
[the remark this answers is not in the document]
nickbattam-tessella: SQWDnDBase is intentionally not a handle class (yet). If it needs to become one later it will be straightforward to update the code and copy function.
There is explicit handling in the SQW `copy()` method to manage the handle-class PixelData attribute. Additions similar to that will need to be added here if the support classes are handle-classes.

horace_core/sqw/@DnDBase/DnDBase.m
```diff
@@ -1,24 +1,221 @@
...
-            obj = obj@SQWDnDBase(varargin{1});
-            
-        end
-        
-        function outputArg = method1(obj,inputArg)
-            %METHOD1 Summary of this method goes here
-            %   Detailed explanation goes here
-            outputArg = obj.Property1 + inputArg;
+            obj = obj@SQWDnDBase();
+        end
```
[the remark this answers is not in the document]
nickbattam-tessella: Nothing to do here at the moment (yet). There are no props on this or the base class. That will change!

horace_core/sqw/@DnDBase/DnDBase.m
```diff
@@ -1,24 +1,221 @@
...
+    methods (Access = protected)
+        [ok, mess] = equal_to_tol_internal(w1, w2, name_a, name_b, varargin);
     end
-    
+
     methods
+        % function signatures
+        [nd, sz] = dimensions(w)
+        wout = copy(w)
+
```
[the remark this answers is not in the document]
nickbattam-tessella: Matlab semicolons appear to be largely optional; I will add them for consistency

horace_core/sqw/@DnDBase/equal_to_tol_internal.m
```diff
@@ -0,0 +1,33 @@
+function [ok, mess] = equal_to_tol_internal(w1, w2, name_a, name_b, varargin)
+% Compare scalar DnD objects of same type
+
+% Check for presence of reorder and/or fraction option(s) (only relevant if sqw-type)
+opt = struct('reorder', true, 'fraction', 1);
```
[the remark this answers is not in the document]
nickbattam-tessella: This is consistent with the codestyle across our codebase: "within blocks other than the outermost one" is indented.

This means that methods are indented inside classes as these are defined inside a `methods ( ) .... end` blocks, but the blocks themselves, contained in the outer `classdef ... end` block, aren't.
For functions defined outside of classes like this the code isn't indented the the only block is outermost `function...[end]`.

horace_core/sqw/@DnDBase/equal_to_tol_internal.m
```diff
@@ -0,0 +1,33 @@
...
+flagnames = {};
+cntl.keys_at_end = false;
+cntl.keys_once = false; % so name_a and name_b can be overridden
+[args, opt, ~, ~, ok, mess] = parse_arguments(varargin, opt, flagnames, cntl);
+if ~ok
```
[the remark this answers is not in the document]
nickbattam-tessella: This is code migrated from the existing implementation and uses the Herbert `parse_arguments` function. It's not obvious from that how to pass optional arguments with validation.
The `args` are passed through to further `equal_to_tol` methods defined elsewhere, so any change to parameter handling in this function would need to be made globally.

_test/test_sqw_class/test_dnd_constructor.m
nickbattam-tessella: I believe that is covered by `test_copy_returns_distinct_object`

horace_core/sqw/@DnDBase/DnDBase.m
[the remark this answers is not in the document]
abuts: Matlab semicolons suppress print-out to stdout when method is executed. They are indeed optional when you define signatures, as signatures are not executed.

horace_core/sqw/@DnDBase/DnDBase.m
```diff
@@ -1,24 +1,221 @@
...
+        iint % Integration range along each of the integration axes. [iint(2,length(iax))]
+             %     e.g. in 2D case above, is the matrix vector [u1_lo, u3_lo; u1_hi, u3_hi]
+        dax %Index into data.pax of the axes for display purposes. [row vector]
+        p % Cell array containing bin boundaries along the plot axes [column vectors]
+          %                       i.e. row cell array{data.p{1}, data.p{2} ...}
+        pax % Index of plot axes into the projection axes  [row vector]
+        s % Cumulative signal
+        e % Cumulative variance
+        npix % Number of contributing pixels to each bin of the plot axes
+    end
```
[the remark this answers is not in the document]
nickbattam-tessella: Added note that this is temporary

horace_core/sqw/@d2d/d2d.m
```diff
@@ -2,9 +2,10 @@
     %D2D Create an 2-dimensional DnD object
     %
     % Syntax:
-    %   >> w = d2d()               % Create a default, zero-dimensional SQW object
+    %   >> w = d2d()               % Create a default, empty, D2D object
+    %   >> w = d2d(sqw)            % Create an D2D object from a file
```
cmarooney-stfc, line 6: Think the comment for this one is wrong

cmarooney-stfc (changes requested):
Just one query on a comment

PR #467, opened 2020-11-27
Do not use Groovy interpolation for Jenkins creds

hsautessella:
If you use Groovy string interpolation when passing credentials to a
shell command, the secret will be executed in plain text on the agent.
The Jenkins logs will still mask the credentials, but any command
logging on the agent machine will leak the secret.

Using the environment variable set by the withCredentials block means
the command does not contain the secret, any command logs will only show
an environment variable was accessed.

See https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#interpolation-of-sensitive-environment-variables.


nickbattam-tessella (commented):
Changes look fine. It would be helpful to add some information into the SMG clearly explaining why this syntax is used, passing the env variable name not value into the script, to ensure any future script updates don't fall back to the simple `${variable}` implementation

nickbattam-tessella (approved):

Issue #429, opened 2020-10-30
Add Unary operations to DND classes


nickbattam-tessella:
The unary operations are defined for both DND and SQW classes.

The APIs and implementation of this method are identical for DND and SQW objects operating on the image s, e via a sigvar object through the unary_op_manager class which has distinct implementations.

Outcome:
- [ ] operations implemented on DnD objects

PR #475, opened 2020-12-04
Add Unary ops to d2d class

nickbattam-tessella:
Add unary ops to D2D object.

This is a trivial change as the public API is shared with the SQW implementation, the unary operations manager, abstract on the `SQWDnDBase` class is the only new code and has been copied from the old class implementation. 

The tests are a reproduction of the SQW ones.

Addresses #429


_test/test_dnd_class/test_unary_ops.m
```diff
@@ -0,0 +1,93 @@
...
+            dnd_obj = d2d();
+            dnd_obj.npix = [num_pix];
+            dnd_obj.s = data.s;
+            dnd_obj.e = data.e;
+
+            % exception will be thrown if method not implemented
+            result = unary_op(dnd_obj);
+
+            % confirm the data (s, e) has been updated
+            assertFalse(equal_to_tol(result.s, dnd_obj.s));
```
oerc0122, line 62: Would it be a better test of whether the functions are defined to operate with the identity and check that things are unchanged? 

If I understand correctly, currently you're testing random data which may happen to be the identity, which would result in false positives. 

Personally, I don't like random data in tests in general.

nickbattam-tessella: I agree that the use of random data in tests is something to consider carefully.


The prime purpose of this test is that the function exists in the class API and does not generate an error when called with an appropriately shaped data array.

In calling it with an `nxm` array of random numbers that lie within "valid input range" for the operations (so we're not, for example, asking for logs of negative numbers) the likelihood of all of these values being the identity value of the function is vanishingly small.

The second test verifies the actual implementation of "an example" unary function to confirm the resulting numerical data is correct.

nickbattam-tessella: This is a cut-down implementation of the sqw unary operator tests where there is an additional layer of complexity in the function is applied to the raw pixel data (data.pix.s) and the image data is recalculated from that (data.s). There are similar tests in the PixelData class too.
Some simplification is needed in these tests and will be made in a new commit.

Comprehensive testing of the actual mathematical implementations of the various unary ops should be done in Herbert where the unary operators are implemented. Those tests are currently missing.

_test/test_dnd_class/test_unary_ops.m
```diff
@@ -0,0 +1,81 @@
...
+    function test_unary_op_updates_image_signal_and_error(~)
+        dnd_obj = d2d();
+        dnd_obj.s = [2, 10245]; % simple dataset for ease of testing
+        dnd_obj.e = [1.5, 1021];
+
+        % arbitrary unary op for test
+        result = log10(dnd_obj);
+
+        % explicit calculation test
+        % reference value calculation match implementation: log10_single
```
hsautessella, line 72: May be just me, but this sentence took a lot of parsing
```suggestion
        % calculate reference values using code similar to 'log10_single'
```
?

horace_core/sqw/@DnDBase/sigvar_set.m
```diff
@@ -0,0 +1,19 @@
+function w = sigvar_set(win, sigvarobj)
+% Set output object signal and variance fields from input sigvar object
+%
+%   >> w = sigvar_set(win, sigvarobj)
+
+if ~isequal(size(win.s),size(sigvarobj.s))
+    error([upper(class(win)) ':sigvar_set'], ...
+        [class(win) ' object and sigvar object have inconsistent sizes']);
```
hsautessella, line 8: This error message would be more helpful if the sizes were printed.
```suggestion
        [class(win), ' object and sigvar object have inconsistent sizes: [%s] and [%s]'], ...
        num2str(size(win.s)), num2str(size(sigvarobj.s));
```

horace_core/sqw/@DnDBase/sigvar_set.m
```diff
@@ -0,0 +1,19 @@
+function w = sigvar_set(win, sigvarobj)
```
hsautessella, line 1: Should we have a couple of tests for this function?

hsautessella (commented):
Looks good to me, just a couple of small things

hsautessella (approved):

PR #479, opened 2020-12-07
Create release process flowcharts

hsautessella:
This PR adds a flowchart showing the deployment process to the "deploy to GitHub" SMG.

I created the flowchart in [Draw.io](https://app.diagrams.net/) and saved the result to a `.svg` file. These files can be included as images in Markdown, but can also be loaded back into Draw.io so that the chart can be edited. Hopefully this way of creating flowcharts is suitable.


oerc0122:
I'm thinking we want to document the user process as much as the pipeline process possibly with more detail. To me, the flowchart should be a simplified mirror of the text list above. 

You don't mention details such as ticket creation/tagging or the required user input data such as the github SHAs, version numbers, etc.

Also for more detail:
- What are the "release artefacts" currently?
  - How are they used?
  - Where do they go?
- Where are the (draft) releases published to?

abuts:
As I have already mentioned on the meeting, I am generally satisfied with this flowchart except it does not include documentation (which, apparently, may be incomplete)  otherwise, looks like reasonable guide for actions, except, despite it is recognized as a best practice,  we do not normally test the release ourselves (except automated testing) and give this honor to users. Testing themselves is time-consuming and boring. 

Why have you build it on your private clone? Not that convenient to look at.

abuts (approved):

nickbattam-tessella (approved):

PR #480, opened 2020-12-09
Add Binary Ops to D2D object

nickbattam-tessella:
Move the binary operation manager into the DnD base class and update logic to handle DnD classes in SQW and PixelData settings.
Split binary op tests between SQW and DND test files to match use cases.

[any changes to the InferiorClasses pattern will be made under a separate PR]

### DIFF
--- a/horace_core/sqw/PixelData/@PixelData/do_binary_op.m
+++ b/horace_core/sqw/PixelData/@PixelData/do_binary_op.m
@@ -15,7 +15,7 @@ function pix_out = do_binary_op(obj, operand, binary_op, varargin)
 %              - scalar double
 %              - double array, the size of the array must be equal to
 %                obj.num_pixels
-%              - object with fields 's' and 'e' (e.g. dnd or sigvar)
+%              - object with fields 's' and 'e' (e.g. sigvar)
 %              - another PixelData object with obj.num_pixels equal
 % binary_op  Function handle pointing to the desired binary operation. The
 %            function should take 2 objects with '.s' and '.e' attributes, e.g.
@@ -43,7 +43,7 @@ elseif isa(operand, 'double')
     pix_out = binary_op_double_(pix_out, operand, binary_op, flip, npix);
 elseif isa(operand, 'PixelData')
     pix_out = binary_op_pixels_(pix_out, operand, binary_op, flip);
-elseif ~isempty(regexp(class(operand), '^d[0-4]d_old$', 'ONCE')) || isa(operand, 'DnDBase') || isa(operand, 'sigvar')
+elseif isa(operand, 'sigvar')
     pix_out = binary_op_sigvar_(pix_out, operand, binary_op, flip, npix);
 end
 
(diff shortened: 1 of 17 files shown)


nickbattam-tessella:
I've had to use `classdef (InferiorClasses = {?d2d}) sqw < SQWDnDBase` in the SQW class definition to retain the dominance of the SQW object in calls to preserve the `dnd .op. sqw -> sqw` pattern.

This is a smell as the SQW class needs to include a list of the concrete DnD classes rather than this coming through the class hierarchy. Classes do not inherit their `Inferior` status so we cannot have `DnDBase` and the named class.

I'm interested in any alternate approaches people can see to this.

abuts:
I do not understand why would we want to keep superior/inferior relationships. These should be independent classes, operated via class methods and the inferiority is something necessary for old style classes

nickbattam-tessella:
`InferiorClass` attribute is part of the new-style class specification: https://www.mathworks.com/help/matlab/matlab_oop/class-precedence.html
The use-case we have here matches their second use-case: "If there is a dominant argument, call the method of the dominant class."
We have defined that that SQW is the dominant argument in any binary operations.

horace_core/sqw/PixelData/@PixelData/do_binary_op.m
```diff
@@ -43,7 +43,7 @@
     pix_out = binary_op_double_(pix_out, operand, binary_op, flip, npix);
 elseif isa(operand, 'PixelData')
     pix_out = binary_op_pixels_(pix_out, operand, binary_op, flip);
-elseif ~isempty(regexp(class(operand), '^d[0-4]d_old$', 'ONCE')) || isa(operand, 'sigvar')
+elseif ~isempty(regexp(class(operand), '^d[0-4]d_old$', 'ONCE')) || isa(operand, 'DnDBase') || isa(operand, 'sigvar')
     pix_out = binary_op_sigvar_(pix_out, operand, binary_op, flip, npix);
```
abuts, line 47:
```suggestion
elseif ~isempty(regexp(class(operand), '^d[0-4]d_old$', 'ONCE')) || isa(operand, 'DnDBase') || isa(operand, 'sigvar')
    pix_out = binary_op_sigvar_(pix_out, operand, binary_op, flip, npix);
 ```
Should it be `^d[0-4]d$' now?
+
Can not judge if we want binary_op on `PixelData` itself rather then doing it as part of operation on `sqw` object. But may be missing something there

**Review by abuts — APPROVED** [review]
With reference to the discussion on superiority\inferiority suitability, looks good

File: horace_core/sqw/PixelData/@PixelData/do_binary_op.m (hunk 1 above)
Comment (nickbattam-tessella) [reply horace_core/sqw/PixelData/@PixelData/do_binary_op.m:47]: The `d0d`-`d4d` (new class) case is covered by `DnDBase` check. 

In the `sqw` binary operations implementation, call made as `sqw .op. non_sqw` are implemented as 
```
sqw.pixeldata .op. non_sqw; 
recompute_image;
```
so the logic for the operations is implemented at PixelData level.

You raise a good point about why PixelData needs to know about DnD objects at though - I'll look into this.

hsautessella (approved):
Looks OK to me 👍

PR #482, opened 2020-12-11
DnD constructors

nickbattam-tessella:
Create constructors for remaining dimensions (0d, 1d, 3d, 4d)
This has moved virtually all the code out of the concrete classes.

### DIFF
--- a/_test/test_dnd_class/test_dnd_constructor.m
+++ b/_test/test_dnd_class/test_dnd_constructor.m
@@ -106,20 +106,21 @@ methods
         end
     end
 
-  function test_copy_constructor_clones_d2d_object(obj)
+    function test_copy_constructor_clones_d2d_object(obj)
         dnd_obj = d2d(obj.test_dnd_2d_fullpath);
         dnd_copy = d2d(dnd_obj);
 
         assertTrue(isa(dnd_obj, 'd2d'));
         assertEqualToTol(dnd_copy, dnd_obj);
-  end
-    function test_copy_constructor_clones_d4d_object(obj)
-          dnd_obj = d4d();
-          dnd_copy = d4d(dnd_obj);
+    end
 
-          assertTrue(isa(dnd_obj, 'd4d'));
-          assertEqualToTol(dnd_copy, dnd_obj);
-      end
+    function test_copy_constructor_clones_d4d_object(obj)
+        dnd_obj = d4d();
+        dnd_copy = d4d(dnd_obj);
+
+        assertTrue(isa(dnd_obj, 'd4d'));
+        assertEqualToTol(dnd_copy, dnd_obj);
+    end
 
     function test_copy_constructor_returns_distinct_object(obj)
         dnd_obj = d2d(obj.test_dnd_2d_fullpath);
@@ -138,7 +139,6 @@ methods
 
         assertFalse(equal_to_tol(dnd_copy, dnd_obj));
     end
-
 
     function test_filename_constructor_returns_populated_class(obj)
         d2d_obj = d2d(obj.test_dnd_2d_fullpath);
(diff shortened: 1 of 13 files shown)


horace_core/sqw/@DnDBase/parse_args_.m
```diff
@@ -0,0 +1,35 @@
...
+
+if isa(input, 'SQWDnDBase')
+    if isa(input, class(obj))
+        args.dnd_obj = input;
+    elseif isa(input, 'sqw')
+        args.sqw_obj = input;
+    else
+        error([upper(class(obj)), ':' class(obj)], ...
+            [upper(class(obj)) ' cannot be constructed from an instance of this object "' class(input) '"']);
+    end
```
cmarooney-stfc, line 26: did you mean "instance of this class"?

horace_core/sqw/@DnDBase/parse_args_.m
```diff
@@ -0,0 +1,35 @@
...
+elseif is_string(parser.Results.input)
+    args.filename = input;
```
cmarooney-stfc, line 28: as you have unpacked parser.Results.input as input, may be better to keep the shorter name here

cmarooney-stfc (changes requested):

_test/test_dnd_class/test_dnd_constructor.m
```diff
@@ -80,13 +106,20 @@ function test_d2d_get_returns_set_properties(obj)
...
         assertEqualToTol(dnd_copy, dnd_obj);
-    end
+  end
+    function test_copy_constructor_clones_d4d_object(obj)
+          dnd_obj = d4d();
+          dnd_copy = d4d(dnd_obj);
+
+          assertTrue(isa(dnd_obj, 'd4d'));
+          assertEqualToTol(dnd_copy, dnd_obj);
+      end
```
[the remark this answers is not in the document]
nickbattam-tessella: The idea is to check that changes made to one object are not visible in the second.

The `test_copy_constructor_clones_...` tests confirm that after a copy the new new objects are "equal" (as far as the `equal_to_tol` definition goes.
This test builds on that and verifies that after that copy I can change some attributes and get an object that is no longer equal.

_test/test_dnd_class/test_dnd_constructor.m
```diff
@@ -26,20 +26,46 @@
...
+        assertEqual(numel(dnd_obj.pax), 3);
+        assertEqual(dnd_obj.dimensions(), 3);
     end
+    function test_d4d_constructor_returns_4d_instance(obj)
+        dnd_obj = d4d();
+
+        assertEqual(numel(dnd_obj.pax), 4);
+        assertEqual(dnd_obj.dimensions(), 4);
+    end
+
```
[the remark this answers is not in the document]
nickbattam-tessella: It is also one pixel long.

horace_core/sqw/@DnDBase/parse_args_.m
hsautessella: Also, `input` is a Matlab built-in, so using a different name would probably be best.

_test/test_dnd_class/test_dnd_constructor.m
```diff
@@ -65,8 +121,27 @@ function test_d2d_contains_expected_properties(obj)
         end
     end
 
+    function test_d0d_get_returns_set_properties(obj)
+        dnd_obj = d0d();
+        obj.assert_dnd_get_returns_set_properties(dnd_obj);
+    end
+    function test_d1d_get_returns_set_properties(obj)
```
hsautessella, line 128: Is the intent of not having whitespace between these tests to group them? Would a comment, or even a section delimiter (`%%`), be a better way of grouping them? If we ever decide on a formatter this grouping will be lost

hsautessella (changes requested):
Just a couple of small things

PR #496, opened 2021-01-06
Create 16-serialiser.md

oerc0122:
Add smg surrounding serialiser


documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Standard format
+The new style has a standardised header format depending on the shape of the data (with the exception of function handles).
+This format allows extension to support arrays of objects and structs in a way which the old serialiser couldn't.
+It also greatly simplifies the tag structure and allows it to align with the C++ MEX API data types.
+Data in these cases is serialised in the linearised order as defined by `data(:)` and the array reshaped by the deserialiser.
+
+This standard format for the header is as follows:
+RANK |     tag    | dim1 | dim2 | ... | Data
+-----|------------|------|------|-----|------
```
cmarooney-stfc, line 17: I'm not finding this clear. What is RANK? Is each of these lines a possible header or do they all occur in one header. What determines N or the max number of headers or header rows? I'll look at the tag structure below (it seemed clear on a first glance). I'm presuming that the ellipsis ... stands for a possible number of dimensions, but you don't say. If so, an additional dimM column before data would make that a lot more clear.  What determines M? (Or, looking at the N row below, is N==M?) I was expecting Data to be separate from the header, following it as a body - does this mean that it is not really just a header but a whole serialized record?

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+NULL | 32 + tag   | 0    |      |     |
+0    |    tag     |      |      |     | Data
+1    | 32 + tag   | nElem|      |     | Data
+2    | 64 + tag   | dim1 | dim2 |     | Data
+N    | N<<5 + tag | dim1 | dim2 | ... | Data
```
cmarooney-stfc, line 22: In the tag column, presuming (possibly incorrectly) that tag is the tag structure described below and then 32,64,N<<5 are further bits in front, or offsets? Also how is NULL recorded?

Looking again at the tag description below, that contains 3 bits for the rank, and 5 for the tag. So it looks as if rank and tag are not 2 separate items but one combined item. Is that correct? And does 32 mean bit=1 at the 32 position etc. And is N limited to max 7 as only 3 rank bits are available? (noted that this is confirmed in the limitations section).

Finally, the tag header on the column and the tag value in the column cell are a bit inconsistent. The column tag appears to be rank+type tag

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+
+Where `<<5` means bitshift left 5. (see [Tag Format][tag])
+
+- Tags are `uint8` data and the "magic" numbers are described in detail in the section [Tag Format][tag] below.
+- Dimensions are serialised as `uint32` 
+
+The single exception to this header is that of `function_handles` which are purely scalar and the rank component of the tag is used to specify the type of function handle:
+
+Handle type         |   tag   | value
+--------------------|---------|-------
```
cmarooney-stfc, line 32: So here we are looking at bit patterns (with 25=11001)
00111001
01011001
01111001
checking if this is the case. As the value is the sum of 25 and the tag bit position, presuming it is not a separately stored field - does it make sense to provide the value separately?

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+|:---:|:--:|:--:|:--:|:-:|:-:|:-:|:-:|
+|  R  | R  | R  | T  | T | T | T | T |
+
+- R - Rank
+- T - Type
+
+### Serialisation formats in detail
+
+#### Simple data
+Simple data (tags 0-12) are serialised as:
```
cmarooney-stfc, line 51: ie logical to uint64 in the table below. If the tag table comes above this, then reading this would be easier - also remembering that tag has mutiple meanings here for (a) type and (b) total 8-bit tag value (R+T)

It may be worth spelling out what the ranks NULL 0 and 1 actually mean by a couple of examples.

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+RANK |     tag    | dim1 | dim2 | ... | Data
+-----|------------|------|------|:---:|------
+NULL | 32 + tag   | 0 
+0    |    tag     |      |      |     | Data
+1    | 32 + tag   | nElem|      |     | Data
+2    | 64 + tag   | dim1 | dim2 |     | Data
+N    | N<<5 + tag | dim1 | dim2 | ... | Data
+
+#### Complex Data
+Complex numerical data (tags 13-22) are serialised as:
```
cmarooney-stfc, line 62: ie complex_double to complex_uint64 in the table below. Do we really support complex integers?

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+RANK |     tag    | dim1 | dim2 | ... |    Data   |   ...
+-----|------------|------|------|:---:|-----------|----------
+NULL | 32 + tag   | 0    |      |     |           |
+0    |    tag     |      |      |     | Real Data | Imag Data
+1    | 32 + tag   | nElem|      |     | Real Data | Imag Data
+2    | 64 + tag   | dim1 | dim2 |     | Real Data | Imag Data
+N    | N<<5 + tag | dim1 | dim2 | ... | Real Data | Imag Data
+
+#### Sparse data
+Sparse numeric data (tags 29-31) are serialised as:
```
cmarooney-stfc, line 73: ie sparse_logical to sparse_complex_double in the table below.

e.g. 95=64+31 giving a rank 2 and type sparse_complex_double.

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+
+RANK |     tag    | dim1 | dim2 | ... |Data|     ...            |    ...     |   ...   |  ...
+-----|------------|------|------|:---:|----|--------------------|------------|---------|------
+NULL | 32 + tag   | 0    |      |     | 33 | length(class_name) | class_name |         |
+0    |    tag     |      |      |     | 33 | length(class_name) | class_name | ser_tag | Data |
+1    | 32 + tag   | nElem|      |     | 33 | length(class_name) | class_name | ser_tag | Data |
+2    | 64 + tag   | dim1 | dim2 |     | 33 | length(class_name) | class_name | ser_tag | Data |
+N    | N<<5 + tag | dim1 | dim2 | ... | 33 | length(class_name) | class_name | ser_tag | Data |
+
+Where
```
cmarooney-stfc, line 132: the first data col seems a bit redundant - you already have tag 32+23=55 for object array. It could be good to understand why this is useful.

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+
+Function handles are serialised as:
+Handle type         |   tag   | data                                       | ... 
+--------------------|---------|--------------------------------------------|-----------------------------------------
+Simple/Class simple |  57     | function name as serialised string         |
+Anonymous           |  89     | anonymous function as serialised string    | relevant workspace as serialised struct
+Scoped              |  121    | function parentage as serialised cell array| 
+
+### Tags
+Tag| Meaning
```
cmarooney-stfc, line 165: For ease of reading the document, it could be good to put this at the start of the new format section. Noting that if the 5 bits of tag are used as above, then this schema is full - it cannot be extended further. It looks quite comprehensive, but just wondering if this is future-proof.

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+ 26| VALUE_OBJECT
+ 27| HANDLE_OBJECT_REF
+ 28| ENUM
+ 29| SPARSE_LOGICAL
+ 30| SPARSE_DOUBLE
+ 31| SPARSE_COMPLEX_DOUBLE
+
+### Limitations
+- Java objects cannot be serialized.
+- Cannot handle arrays of more than rank-7.
```
cmarooney-stfc, line 202: Consistent with my comments above on the 3-bit rank in the tag. May be worth spelling this out, and again, wondering if this is future proof - although I think (with my limited vision) that 7 dimensions is plenty for an array before it gets split into components.

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+- Cannot handle arrays where the size of any dimension > (2^32)-1 due to serialising dimensions as `uint32`.
+  __NB.__ This includes sparse matrices.
```
cmarooney-stfc, line 204: Which is about 10^10. Is that a limitation on pixel array (experimental detector array) sizes?

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+- Handles to nested/scoped functions can only be deserialized when their parent functions
+  support the BCILAB argument reporting protocol (e.g., by using arg_define).
+- New MATLAB objects need to be reasonably friendly to serialization; either they support
```
cmarooney-stfc, line 207: BY new, guessing you mean further object formats we may create as we develop? What kind of objects would not fall within this scope? Are we expecting that such objects need not be serialised if we do need that kind of class structure? Are nested objects (class object within class object) ok with this?

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+  construction from a struct, or they support saveobj/loadobj(struct), or all their important
+  properties can be set via set(obj,'name',value).
+
+
+## Old
```
cmarooney-stfc, line 212: The headers are getting a bit abbreviated here.

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+  support the BCILAB argument reporting protocol (e.g., by using arg_define).
+- New MATLAB objects need to be reasonably friendly to serialization; either they support
+  construction from a struct, or they support saveobj/loadobj(struct), or all their important
+  properties can be set via set(obj,'name',value)
+- In anonymous functions, accessing unreferenced variables in the workspace of the original
+  declaration via eval(in) works only if manually enabled via the global variable
+  tracking.serialize_anonymous_fully (possibly at a significant performance hit).
+  note: this feature is currently not rock solid and can be broken either by Ctrl+C'ing
+        in the wrong moment or by concurrently serializing from MATLAB timers.
+
```
cmarooney-stfc, line 273: Header for references?

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+ 24| STRUCT
+ 25| FUNCTION_HANDLE
+ 26| VALUE_OBJECT
+ 27| HANDLE_OBJECT_REF
+ 28| ENUM
+ 29| SPARSE_LOGICAL
+ 30| SPARSE_DOUBLE
+ 31| SPARSE_COMPLEX_DOUBLE
+
+### Limitations
```
cmarooney-stfc, line 200: The anonymous function limitation in the old format appears to have disappeared - may be worth commenting on (and any other changes I missed.)

cmarooney-stfc (changes requested):

documentation/smg/16-serialiser.md
oerc0122: Rank is the array rank as in number of dimensions. The table details the whole header and data simply follows the header as a body, but the type and representation of the data is not described here.

documentation/smg/16-serialiser.md
oerc0122: That is correct. I guess we could, but not sure we need to

documentation/smg/16-serialiser.md
oerc0122: I presume that rank of an array is an established and known term, I mean strictly we're not dealing with mathematical rank because we don't check irreducibility, but array rank is well known.

documentation/smg/16-serialiser.md
oerc0122: I'm not sure by usage in Herbert we strictly need to serialise complexes at all, but it is free to do and better safe than sorry.

documentation/smg/16-serialiser.md
oerc0122: Mostly because it is serialised by a call to `serialise_simple_data`, which adds the tag and dims.

documentation/smg/16-serialiser.md
oerc0122: PixelData objects are to be serialised by their own serialiser, so strictly, no, they can be whatever size and format they wish.

oerc0122: Also, to extend it is simply a case of replacing `uint32` with `uint64` or `double` in 4 places.

documentation/smg/16-serialiser.md
oerc0122: I don't know, that is taken from the original documentation and don't know MATLAB well enough, but given the current restructure is to the "new" object format, I presume that. Nested objects are fine and work as expected.

documentation/smg/16-serialiser.md
oerc0122: Alex said in the Slack I shouldn't bother documenting the old stuff, basically this is leftover that I didn't want to delete immediately (as it was a bit of work to extract) in case someone came back and said "yes, you should".

documentation/smg/16-serialiser.md
oerc0122: They are macros, not references.

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
+# Serialisation details
```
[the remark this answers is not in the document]
oerc0122: I'm not really sure what introduction you want for a serialiser, they're fairly standard things and I don't see the need to describe what it does or is used for. I couldn't really think of any sort of introduction section. Any ideas?

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+
+As of the integration of the C++ serialiser, the serialised format has been updated to be more consistent and to have greater capabilities.
```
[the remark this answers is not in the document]
oerc0122: Merging the serialiser into Herbert.

documentation/smg/16-serialiser.md
oerc0122: Oh sorry, reread the question and it makes more sense, I was originally planning on also writing a quick reference for each, but decided against it as it is unlikely anyone who is going to be manually decoding serialised data will not be able to do a simple addition.

documentation/smg/16-serialiser.md
```diff
@@ -0,0 +1,275 @@
...
+
+For reference, the old format is also detailed below
+
+## New format
+
+### Standard format
+The new style has a standardised header format depending on the shape of the data (with the exception of function handles).
+This format allows extension to support arrays of objects and structs in a way which the old serialiser couldn't.
```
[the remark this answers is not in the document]
oerc0122: I see "header" regarding data transfer, as only possibly being a preceding chunk which is not the data which outlines how the data should be interpreted. Am I not outlining the structure below?

documentation/smg/16-serialiser.md
oerc0122: Would you rather I removed the "data" column here?

documentation/smg/16-serialiser.md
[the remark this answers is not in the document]
cmarooney-stfc: Just a brief comment on what is being serialised and why (data needed to be persisted on file, data needed to be passed from matlab to c++, other reason for serialising.) The kind of thing that would be useful for a new developer coming on board to read and understand the context of the smg.

documentation/smg/16-serialiser.md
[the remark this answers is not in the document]
cmarooney-stfc: It would be good to say that in the text. In general, if it is not obvious to the reviewer what you are referring to, then it is worth adding to the text to make it clear. In this case, you could say "In the context of merging the serialiser into Herbert, the serialised format has been updated ...." This also begs the question of where the serialiser was before so that a merge into Herbert is needed, and you could also briefly add that information,

documentation/smg/16-serialiser.md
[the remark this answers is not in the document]
cmarooney-stfc: What I'm saying is that you include "Data" in your description of the header. I'm expecting that Data will follow the header but not be part of it. I'm suggesting that you remove Data from the header description, and, if it is needed, say something like "the header will be followed by the data it describes; these headers include all information on the data length so that no "footer" information is required to tag end of data."

documentation/smg/16-serialiser.md
oerc0122: The other thing to bear in mind is that this is per-dimension, i.e. you can serialise a 10^10x10^10 object.

documentation/smg/16-serialiser.md
cmarooney-stfc: As noted above, I think omitting the Data column would make it clearer that you are describing the header and not the data. I am also unclear why there is a separate rank column when that information is available in the tag column.

documentation/smg/16-serialiser.md
cmarooney-stfc: OK so that is the extension mechanism, thanks.

documentation/smg/16-serialiser.md
cmarooney-stfc: OK thanks

documentation/smg/16-serialiser.md
cmarooney-stfc: For those who are familiar with the old stuff, Alex is probably correct, but I think it is useful to understand what the new format brings that the old format didn't. Apart from that I just think it is neater to keep the formatting consistent - you said "New format" so now you say "Old format".

documentation/smg/16-serialiser.md
cmarooney-stfc: Ok thanks

documentation/smg/16-serialiser.md
cmarooney-stfc: The assumption is that rank is an array rank. It may have been a term in a different context. Saying array rank for the data could remove any ambiguity. Other than that, NULL and 1 appear to have the same tag, and the difference is in the size of the first dimension. If the rank is stored both in the first column (see comment above) and the tag, then the tag rank runs from 0 to 7 in options and I don't know what you do to represent NULL in the first column (what value?) and I don't know how you represent NULL in the 3 rank bits of the tag,

cmarooney-stfc: If you added "type" to "tag" (so "type tags 0-12") then this would complete the reference to the header for the type tag table and to the T bits in the tag definition.

documentation/smg/16-serialiser.md
cmarooney-stfc: Assuming that "serialise_simple_data" is your code rather than matlab code, as the code is not necessarily in front of the reader of this document as they read, a quick explanation of this would stop the reader puzzling over why the magic number 33 is here. Equally, as 33 is part of the data, it could just be omitted as it is not required to understand they header.

documentation/smg/16-serialiser.md
oerc0122: The rank is not a part of the data header, but representative of the data, I should probably make that clearer.

documentation/smg/16_serialiser.md
```diff
@@ -0,0 +1,229 @@
+# The serialiser
+
+In order to efficiently transfer MATLAB information between storage, be it via networked interaction (such as MPI) or to disc efficiently, the information is first "serialised" into a representation. This document details the operation of the serialisers defined in:
+```
+<HERBERT_ROOT>/herbert_core/utilities/misc/hlp_serialise
+<HERBERT_ROOT>/herbert_core/utilities/misc/hlp_deserialise
+<HERBERT_ROOT>/_LowLevelCode/cpp/cpp_communicator/serialiser/c_serialise
+<HERBERT_ROOT>/_LowLevelCode/cpp/cpp_communicator/serialiser/c_deserialise
+```
+
```
cmarooney-stfc, line 10: Thanks for this. Looking at the tickets, I was expecting you to clarify that the serialization is for MPI data transfer, and it is very useful to know that it is intended to be generic.

documentation/smg/16_serialiser.md
```diff
@@ -0,0 +1,229 @@
...
+The serialised format has recently been updated to be more consistent and to have greater capabilities such as serialising arrays of structs/objects.
+
+# New format
+
+## Standard header format
+
+The new style has a standardised header format depending on the shape of the data (with the exception of function handles).
+This format allows extension to support arrays of objects and structs in a way which the old serialiser couldn't.
+It also greatly simplifies the tag structure and allows it to align with the C++ MEX API data types.
+Data in these cases is serialised in the linearised order as defined by `data(:)` and the array reshaped by the deserialiser.
```
cmarooney-stfc, line 20: Sorry, do I know what 'data(:)' is?

documentation/smg/16_serialiser.md
```diff
@@ -0,0 +1,229 @@
...
+_2_    | 64 + type  | dim1 | dim2 |     | Real Data | Imag Data
+_N_    | N<<5 + type| dim1 | dim2 | ... | Real Data | Imag Data
+
+#### Sparse data
+Sparse numeric data (type tags 29-31) are serialised as:
+
+_Rank_ |   Type  |     tag  | dim1 | dim2 | Data | ... |    ...    | ...
+-------|---------|----------|------|------|:----:|:---:|-----------|------
+_2_    | Real    | 64 + type| dim1 | dim2 |   i  |  j  | Data      |
+_2_    | Bool    | 64 + type| dim1 | dim2 |   i  |  j  | Data      |
```
cmarooney-stfc, line 126: You said above that you were excluding the data description from the header format. In this section you have been continuing to include the data - which may be a good thing for clarification purposes - but at this point you extend that data to include the sparseness indices and this reader was then wondering if i and j are part of the header or part of the data - and it may be ambiguous which it should be. My instinct is that your tendency to include the data may be an inclination to document the data structure - would it be easier to include that in this document? Apart from that suggestion, I think you need to be consistent in either including or excluding the data, and at least saying where its structure is documented.

documentation/smg/16_serialiser.md
```diff
@@ -0,0 +1,229 @@
...
+_N_    | N<<5 + type| dim1 | dim2 | ... | Data
+
+
+Where data is formatted as follows:
+
+nFields | length(fieldName1) | length(fieldName2) | ... | fieldName1 | fieldName2 | ... | struct2cell(data)
+--------|--------------------|--------------------|:---:|------------|------------|:---:|------------------
+
+nFields and fieldName lengths are serialised as `uint32`
+
```
cmarooney-stfc, line 153: Again, discussing the data format - which may be good, but should be consistent.

documentation/smg/16_serialiser.md
```diff
@@ -0,0 +1,229 @@
...
+_NULL_ | 32 + type  | 0    |      |     | 33 | length(class_name) | class_name |         |
+_0_    |    type    |      |      |     | 33 | length(class_name) | class_name | ser_tag | Data |
+_1_    | 32 + type  | nElem|      |     | 33 | length(class_name) | class_name | ser_tag | Data |
+_2_    | 64 + type  | dim1 | dim2 |     | 33 | length(class_name) | class_name | ser_tag | Data |
+_N_    | N<<5 + type| dim1 | dim2 | ... | 33 | length(class_name) | class_name | ser_tag | Data |
+
+Where the first 3 elements of data 
+(`33`,`length(class_name)`,`class_name`) are a serialised representation of the class name (see [Simple data][simpledata]),
+`ser_tag` is a `uint8` tag (described below) and
+`data` is the serialisation of the object by means described by `ser_tag`
```
cmarooney-stfc, line 183: Again, discussing the data format.

documentation/smg/16_serialiser.md
```diff
@@ -0,0 +1,229 @@
...
+Simple/Class simple | 32 + 25
+Anonymous           | 64 + 25
+Scoped              | 96 + 25
+
+Function handles are serialised as:
+Handle type         |   tag   | data                                       | ...
+--------------------|---------|--------------------------------------------|-----------------------------------------
+Simple/Class simple |  57     | function name as serialised string         |
+Anonymous           |  89     | anonymous function as serialised string    | relevant workspace as serialised struct
+Scoped              |  121    | function parentage as serialised cell array|
```
cmarooney-stfc, line 208: Again it's the data structure. If you are going to include this, it would be worth including what 'relevant workspace' means. I note that matlab documents function workspaces, but it is not clear that a function would need its function workspace serialising unless you are serialising in the middle of the function call. Is that the case here?

cmarooney-stfc (changes requested):
Hi Jacob, this is now becoming much clearer, thank you. My main comments in this hopefully final review are about whether to go the whole hog and include the data formats. The text keeps veering in that direction and it may be simpler to document them here and change the intro text appropriately. If not, it should be clear where those formats are documented.

documentation/smg/16_serialiser.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Limitations
+- Java objects cannot be serialized.
+- Cannot handle arrays of more than rank-7.
+- Cannot handle arrays where the size of any dimension > (2^32)-1 due to serialising dimensions as `uint32`.
+  __NB.__ This includes sparse matrices.
+- Handles to nested/scoped functions can only be deserialized when their parent functions
+  support the BCILAB argument reporting protocol (e.g., by using arg_define).
```
nickbattam-tessella, line 216: What is the BCILAB argument reporting protocol? Can you add a reference to some documentation

nickbattam-tessella (changes requested):

documentation/smg/16_serialiser.md
oerc0122: Turns out it __is__ the brain-computer interface stuff, seems they've done some general MATLAB improvements in their branch.

https://sccn.ucsd.edu/wiki/Arg_define

documentation/smg/16_serialiser.md
oerc0122: This is the documentation for the serialisation of the data, not the header. The point of the header section is that it is a standardised form as we can see throughout here. If anything the header should be removed here, but I think it provides a more complete view of what you're going to see in the serialised data, rather than splitting them.

documentation/smg/16_serialiser.md
oerc0122: From what I understand if you have an anonymous function which calls upon static variables defined in their workspace, this becomes their relevant workspace. So if you have an object which has a function as part of its operation. E.g.
```a = PixData(...); 
n = 1.5;
b = @(x)(x * n); % Trying to increase contrast
a = apply(b, a); % I don't know what the actual Horace syntax will be
```
If we want to apply `b` in parallel, all processes need to know what `b` is, and `b` depends on `n` (trivially in this case, but not in general), hence `n` would be part of `b`'s relevant workspace.

documentation/smg/16_serialiser.md
oerc0122: `array(:)` is standard MATLAB notation for treating a multidimensional-array as a vector (i.e. in memory order). 
e.g.
```
a = eye(4); %2-D identity
a(:)
```

cmarooney-stfc: @oerc0122 Sorry, that wasn't my point. 'data' appears to be a particular variable, which I don't think has been defined. I just added the (:) to quote the complete string. Is 'data' supposed to be the Data following the header? I think this is the first time it has been identified in this document as a MATLAB variable as opposed to a general entity.

oerc0122: Sorry, I'll try to make that more clear

documentation/smg/16_serialiser.md
cmarooney-stfc: @oerc0122 So it's the workspace for the anonymous function option value 89, not a general datum for all the options. If that were specified in the "relevant workspace" section this would be fine.

cmarooney-stfc: @oerc0122 And sorry, looking at the format and realising you actually intend this, as it's aligned with 89, but it looks to a casual reader as if it applies to all the options. May still be worth specifying to avoid this.

oerc0122: Don't forget in the rendered document this is a table with a blank space there. 

https://github.com/pace-neutrons/Horace/blob/233_c%2B%2B_serialiser_smg/documentation/smg/16_serialiser.md#function-handles

documentation/smg/16_serialiser.md
cmarooney-stfc: @oerc0122 The whole document may be for the serialisation of the data, but we are in the section (I think, following the ##etc structure) for the tag format, which is part of the header format. In fact most of the document appears to be header and tag. You may want to think about the section titles and any introductory text to the sections. I would personally put tag as a sub-head of header as that is where tags are used. Happy to discuss. But a section starting ## Data would be useful to indicate when you move onto that from the header.

nickbattam-tessella (approved):

documentation/smg/16_serialiser.md
```diff
@@ -0,0 +1,244 @@
...
+The serialised format has recently been updated to be more consistent and to have greater capabilities such as serialising arrays of structs/objects.
+
+# New format
+
+## Standard header format
+
+The new style has a standardised header format depending on the shape of the data (with the exception of function handles).
+This format allows extension to support arrays of objects and structs in a way which the old serialiser couldn't.
+It also greatly simplifies the tag structure and allows it to align with the C++ MEX API data types.
+Data is serialised in the linearised order as defined by `array(:)` and the array is subsequently reshaped by the deserialiser to the shape as defined by the header.
```
cmarooney-stfc, line 20: @oerc0122 Hopefully last item. I still don't know what you mean by "as defined by `array(:)`". Do you mean a specific variable 'array', or do you mean the generic unpacking of the MATLAB array that represents data. Or something else. Either way, saying what 'array' refers to. Thanks.

cmarooney-stfc (changes requested):

cmarooney-stfc (approved):
This does the job - thanks

PR #498, opened 2021-01-06
SQW object API migration to new class

nickbattam-tessella:
First stage of migration for API from old SQW class to new.

Note:
- basic tests still incomplete
- fit functions explicitly excluded from this PR


nickbattam-tessella:
Note: re-enabling the loadobj method has broken the tests still using the old sqw class that had been fixed to re-interpret a .mat sqw file as a sqw_old class. Working on a fix.

nickbattam-tessella:
Incomplete tests are disabled to allow merge into the feature branch to unblock work on #540

hsautessella (commented):
It's difficult to tell what has been changed and what hasn't. The `@sqw` methods all look like they're new? Would it be sensible to merge master into this branch so that we can be sure the files you've moved are equivalent/updates of what's currently in master? Or is the plan to do this later?

Issue #481, opened 2020-12-10
Remove 'size' overload on PixelData class


hsautessella:
The overload creates ambiguity about whether a `PixelData` object is a single object or an array of objects. Indeed, an array of pixel data objects has a very strange size: 
```matlab
>> size([PixelData(10), PixelData(20)])

ans =

     9     9    10    20
```

From memory, there are `size` calls on `PixelData` object throughout the codebase. It will be necessary to find these and replace them with `.num_pixels`: `size(pix, 2) → pix.num_pixels`. If there are any calls that get the size of a `PixelData` object's first dimension (`size(pix, 1)`), a new `.num_fields` (or similar) attribute may need to be added.

PR #499, opened 2021-01-07
Remove size overload on PixelData class

hsautessella:
This PR removes the `size` overload from the `PixelData` class.

I did a grep through code to try and find all the calls to `size` on `PixelData` objects - hopefully I've caught them. Most of them were in tests anyway.

Alternatives to using `size`:
```matlab
size(pix, 1) == size(pix.data, 1);  % should be avoided in production code as structure of `.data` is not guaranteed
size(pix, 2) == pix.num_pixels;
```
At the point we need it, we will want to add a method like `pix.num_fields` (which would usually be 9), so we can avoid doing `size(pix.data, 1)`.

Addresses #481

### DIFF
--- a/_test/test_mex_nomex/test_main_mex.m
+++ b/_test/test_mex_nomex/test_main_mex.m
@@ -148,13 +148,13 @@ classdef test_main_mex < TestCase
             hcf.use_mex = 0;
             [u_to_rlu_matl,urange_matl,pix_m]=rd.calc_projections();
 
-            assertEqual(size(pix_m,1),9)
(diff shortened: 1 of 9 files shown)


nickbattam-tessella (approved):
Looks good.

~Is it worth adding the `num_fields` attribute now too? While we don't have custom columns this will just return `PixelData.DEFAULT_NUM_PIX_FIELDS`.~
After discussion with Harry agree that there is no use case for this at the moment.

cmarooney-stfc (approved):

oerc0122 (approved):

Issue #500, opened 2021-01-07
mask_random_pixels does not work


abuts:
User reported bug with Horace3.4 pixels masking
Should be fixed

`Hi Alex,
 
Here is the script I was using on IDAaaS, just takes a cut and then tries to mask 90% of pixels, with the latter giving the error I showed you before.
Namely:
cc_small=mask_random_fraction_pixels(cc,0.1);

 

Error using PixelData/set.data (line 465)

Cannot set pixel data, invalid dimensions. Axis 2 must have num elements matching current pagesize (22933292), found '2293329'.

Error in sqw/mask_pixels (line 85)

wout.data.pix.data=win.data.pix.get_pixels(mask_array).data;

Error in sqw/mask_random_fraction_pixels (line 69)   

wout(i) = mask_pixels(win(i),mask0);

 

Russell
```
sqw_file='/instrument/MAPS/RBNumber/RB2010594/ei50_horace_base_psi_sum.sqw'; 
proj.u=[1,1,0]; proj.v=[0,0,1]; proj.type='rrr';proj.nonorthogonal=true;
proj1.u=[1,0,0];proj1.v=[0,0,1];proj1.type='rrr';
proj2.u=[1,0,0];proj2.v=[-0.5,1,0];proj2.w=[0,0,1];proj2.type='rrr';proj2.nonorthogonal=true;

cc=cut_sqw(sqw_file,proj,[-1,0.05,1],[-5,5],[-0.5,0.5],[0,0.4,30]);
plot(compact(cc))

%Size of pix array is 22933292. Mask 90% of pixels randomly to generate
%smaller object for faster fitting.

%Mask 90% of pixels, to speed up.
cc_small=mask_random_fraction_pixels(cc,0.1);
```

hsautessella:
I'm a bit confused because v3.4 does not have `PixelData` objects. Are we sure he's not using v3.5?

cmarooney-stfc:
@hsautessella Hi Harry. It's unclear why this is in review/QA. I see your query on the versioning but I thought it should stay in "In Progress" until that is resolved. In the old days I would have this in a category "Need More Info" which is different from Review=Fixed pending testing it now works. Have we done an initial test on Russell's data and has Russell been copied into the loop?

cmarooney-stfc:
@hsautessella Hi again. I have now seen PR 501, so sorry, missed the connection. Should we be putting the PR number into the ticket as well as the ticket number into the PR? Thanks

[...]

PR #501, opened 2021-01-08
Mask random pixels does not work

hsautessella:
This PR fixes and adds tests for `sqw.mask_pixels` which was failing as it had not been adapted to use `PixelData.mask` (see 
eae2783).

I also noticed that `PixelData.compute_bin_data` was throwing an unexpected error if the `PixelData` object contained no pixels. I took the decision to just return empty arrays in this case, another option was to throw an error - what do you think?

Addresses #500 

_Once this PR is merged into `rel_3_5` I'll open another PR to merge it into `master`_

### DIFF
--- a/_test/test_sqw/test_PixelData_operations.m
+++ b/_test/test_sqw/test_PixelData_operations.m
@@ -164,6 +164,28 @@ methods
         assertEqual(e, obj.ref_e_data, '', obj.FLOAT_TOLERANCE);
     end
 
+    function test_compute_bin_data_mex_rets_empty_arrays_if_num_pix_is_zero(obj)
+        cleanup_handle = ...
+            set_temporary_config_options(hor_config(), 'use_mex', true);
+
+        p = PixelData();
+        [s, e] = p.compute_bin_data([]);
+
+        assertTrue(isempty(s));
+        assertTrue(isempty(e));
+    end
+
+    function test_compute_bin_data_nomex_empty_arrays_if_npix_is_zero(obj)
+        cleanup_handle = ...
+            set_temporary_config_options(hor_config(), 'use_mex', false);
+
+        p = PixelData();
+        [s, e] = p.compute_bin_data([]);
+
+        assertTrue(isempty(s));
+        assertTrue(isempty(e));
+    end
+
     function test_do_unary_op_returns_correct_output_with_cosine_gt_1_page(obj)
         data = rand(PixelData.DEFAULT_NUM_PIX_FIELDS, 50);
         npix_in_page = 11;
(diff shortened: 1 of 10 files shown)


nickbattam-tessella (approved):

hsautessella:
Test script:
```matlab
sqw_file='.../ei50_horace_base_psi_sum.sqw'; 

proj.u=[1,1,0]; proj.v=[0,0,1]; proj.type='rrr';proj.nonorthogonal=true;
proj1.u=[1,0,0];proj1.v=[0,0,1];proj1.type='rrr';
proj2.u=[1,0,0];proj2.v=[-0.5,1,0];proj2.w=[0,0,1];proj2.type='rrr';proj2.nonorthogonal=true;

cc=cut_sqw(sqw_file,proj,[-1,0.05,1],[-5,5],[-0.5,0.5],[0,0.4,30]);
plot(compact(cc))

%Size of pix array is 22933292. Mask 90% of pixels randomly to generate
%smaller object for faster fitting.

%Mask 90% of pixels, to speed up.
cc_small=mask_random_fraction_pixels(cc,0.1);

disp(['Num pixels pre-mask: ', num2str(cc.data.pix.num_pixels)]);
disp(['Num pixels post-mask: ', num2str(cc_small.data.pix.num_pixels)]);
```

Outputs a plot and the following:
```
Taking cut from data in file .../ei50_horace_base_psi_sum.sqw...
Step   1 of    3; Have read data for 10000820 pixels -- now processing data... ----->  retained  8068814 pixels
Step   2 of    3; Have read data for 10000220 pixels -- now processing data... ----->  retained  8364260 pixels
Step   3 of    3; Have read data for 8445481 pixels -- now processing data... ----->  retained  6500218 pixels
>> mask_bug_500
Taking cut from data in file .../ei50_horace_base_psi_sum.sqw...
Step   1 of    3; Have read data for 10000820 pixels -- now processing data... ----->  retained  8068814 pixels
Step   2 of    3; Have read data for 10000220 pixels -- now processing data... ----->  retained  8364260 pixels
Step   3 of    3; Have read data for 8445481 pixels -- now processing data... ----->  retained  6500218 pixels
Num pixels pre-mask: 22933292
Num pixels post-mask: 2293329
```

nickbattam-tessella:
The implementation of the different `mask` methods all raise an error if the number of pixels retained after the mask is applied is zero, or the number of pixels to keep is greater than the number of pixels.

As a result it's not possible for an empty object to be returned.

From that it would be more consistent to throw an error if the PixelData object contains no pixels too.

But that is then inconsistent with the operation of other PixelData methods (e.g. noisify) when called on an empty object.

abuts (commented):
Looks good but would this ticket include combining mask_random_pixels.m and mask_random_fraction_pixels.m within the same interface?

hsautessella:
The failing test is broken on the `rel_3_5` branch. I'll open another issue for that. I believe this PR is OK to go in since the failing test does not touch any of the edited code.

PR #505, opened 2021-01-11
duplicated projection matrix function has been removed 

abuts:
simple fix should remove duplicated projection matrix and fix the issu


hsautessella (approved):
Tests all passing now 👍

PR #515, opened 2021-01-14
mask random pixels bug master

abuts:
This ticket to merge changes, done on ticket #500 into master

### DIFF
--- a/_test/test_sqw/test_PixelData_operations.m
+++ b/_test/test_sqw/test_PixelData_operations.m
@@ -164,6 +164,28 @@ classdef test_PixelData_operations < TestCase
             assertEqual(e, obj.ref_e_data, '', obj.FLOAT_TOLERANCE);
         end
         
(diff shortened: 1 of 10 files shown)


nickbattam-tessella:
Why is this a new branch rather than merging the `500_mask_random_pixels_does_not_work` branch into master? It's hard to confirm that these two branches contain the same changes

abuts:
I have followed @hsautessella' advice and tried rebase rather them merge. The branch is created by rebasing #500 onto master, and necessary because one needs to test changes locally due to conflicts. Experimental -- will not do it again, as merge shows all conflicts together and rebase replays them commit by commit. 

 Rebase works better without conflicts and creates clearer history (nobody looks at this history anyway, but if somebody wants....) I believe, as the result of our discussion today we decided do a ticket to merge, branch for ticket and PR from the ticket -- this is how it would look (without rebase)

nickbattam-tessella:
It wasn't obvious that this branch was a rebase of the original PR branch.

Issue #514, opened 2021-01-14
Implement file-to-file cuts with temp files using file-backed pixel data


hsautessella:
As part of #441, large, out of memory, cuts need to be implemented. This is an issue to branch off the work done on [`441_implement_cut_with_file-backed_pixel_data`](https://github.com/pace-neutrons/Horace/tree/441_implement_cut_with_file-backed_pixel_data) and write the code to perform these cuts.

If possible, we should use the temporary file management implemented in `PixelData` so that we're not writing too much extra code.

Child of #441

PR #524, opened 2021-01-18
Implement file to file cuts with temp files using file backed pixel data

hsautessella:
This PR restores the ability to take out-of-memory cuts from an SQW file. This is useful when taking cuts where the output cut is too large to fit in memory. I've not implemented a new process for these types of cut, I've simply dropped in the old method using a `pix_combine_info` object.

The out-of-memory cuts work by:
1) Looping over input file's pixels
2) Retaining only the pixels in the range of the cut
3) Saving the pixels to be retained in a temporary file, along with an `npix` array that specifies which pixels belong in which bin (of the output cut)
4) Generating an SQW object (`sqw_cut_obj`) whose pixels are replaced with a `pix_combine_info` object (this holds a reference to the temp files and info required to combine them)
5) When `sqw_cut_obj` is saved, `put_sqw_data_pix_from_file_` iterates over the image bins of `sqw_cut_obj` and over the temporary files, loading pixels for the given bin from each temporary file and writing them to the output file (this is a very similar process to `gen_sqw`).
6) The temporary files can be deleted once the cut is complete.

To take an out-of-memory cut, you should call `cut` with no output arguments whilst specifying an output file to write the output cut to:
```matlab
cut(infile, proj, u_axis_lims, v_axis_lims, w_axis_lims, en_axis_lims, outfile);
```

_Note that you cannot perform an out-of-memory cut if you call `cut` with an output argument, hence `cut` can never return an SQW whose pixels are replaced with a `pix_combine_info` object._

I've also applied `clang-format` to a couple of C++ files, but have not changed their content otherwise (apologies for the extra files in the PR).

Addresses #514

### DIFF
--- a/horace_core/sqw/@sqw/private/cut_accumulate_data_.m
+++ b/horace_core/sqw/@sqw/private/cut_accumulate_data_.m
@@ -204,20 +204,27 @@ end
 function pci = init_pix_combine_info(nfiles, nbins)
     % Create a pix_combine_info object to manage temporary files of pixels
     wk_dir = get(parallel_config, 'working_directory');
-    gen_fpath = @(x) fullfile( ...
-        wk_dir, ['horace_subcut_', rand_digit_string(16), '.tmp'] ...
-    );
-    tmp_file_names = cellfun(gen_fpath, cell(1, nfiles), 'UniformOutput', false);
+    tmp_file_names = gen_array_of_tmp_file_paths(nfiles, wk_dir);
     pci = pix_combine_info(tmp_file_names, nbins);
 end
 
 
-function str = rand_digit_string(n)
-    % Create string of n random digits
-    rand_ints = randi([0, 9], [1, n]);
-    str = blanks(n);
-    for i=1:n
-        str(i) = int2str(rand_ints(i));
+function paths = gen_array_of_tmp_file_paths(nfiles, base_dir)
+    % Generate a cell array of paths for temporary files to be written to
+    % Format of the file names follows:
+    %   horace_cut_<UUID>_<counter_with_padded_zeros>.tmp
+    if nfiles < 1
+        error('CUT:cut_accumulate_data_', ...
+              ['Cannot create temporary file paths for less than 1 file.' ...
+               '\nFound %i.'], nfiles);
+    end
+    prefix = 'horace_cut';
+    uuid = char(java.util.UUID.randomUUID());
+    counter_padding = floor(log10(nfiles)) + 1;
+    format_str = sprintf('%s_%s_%%0%ii.tmp', prefix, uuid, counter_padding);
+    paths = cell(1, nfiles);
+    for i = 1:nfiles
+        paths{i} = fullfile(base_dir, sprintf(format_str, i));
     end
 end
 
--- a/horace_core/sqw/@sqw/private/cut_accumulate_data_.m
+++ b/horace_core/sqw/@sqw/private/cut_accumulate_data_.m
@@ -1,5 +1,5 @@
-function [s, e, npix, pix_out, urange_pix] = ...
-    cut_accumulate_data_(obj, proj, keep_pix, log_level)
+function [s, e, npix, pix_out, urange_pix, pix_comb_info] = ...
+    cut_accumulate_data_(obj, proj, keep_pix, log_level, return_cut)
 %%CUT_ACCUMULATE_DATA Accumulate image and pixel data for a cut
 %
 % Input:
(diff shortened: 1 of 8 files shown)


mducle:
> ```
> cut(infile, proj, u_axis_lims, v_axis_lims, w_axis_lims, en_axis_lims, outfile);
> ```
> Note that you cannot perform an out-of-memory cut if you call cut with an output argument, hence cut can never return an SQW whose pixels are replaced with a pix_combine_info object.

As a user, this syntax could be confusing to me... especially if I have a script already where I do in-memory cuts but then find that I don't have enough memory, and want to replace them with file-backed cuts. Is it possible to make this behaviour toggle-able by a flag instead of an absence of an requested output variable? E.g. `-filebacked` (similar to `-nopix`). Then the output file name can be returned instead of the `sqw` object? This could also obviate the need for the `outfile` input if the user doesn't care about what the file-backed store is called.

E.g. there would be two variants:

```
sqwfile = cut(infile, proj, u_axis_lims, v_axis_lims, w_axis_lims, en_axis_lims, '-filebacked')
```

or

```
sqwfile = cut(infile, proj, u_axis_lims, v_axis_lims, w_axis_lims, en_axis_lims, 'filebacked', true)
```

would generate a temporary file (in a temporary folder which is periodically cleaned up by the OS) and return that filename to the user... then can then use it in other `cut` commands etc. just like they would use `cut` on an in-memory object.

```
sqwfile = cut(infile, proj, u_axis_lims, v_axis_lims, w_axis_lims, en_axis_lims, 'filebacked', outfile)
```

or

```
cut(infile, proj, u_axis_lims, v_axis_lims, w_axis_lims, en_axis_lims, 'filebacked', outfile)
```

where `sqwfile == outfile`, where the user explicitly names a file to store the filebacked cut.

This syntax would make the Python interface a lot easier to handle since determining the equivalent of `nargout` in Python is a bit tricky.

hsautessella:
@mducle I think I, and the rest of the group, agree with you: having the flag would be nice. The syntax I described above mirrors how you would do an out-of-memory cut on master, so we're unlikely to change that for the moment, although we're all in favour of deprecating that syntax in favour of the explicit flag. 

I'll open a ticket to do that (as I think it's out of the scope of this PR) and reference it here.

abuts (commented):
I would add defined numbers to tmp cuts files names e.g. file_NNNN_1.tmp file_MMMM_2.tmp....  to simplify debugging if necessary. Otherwise looks ok.

one thing to pay attention to -- put_sqw should be on `PixelData` as this would be an algorithm, running range of codes, i.e. slow matlab code combining tmp files, mex code (different for hdf and binary files) and MPI job, which does combining on different processes. 

put_sqw should not be on a loader, which will be an interface to a different format files on disk.

horace_core/sqw/@sqw/private/cut_accumulate_data_.m
```diff
@@ -168,3 +199,43 @@
     s(no_pix) = 0;
     e(no_pix) = 0;
 end
+
+
+function pci = init_pix_combine_info(nfiles, nbins)
+    % Create a pix_combine_info object to manage temporary files of pixels
+    wk_dir = get(parallel_config, 'working_directory');
+    gen_fpath = @(x) fullfile( ...
+        wk_dir, ['horace_subcut_', rand_digit_string(16), '.tmp'] ...
```
cmarooney-stfc, line 208: This does not look as if it guarantees all strings to be different. Highly likely that they are but not certain.

horace_core/sqw/@sqw/private/cut_accumulate_data_.m
```diff
@@ -168,3 +199,43 @@
...
+    );
+    tmp_file_names = cellfun(gen_fpath, cell(1, nfiles), 'UniformOutput', false);
+    pci = pix_combine_info(tmp_file_names, nbins);
+end
+
+
+function str = rand_digit_string(n)
+    % Create string of n random digits
+    rand_ints = randi([0, 9], [1, n]);
```
cmarooney-stfc, line 217: I can't actually see this combination of arguments in the randi docs - could you point me to it?

cmarooney-stfc (commented):
Couple of comments please

horace_core/sqw/@sqw/private/cut_accumulate_data_.m
hsautessella:
```matlab
randi([imin, imax], sz)
```
`sz` can be a row vector of integers:
https://www.mathworks.com/help/matlab/ref/randi.html#buf2csg-sz
> sz — Size of each dimension (as a row vector)
> integer values

So the call generates a `[1, n]` array of random integers between 0 and 9 (inclusive).

horace_core/sqw/@sqw/private/cut_accumulate_data_.m
nickbattam-tessella: If absolute uniqueness is required `java.util.UUID.randomUUID()`; would provide that at the cost of a longer string.

nickbattam-tessella (approved):

Issue #484, opened 2020-12-14
Create ADR for C vs C++ mex API decision


nickbattam-tessella:
Create ADR to record decision to continue using the C-MEX API

This should include the discussed reasons why this is preferable to the C++ API for the Horace use-case.

PR #530, opened 2021-01-21
Add ADR for decision to prefer C Mex API over C++

hsautessella:
This PR adds an ADR explaining our decision to prefer the C Mex API over the C++. 

I've left a TODO in the ADR in the hope that a reviewer can provide an example of where the C++ API was too restrictive for our use-case. I know we discussed such a case existing, does anyone have a concrete example of this?

Addresses #484

### DIFF
--- a/documentation/adr/0020-use-c-mex-api.md
+++ b/documentation/adr/0020-use-c-mex-api.md
@@ -31,8 +31,8 @@ Herbert/Horace will continue to use the C API for new Mex functions.
 
 ## Consequences
 
-- The C++ code we write will be compatible with all Matlab releases,
-  including those older than R2018a.
+- The C++ code we write will be compatible with all[<sup>[1]</sup>](#ref_1)
+  Matlab releases, including those older than R2018a.
 - The C++ code we have written will be consistent with new code we write.
 - We will not get the benefits of the newer C++ Mex API
   (e.g. smart pointers, [C++ class wrappers](https://www.mathworks.com/help/matlab/matlab-data-array.html?s_tid=CRUX_lftnav)).
@@ -41,3 +41,19 @@ _Note: this decision need not be binding,
 but provides a guideline for us to be consistent.
 If a use-case arises where using the C++ API is advantageous,
 then there should be nothing stopping us from using it._
+
+---
+
+<a name="ref_1">
+</a><sup>[1]</sup> This comes with a caveat when using complex numbers.
+In R2018a Matlab introduced the
+["interleaved" complex API](https://www.mathworks.com/help/matlab/matlab_external/matlab-support-for-interleaved-complex.html),
+which stores complex numbers as
+[real_1, imag_1, real_2, imag_2, ..., real_n, imag_n].
+This differs in releases R2017b and earlier, where complex numbers were stored
+in two separate arrays, once for the real parts one for the imaginary.
+You switch between which API to use in Mex files by defining the macro
+`MATLAB_DEFAULT_RELEASE` to be `R2017b` or `R2018b`.
+Or by passing the `-R2017b` or `-R2018b` flag to Matlab's `mex` command.
+To remain compatible with older Matlab versions we must compile with the
+`-R2017b` flag.
--- a/documentation/adr/0020-use-c-mex-api.md
+++ b/documentation/adr/0020-use-c-mex-api.md
@@ -50,19 +50,21 @@ then there should be nothing stopping us from using it._
 
 ---
 
-<a name="ref_1">
-</a><sup>[1]</sup> This comes with a caveat when using complex numbers.
-In R2018a Matlab introduced the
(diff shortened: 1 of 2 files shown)


documentation/adr/0020-use-c-mex-api.md
```diff
@@ -0,0 +1,43 @@
...
+
+- The C++ code we write will be compatible with all Matlab releases,
+  including those older than R2018a.
+- The C++ code we have written will be consistent with new code we write.
+- We will not get the benefits of the newer C++ Mex API
+  (e.g. smart pointers, [C++ class wrappers](https://www.mathworks.com/help/matlab/matlab-data-array.html?s_tid=CRUX_lftnav)).
+
+_Note: this decision need not be binding,
+but provides a guideline for us to be consistent.
+If a use-case arises where using the C++ API is advantageous,
```
nickbattam-tessella, line 42: To justify this (worthwhile) note we need to add that the C and C++ APIs can be used in the same suite of tools (i.e. PACE) to the Context section. 

Can a mix of the C and C++ APIs be used in the same project or would we need build pipelines for the two APIs?

oerc0122: C/C++ can be used in the same project with no problem whatever beyond maintenance.

documentation/adr/0020-use-c-mex-api.md
```diff
@@ -0,0 +1,43 @@
...
+TODO: do we have examples of where the C++ API did not work for our use-case?
+
+## Decision
+
+Herbert/Horace will continue to use the C API for new Mex functions.
+
+## Consequences
+
+- The C++ code we write will be compatible with all Matlab releases,
+  including those older than R2018a.
```
mducle, line 35: Add additional distinction is between the "interleaved complex API" and the "separate complex API" - essentially whether complex matrices are stored in memory as separate real and imaginary matrices ("separate") or as a single matrix with elements being a complex data type ("interleave") - this was also introduced in R2018a. If we want to support Matlab versions before R2018a, we have to use the "separate complex API"... not sure if the Horace mex files uses complex numbers or not though, so this could be moot...

hsautessella: Good point, I'll add that we'll have to continue using the separate complex API to maintain backwards compatibility where we use complex numbers. The only place I can see this being an issue is the C++ serializer (I think that's the only place we worry about complex numbers), would this be an issue @oerc0122?

Do you know if using the C++ Data API necessarily means using the newer "interleaved complex API"? I can't seem to find anything in the docs that says one way or the other. I had a quick check and I seem to be able to run a Mex file using the C++ Data API compiled with the `-R2017b` flag.

oerc0122: There is absolutely no problem there and it's accounted for through the `MEX_SUPPORTS_INTERLEAVED_COMPLEX` or similar conditional compilation flag. I think the C++ API is always interleaved (due to native complex support in C++ not in C), but I may be wrong.

documentation/adr/0020-use-c-mex-api.md
```diff
@@ -41,3 +47,19 @@ _Note: this decision need not be binding,
...
+["interleaved" complex API](https://www.mathworks.com/help/matlab/matlab_external/matlab-support-for-interleaved-complex.html),
+which stores complex numbers as
+[real_1, imag_1, real_2, imag_2, ..., real_n, imag_n].
+This differs in releases R2017b and earlier, where complex numbers were stored
+in two separate arrays, once for the real parts one for the imaginary.
+You switch between which API to use in Mex files by defining the macro
+`MATLAB_DEFAULT_RELEASE` to be `R2017b` or `R2018b`.
+Or by passing the `-R2017b` or `-R2018b` flag to Matlab's `mex` command.
+To remain compatible with older Matlab versions we must compile with the
+`-R2017b` flag.
```
oerc0122, line 65: Or use the `#if MX_HAS_INTERLEAVED_COMPLEX` preprocessor to support both. However, if this is not done everywhere we may run into problems and drop support for either one accidentally

hsautessella: I'll add this to the note 👍

oerc0122 (approved):

documentation/adr/0020-use-c-mex-api.md
mducle: Yes, I think the C++ Data API is always interleaved. For the separate complex API there is a C function (`mxGetPi` [compared with `mxGetPr` for the real part) to get a pointer to the imaginary matrix and the equivalent function in C++ does not exist.

I think the ADR looks fine now. 

(But, just a note for future... I think that internally within newer versions of Matlab complex numbers are stored in the interleaved format; so mex files which use the old API, whilst compatible with older versions of Matlab will suffer a performance penalty on newer versions of Matlab because Matlab will do an internal copy from the interleaved format to the separate format and vice versa on return from the mex file; This is one reason that for the update of SpinW we're targeting Matlab >= R2018a and using the C++ Data API.)

PR #536, opened 2021-01-26
455 jenkins docs pipeline

oerc0122:
Create pipelines for the automatic deployment of documentation nightly. 

Need decision on which build is going to push.


tools/build_config/Jenkinsfile
```diff
@@ -359,6 +359,61 @@ pipeline {
       }
     }
 
+    stage('Build-Docs') {
+      steps {
+        script {
+          if (isUnix()) {
+            sh '''
```
nickbattam-tessella, line 366: these code blocks (both bash and powershell) should be extracted to helper scripts in `tools/bash` and `tools/pwsh`

Ideally this should just be `module load python/3.6 && build_docs.sh` or similar.

tools/build_config/Jenkinsfile
```diff
@@ -403,6 +458,53 @@ pipeline {
...
+
+    stage('Update docs') {
+      when {
+        expression {env.RELEASE_TYPE == "nightly"}
+      }
+      steps {
+        withCredentials([string(credentialsId: 'GitHub_API_Token',
+                                variable: 'api_token')]) {
+          script {
+            if (isUnix()) {
```
nickbattam-tessella, line 471: Similarly I'm not sure these large code blocks should be contained in the main Jenkinsfile, as there is no way for a developer to run the same operation outside of Jenkins. 

I understand this is a more subtle case to the build above, the git config and remote set-url are clearly specific to the Jenkinsfile setting up the git environment.

tools/build_config/build.ps1
```diff
@@ -36,6 +36,10 @@ param (
   [switch][Alias("p")]$package,
   # Print the versions of libraries being used e.g. Matlab.
   [switch][Alias("v")]$print_versions,
+  # Build docs
+  [switch][Alias("d")]$docs,
+  # Push docs to github
+  [switch]$push_docs,
```
nickbattam-tessella, line 42: Is not  having an alias for push deliberate to make it unlikely to happen by accident?

tools/build_config/build.ps1
```diff
@@ -188,6 +192,43 @@ function Invoke-Package {
   }
 }
 
+function Invoke-Build-Docs {
+  (Get-Content "./build/CPackConfig.cmake" | Where-Object {$_ -match 'CPACK_PACKAGE_FILE_NAME'}) -match '.*"Horace-([^"]+)".*'
```
nickbattam-tessella, line 196: what is this doing?!

tools/build_config/build.sh
```diff
@@ -85,6 +85,41 @@ function run_package() {
   echo_and_run "cpack -G TGZ"
 }
 
+function build_docs() {
+    # Update release numbers
+    build_id=$(sed -nr '/CPACK_PACKAGE_FILE_NAME/{s/.*"Horace-([^"]+)".*/\\1/p};' ./build/CPackConfig.cmake)
+    sed -i -r 's/release = .*/release = "'${build_id}'"/' ./documentation/user_docs/docs/conf.py
```
nickbattam-tessella, line 91: can `conf.py` be updated pull this data from file rather than performing a temporary `sed` update?

tools/build_config/build.ps1
```diff
@@ -188,6 +192,43 @@ function Invoke-Package {
...
+  }
+  Compress-Archive -Path ./documentation/user_docs/build/html/* -DestinationPath "./docs.zip"
+  if ($LASTEXITCODE -ne 0) {
+    exit $LASTEXITCODE
+  }
+}
+
+function Invoke-Push-Docs {
+  git config --local user.name "PACE CI Build Agent"
+  git config --local user.email "pace.builder.stfc@gmail.com"
```
nickbattam-tessella, line 216: should the username and url be passed in rather than hard coded in the script?

tools/build_config/build.sh
```diff
@@ -85,6 +85,41 @@ function run_package() {
...
+    cd ./documentation/user_docs/build/html
+    tar -czf docs.tar.gz ./*
+    cd -
+    mv ./documentation/user_docs/build/html/docs.tar.gz .
+}
+
+function push_built_docs() {
+    git config --local user.name "PACE CI Build Agent"
+    git config --local user.email "pace.builder.stfc@gmail.com"
+    git remote set-url --push origin "https://pace-builder:"\${api_token## }"@github.com/pace-neutrons/Horace"
```
nickbattam-tessella, line 111: should the URL and email be passed in rather than hard-coded?

The powershell script trims the token before the `@`, this explicitly has `## `; is this difference platform dependent?

nickbattam-tessella (commented):
Should the `--push-docs` function check that built docs exist and abort quickly if that folder is missing?

tools/build_config/build.sh
```diff
@@ -153,6 +188,8 @@ function main() {
         -t|--test) test=$TRUE; shift ;;
         -a|--analyze) analyze=$TRUE; shift ;;
         -p|--package) package=$TRUE; shift ;;
+        -d|--docs) docs=$TRUE; shift;;
```
hsautessella, line 191: Could you add this flag to the message printed by `print_help`?

tools/build_config/build.ps1
```diff
@@ -188,6 +192,43 @@ function Invoke-Package {
...
+  $build_id = $Matches[1]
+  (Get-Content ./documentation/user_docs/docs/conf.py) | Foreach-Object {$_ -replace 'release = .*', "release = '${build_id}'"} | Set-Content  ./documentation/user_docs/docs/conf.py
+
+  ./documentation/user_docs/make.bat html
+  ./documentation/user_docs/make.bat html
+  # Undo change to allow checkout
+  git checkout ./documentation/user_docs/docs/conf.py
+
+  Foreach ($f in Get-ChildItem -Path './documentation/user_docs/build/html' -Filter *.html) {
+      (Get-Content "./documentation/user_docs/build/html/$f") | Where-Object {$_ -notmatch '\\[NULL\\]'} | Set-Content "./documentation/user_docs/build/html/$f"
```
hsautessella, line 206: This line is quite long, can we split the lines on the pipes? The line continuation for PowerShell is a backtick

tools/build_config/build.ps1
```diff
@@ -188,6 +192,43 @@ function Invoke-Package {
   }
 }
 
+function Invoke-Build-Docs {
+  (Get-Content "./build/CPackConfig.cmake" | Where-Object {$_ -match 'CPACK_PACKAGE_FILE_NAME'}) -match '.*"Horace-([^"]+)".*'
+  $build_id = $Matches[1]
+  (Get-Content ./documentation/user_docs/docs/conf.py) | Foreach-Object {$_ -replace 'release = .*', "release = '${build_id}'"} | Set-Content  ./documentation/user_docs/docs/conf.py
+
+  ./documentation/user_docs/make.bat html
+  ./documentation/user_docs/make.bat html
```
hsautessella, line 201: Is it necessary to run `make` twice? I see you're doing the same in Bash. If it is necessary, a comment explaining why would be good.

tools/build_config/build.sh
oerc0122: It's a lazy in-place trim

`${var## }` and `${var%% }` are bash commands for stripping whitespace,  `#` -> strip front, `%` strip back, when doubled repeat operation. Perhaps best explored with 
```bash
var='test.tar.gz'
echo ${var%.*}
echo ${var%%.*}
```
To get the basename.

tools/build_config/build.sh
oerc0122: I think so, it is basically a python script, I'll have a play, but it will require users to have used the CMake pathway.

tools/build_config/build.ps1
oerc0122: Possible, I suppose, but it's hardcoded in the API token anyway.

tools/build_config/build.ps1
oerc0122: Pulling the build details from `build/CPackConfig.cmake` i.e. the version/SHA/etc.

tools/build_config/build.sh
```diff
@@ -85,6 +85,41 @@ function run_package() {
   echo_and_run "cpack -G TGZ"
 }
 
+function build_docs() {
+    # Update release numbers
+    build_id=$(sed -nr '/CPACK_PACKAGE_FILE_NAME/{s/.*"Horace-([^"]+)".*/\\1/p};' ./build/CPackConfig.cmake)
```
hsautessella, line 90: Is `build_id` the right name for this variable? Aren't you pulling the version here?

tools/build_config/build.ps1
oerc0122: Deliberate, sensible choices would have been `-p` or `-d` and those would lead too easily to typos, and users shouldn't be doing it anyway.

tools/build_config/build.sh
oerc0122: **N.B.** `.*` there is not RegEx match anything, but bash style globbing and a literal dot, anything equivalent to `\..*` in RegEx

tools/build_config/build.ps1
oerc0122: Possibly a leftover habit from LaTeX paranoia. I'm not sure if it's necessary, I will do some testing.

tools/build_config/build.sh
oerc0122: Depends. If it is the nightly, I think it's the SHA/date, which isn't the version?

documentation/user_docs/docs/conf.py
```diff
@@ -25,6 +25,7 @@
 
 
 # The full version, including alpha/beta/rc tags
+version = '3.5.0'
```
nickbattam-tessella, line 28: these version numbers shouldn't be hard-coded in the the Python.

The `VERSION` file in the main project source tree is the reference version number;  this was created so we didn't have the same version number stored in lots of places across the source. Could this pull it from there? There are routines in the CMake scripts that use this to create the "fully decorated" version number that goes into Matlab and C files.

oerc0122: These numbers are overwritten by the build, but true. I could use a CMake template to create this file and not include it in the docs, but that means you can't build the docs without running CMake first (not too unreasonable).

oerc0122: That would also mean that I don't need quite as much sed magic.

tools/build_config/Jenkinsfile
```diff
@@ -359,6 +369,42 @@ pipeline {
       }
     }
 
+    stage('Build-Docs') {
+      when {
+        expression {env.build_docs?.toBoolean()}
```
hsautessella, line 374: Should this be `build_docs || push_docs`? Or can we push the docs without building?

tools/build_config/build.ps1
```diff
@@ -188,6 +192,57 @@ function Invoke-Package {
...
+function Invoke-Docs {
+    (Get-Content "./build/CPackConfig.cmake" |
+      Where-Object {$_ -match 'CPACK_PACKAGE_FILE_NAME'}) -match '.*"Horace-([^"]+)".*'
+  $build_id = $Matches[1]
+
+  (Get-Content ./documentation/user_docs/docs/conf.py) |
+    Foreach-Object {
+        $_ -replace '(version|release) = .*', "`$1 = '${build_id}'"
+    } |
+      Set-Content  ./documentation/user_docs/docs/conf.py
```
nickbattam-tessella, line 204: As you're setting the `version=` and `release=` lines in `conf.py` can you remove the "plausible" placeholder values currently in that file (`3.5.0`) and replace them with something that is clearly a placeholder?

tools/build_config/build.ps1
```diff
@@ -216,3 +271,11 @@ if ($test -eq $true) {
...
   Invoke-Package -build_dir "$build_dir"
 }
+
+if ($docs -eq $true) {
+  Invoke-Docs
+}
+
+if ($push_docs -eq $true) {
+  Invoke-Push
+}
```
hsautessella, line 281: I realise you were just following what was done previously, but these `-eq $true`s can be removed from the `if` statements

tools/build_config/build.sh
```diff
@@ -85,6 +85,41 @@ function run_package() {
   echo_and_run "cpack -G TGZ"
 }
 
+function build_docs() {
+    # Update release numbers
+    build_id=$(sed -nr '/CPACK_PACKAGE_FILE_NAME/{s/.*"Horace-([^"]+)".*/\1/p};' ./build/CPackConfig.cmake)
+    sed -i -r 's/(release|version) = .*/\1 = "'${build_id}'"/' ./documentation/user_docs/docs/conf.py
+    make -C ./documentation/user_docs/ html
```
nickbattam-tessella, line 92: this `make` call is the actual build so would benefit from some whitespace above it, at the moment it looks like part of the 'update release numbers' operation and is easily overlooked

tools/build_config/Jenkinsfile
```diff
@@ -403,6 +449,27 @@ pipeline {
         )
       }
     }
+
+
+    stage('Update docs') {
```
hsautessella, line 454: Should this be `Push-Docs` to be consistent with the parameter naming?

tools/build_config/Jenkinsfile
```diff
@@ -156,6 +156,16 @@ properties([
       description: 'The version of CppCheck tooling to load to provide the code-style checks.',
       name: 'CPPCHECK_VERSION',
       trim: true
+    ),
+    booleanParam(
+      defaultValue: false,
+      description: 'If this pipeline is to build docs.',
+      name: 'build_docs'
```
hsautessella, line 163: Our convention for parameters has been all caps

hsautessella (commented):
Looks good, just a couple of comments

tools/build_config/Jenkinsfile
oerc0122: I went for the opposite way round, if you push docs without building, it will do nothing. Perhaps that was a bad choice.

tools/build_config/build.ps1
oerc0122: Does that mean you want all of them removing?

cmake/PACE_Docs.cmake
```diff
@@ -13,7 +13,7 @@ if (sphinx-build)
       )
   else()
     add_custom_target(denullify-docs
-      COMMAND Foreach($f in Get-ChildItem -Path '${DOCS_DIR}/build/html' -Filter *.html) {(Get-Content $f.FullName) | Where-Object {$_ -notmatch '\[NULL\]'} | Set-Content $F.FullName}
+      COMMAND powershell -command "Foreach($f in Get-ChildItem -Path '${DOCS_DIR}/build/html' -Filter *.html) {(Get-Content $f.FullName) | Where-Object {$_ -notmatch '\[NULL\]'} | Set-Content $F.FullName}"
```
hsautessella, line 16: I'd add `-ExecutionPolicy Bypass` to the PowerShell command https://stackoverflow.com/questions/56726926/running-powershell-script-with-cmake-through-visual-studio-without-changing-syst

cmake/PACE_Docs.cmake
```diff
@@ -0,0 +1,65 @@
+# Build documentation
+
+set(DOCS_DIR "${Horace_ROOT}/documentation/user_docs" CACHE STRING "Directory containing user documentation")
```
hsautessella, line 3: Does this need to be a cache variable? I'd usually only make a variable a cache variable if I wanted to make it configurable

cmake/PACE_Docs.cmake
```diff
@@ -0,0 +1,65 @@
...
+find_program(latexmk NAMES latexmk)
+if (sphinx-build)
+  if (UNIX)
+    add_custom_target(denullify-docs
+      COMMAND sed -i -r "/\[NULL\]/d" "${DOCS_DIR}/build/html/*html"
+      DEPENDS build-docs
+      )
+  else()
+    add_custom_target(denullify-docs
+      COMMAND powershell -ExecutionPolicy Bypass -command "Foreach($f in Get-ChildItem -Path '${DOCS_DIR}/build/html' -Filter *.html) {(Get-Content $f.FullName) | Where-Object {$_ -notmatch '\\[NULL\\]'} | Set-Content $F.FullName}"
```
hsautessella, line 16: Splitting this command over a few lines would make it more readable:
```suggestion
      COMMAND
        powershell
          -ExecutionPolicy Bypass
          -command "\
            Foreach($f in Get-ChildItem -Path '${DOCS_DIR}/build/html' -Filter *.html) { \
              (Get-Content $f.FullName) \
                | Where-Object {$_ -notmatch '\\[NULL\\]'} \
                | Set-Content $F.FullName \
            }"
```

cmake/PACE_Docs.cmake
```diff
@@ -0,0 +1,65 @@
...
+    )
+
+  add_custom_target(docs
+    DEPENDS build-docs
+    DEPENDS denullify-docs
+    )
+
+  if (pdflatex AND latexmk)
+    add_custom_command(OUTPUT horace.tex
+      COMMAND ${sphinx-build} -b latex "${DOCS_DIR}/docs" "${DOCS_DIR}/build/latex" ${SPHINX_OPTS} -D "release=${${PROJECT_NAME}_SHORT_VERSION}" -D "version=${${PROJECT_NAME}_SHORT_VERSION}"
```
hsautessella, line 35: This is another command that could benefit from some line breaks
```suggestion
      COMMAND 
        ${sphinx-build}
          -b latex "${DOCS_DIR}/docs" "${DOCS_DIR}/build/latex" ${SPHINX_OPTS}
          -D "release=${${PROJECT_NAME}_SHORT_VERSION}"
          -D "version=${${PROJECT_NAME}_SHORT_VERSION}"
      WORKING_DIRECTORY "${DOCS_DIR}"
```

cmake/PACE_Docs.cmake
```diff
@@ -0,0 +1,65 @@
+# Build documentation
+
+set(DOCS_DIR "${Horace_ROOT}/documentation/user_docs" CACHE STRING "Directory containing user documentation")
+
+find_program(sphinx-build NAMES sphinx-build)
+find_program(pdflatex NAMES pdflatex)
+find_program(latexmk NAMES latexmk)
+if (sphinx-build)
+  if (UNIX)
+    add_custom_target(denullify-docs
```
hsautessella, line 10: Is it possible to perform some of these processes using `add_custom_command`? I think we're exposing more build targets than we need to here. I think we only really need the `docs` and `manual` processes to be targets

hsautessella (changes requested):
Builds locally and the docs look good. I think it was a good choice to define the build in CMake.

Issue #538, opened 2021-01-27
PixelData test failing due to temporary file move


hsautessella:
The place where `PixelData` objects hold their temporary files changed from `tempdir()` to `get(parallel_config, 'working_directory')` in 14978979eac0f2762d137b2c51d29c02fcccfd37. This went undetected at first because the two directories are often the same, but many Windows builds are now failing because some `PixelData` tests are still looking for the temporary files in `tempdir()`.

Failing tests are: 

- [`test_PixelData::test_unedited_dirty_pages_are_not_rewritten`](https://github.com/pace-neutrons/Horace/blob/03f1d1a3c007ea413e97efe5ffabf774a8e093cf/_test/test_sqw/test_PixelData.m#L887) 
- [`test_PixelData::test_dirty_pix_tmp_files_are_deleted_when_pix_out_of_scope`](https://github.com/pace-neutrons/Horace/blob/03f1d1a3c007ea413e97efe5ffabf774a8e093cf/_test/test_sqw/test_PixelData.m#L596).

**Output:**
- Above tests passing even if `tempdir() ~= get(parallel_config, 'working_directory')`

PR #539, opened 2021-01-27
PR: PixelData test failing due to temporary file move

hsautessella:
A couple of `PixelData` tests have been updated to look for pixel temporary files in the correct place.

Addresses #538


oerc0122 (approved):

_test/test_sqw/test_PixelData.m
```diff
@@ -597,7 +597,10 @@ function test_dirty_pix_tmp_files_are_deleted_when_pix_out_of_scope(obj)
             old_rng_state = rng();
             fixed_seed = 774015;
             rng(fixed_seed, 'twister');  % this seed gives an expected object_id_ = 54452
-            expected_tmp_dir = fullfile(tempdir(), 'sqw_pix54452');
+            expected_tmp_dir = fullfile( ...
+                get(parallel_config, 'working_directory'), ...
+                'sqw_pix54452' ...
+            );
             clean_up = onCleanup(@() rng(old_rng_state));
```
cmarooney-stfc, line 604: Shouldn't the onCleanup call precede the rng call on L.599? Ditto at L.899

nickbattam-tessella: They will work fine where they are: the `old_rng_state` variable is set as the rng is reseeded. This call creates function pointer that is executed at the end of the method to restore that.

nickbattam-tessella (approved):

_test/test_sqw/test_PixelData.m
cmarooney-stfc: @nickbattam-tessella What if rng or fullfile errors?

hsautessella: I've moved the call up 👍

Issue #312, opened 2020-07-29
Remove support for legacy multifit and tobyfit


nickbattam-tessella:
The code currently includes "legacy" implementations of Tobyfit and Multifit.
Both are clearly marked with a deprecated warning:
```
%   If you are still using the legacy version then it is strongly recommended
%   that you change to the new operation. Help for the legacy operation can
%   be <a href="matlab:help('sqw/multifit_legacy_sqw');">found here</a>]
```

These, and the supporting infrastructure should be removed from the codebase.

nickbattam-tessella:
Removal confirmed by Toby via email on 29-Jul-2020

nickbattam-tessella:
These will be removed as (new) fit are moved into the new-style class SQW folder.

Part of #288

cmarooney-stfc:
@nickbattam-tessella Hi Nick, Checking this should still be tagged blocked since you have moved it into In Progress. Thanks Chris

cmarooney-stfc:
@nickbattam-tessella Hi Nick, noting you have moved this back into sprint backlog, but also that PR 313 is also outstanding albeit marked do-not-merge. What is the status of the PR? thanks

cmarooney-stfc:
MOved this ticket back to sprint backlog as the complete removal of legacy has not yet been done. What has been done is in pace-neutrons/Herbert#392 which is a PR to deprecate use of legacy at the point of use. The larger removal task will be prioritised later

PR #540, opened 2021-01-27
508 Add fit functions to new sqw object

cmarooney-stfc:
Add fit functions to new sqw. Place some old sqw private methods, mostly plotting, variously into new sqw public methods, new sqw_graphics folder under horace_core/sqw, or horace_core/Tobyfit. Make the tests in test_tobyfit/*.m and test_sqw/test_sqw_dnd_eval.m work with new_sqw. Note that everything moved has been moved to make these tests work, so I am presuming that no additional tests are required. There may be other methods in sqw_old that will need moving for other *fit operations on new sqw, but at this stage we have no use cases for them. 

For these tests, there are existing stored old-style sqw objects in .mat files. It turns out that while individual sqw objects can be read from them and then converted into proper new-style sqw objects, stored arrays will only load the first element. Hence these have been loaded locally outside this PR into MATLAB and re-saved as individual (old-style) sqw objects - in the tests, these objects in modified .mat files are loaded instead.


horace_core/sqw/@sqw/multifit_func.m
```diff
@@ -0,0 +1,72 @@
...
+%   <#file:>  <doc_fit_functions>
+%
+% See also multifit_sqw multifit_sqw_sqw
+%
+%   <#file:>  <doc_multifit_legacy_footnote>  <class_name>/<method_name_legacy>
+% <#doc_end:>
+%-------------------------------------------------------------------------------
+
+
+if ~mfclass.legacy(varargin{:})
```
nickbattam-tessella, line 67: Have you intentionally left the legacy calls here for deletion as part of a separate ticket (see #312)

horace_core/sqw/@sqw/noisify.m
```diff
@@ -0,0 +1,39 @@
...
+%           a point is the value of pixel signal.
+%
+% Additional developer options are possible in varargin to test this
+% functionality. See the PixelData noisify paging method (called below)
+% and the Herbert noisify function which implements the noise addition.
+%=========================================================================
+
+wout=w;
+for i=1:numel(w)
+    if isa(w(i),'sqw')   % determine if sqw or dnd type
```
nickbattam-tessella, line 28: this is an SQW class method, do you need the check for DnD ness? The DnD version belongs on DnDBase

horace_core/sqw/@sqw/resolution_plot.m
```diff
@@ -0,0 +1,231 @@
+function varargout = resolution_plot (w, varargin)
```
nickbattam-tessella, line 1: should this be in the plot folder, or is this transitory? It has detailed knowledge of specific instruments that don't belong in the core SQW class

nickbattam-tessella (commented):
- can you `skipTest` rather than `DISABLE_` them now that is available in the framework?
- do we need a set of `IX_dataset_nD` methods on the SQW class? That breaks the encapsulation of the dimension (in both SQW and DnDs) and they are graphics data objects

horace_core/sqw/@sqw/sqw.m
```diff
@@ -79,8 +80,21 @@
...
+               obj = S;
+               return
+            end
+            if numel(S)>1
+               tmp = sqw();
+               obj = repmat(tmp, size(S));
+               for i = 1:numel(S)
+                   obj(i) = sqw(S(i));
+               end
+               return;
```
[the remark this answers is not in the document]
cmarooney-stfc: agree

horace_core/sqw/@sqw/sqw.m
```diff
@@ -79,8 +80,21 @@
             % -------
             % Output:
             %   obj     An instance of this object
+            if isa(S,'sqw')
+               obj = S;
+               return
+            end
+            if numel(S)>1
+               tmp = sqw();
```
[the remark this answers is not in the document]
cmarooney-stfc: I can, but then the intent that tmp is just a placeholder is obscured.

horace_core/sqw/@sqw/multifit_func.m
cmarooney-stfc: Good point, but if I remove the legacy else while the legacy deletions haven't been done, then this method will fail if used on them. Better to remove mfclass.legacy_call and then mop up the failures.

horace_core/sqw/@sqw/resolution_plot.m
cmarooney-stfc: At the moment the detailed instrument stuff still exists in @sqw i.e. methods tobyfit_DG${inst}_*. It would be better to move all that out systematically when the conversion of sqw_old to sqw is complete.

horace_core/sqw/@sqw/sqw.m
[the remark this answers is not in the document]
cmarooney-stfc: done.

Replying to general comment in review about position of IX_dataset_*d methods. If they get moved, the code still runs but the plots are blank. I don't think it's worth digging to find out why just now.

horace_core/sqw/@sqw/noisify.m
nickbattam-tessella: Reviewing my previous comment ... should this be a check for `has_pixels(w(i))`? No DnD objects (should) be passed in here, and if a pixel-less SQW object is called the `noisify`/`recompute_bin_data` action will fail.

_test/test_sqw_class/test_migrated_apis.m
```diff
@@ -305,7 +305,9 @@ function test_value(obj)
         end
 
         %% xye
-        function test_xye_returns_bin_centres_and_errors(obj)
+        %DISABLED to pass the tests in the absence of the .mat
+        %         file - to be restored when they are reunited
+        function DISABLED_test_xye_returns_bin_centres_and_errors(obj)
```
nickbattam-tessella, line 310: `skipTest( ... )` can be used in place of `DISABLED_` now

_test/test_tobyfit/test_let/test_tobyfit_let_resfun_1.m
```diff
@@ -111,7 +111,7 @@ function test_tobyfit_let_resfun_1 (option)
 
 % q-e plot along [0,0,1]
 % ----------------------
-plot(w_nb_qe)
+iplot(w_nb_qe)
```
nickbattam-tessella, line 114: Is `sqw.iplot` a rename of `sqw_old.plot`?

nickbattam-tessella (commented):
Why has a migration of the fit algorithms needed the migration of the plot routines? 
Other than the rename (plot->iplot), have these been done in a way that will be compatible with existing user code?

There are a lot of `sqw(struct(sqw_obj)` and `sqw_old(struct(sqw_obj))` type conversions in the code. What is the plan to remove these? 

Are these conversions that could be done through a minor extension to the `manage_legacy_sqw_class_rename` code (shared from Herbert tests), adding sqw->sqw_old and sqw_old (obj) -> sqw (obj)? 

That would make removing it once the migration is complete much simpler as there would be a single function call to find.

_test/test_tobyfit/test_let/test_tobyfit_let_resfun_1.m
cmarooney-stfc: Yes

cmarooney-stfc:
@nickbattam-tessella Current master behaviour for the plot functions is 
- sqw ones live in horace_core/herbert/@sqw
- the plot itself is handled by a IX_dataset_2d constructed from the sqw
- additional aspect ratio stuff for the plot is made by extracting the data from the sqw 
- equivalent dnd ones live in horace_core/herbert/@d[n]d
- they work by converting their dnd into an sqw and running the sqw version
They do not look like they should be parts of an sqw. Hence it may be desirable to get them out of a sqw-type class.
Alternatively the original behaviour "might" be restored by reconverting the herbert/@sqw_old folders back to @sqw.
Happy to hear your thoughts on this.

nickbattam-tessella (approved):

PR #543, opened 2021-01-28
Create v3.6.0 release notes

oerc0122:
Hastily add release notes for merged stuff


nickbattam-tessella (approved):

documentation/release_notes/v3.6.0
```diff
@@ -0,0 +1,11 @@
+# Release Notes v3.6.0
+
+
+
+### Critical changes
+ - `hlp_(de)serialize` and `hlp_serial_size` rewritten and now called via `(de)serialise` and `serial_size` wrappers which call relevant MEX/non-MEX functions based on config.
+ 
+### Minor changes
+ - Added `skipTest` functionality to test suite for better reporting of test features
```
hsautessella, line 9: Is this change relevant to our users?

documentation/release_notes/v3.6.0
```diff
@@ -0,0 +1,11 @@
+# Release Notes v3.6.0
+
+
+
+### Critical changes
+ - `hlp_(de)serialize` and `hlp_serial_size` rewritten and now called via `(de)serialise` and `serial_size` wrappers which call relevant MEX/non-MEX functions based on config.
```
hsautessella, line 6: It would be good to address the main purpose of this change, i.e. to improve the performance of object serialization and the performance of MPI jobs. I think this is what is most relevant to users - especially because I don't see a use-case for users using this serializer directly

hsautessella (changes requested):

documentation/release_notes/v3.6.0
oerc0122: I would say that some users will notice a "N skipped tests" in their test log if they are observant and might question where it comes from.

documentation/release_notes/v3.6.0
oerc0122: Good point. Will update.

documentation/release_notes/v3.6.0
hsautessella: Most users won't be running tests though, particularly as the test framework is not part of the release package

oerc0122: Is it not? That surprises me. I guess the other question I have is whether these release notes are purely for the users or also for devs to look back on.

hsautessella: In my view, release notes are for users. They are displayed alongside releases on the download page and update users on what is new in each release. To encourage users to actually read them, I think the notes should be as brief possible and not include any information not relevant to them.

They could be useful to developers as an easy way to tell what is in what release, but Git and tags give us this.

hsautessella (approved):

PR #546, opened 2021-02-02
parpool failure

abuts:
This fixes the issue, related to the failure of parpool MPI tests on the windows node. The reason for that is actually not windows but the fact that only windows runs 4 parallel workers and Unix has, I believe 2. Independent parallel workers were numbering independent runs, processed in parallel differently from single thread, which caused the failure. 

The solution is consistent numbering of runs regardless of the way tmp files are calculated.

The ticket rises a question about pixel comparison validity, as mismatch, reported by the routine is confusing, though certainly caused by incorrect run_id.


_test/test_algorithms/test_duplicated_id.m
```diff
@@ -0,0 +1,100 @@
...
+                    id = 0;
+                else
+                    id = run.run_id;
+                end
+            end
+            ids = cellfun(@get_id,rf_proc,'UniformOutput',true);
+            assertEqual(ids,[1,0,2,6,0,3,7,8]);
+        end
+        
+        function test_do_nothing_inverted(~)
```
nickbattam-tessella, line 65: It's not clear from the names why this these do_nothing tests are expected to do nothing

_test/test_gen_sqw_workflow/gen_sqw_accumulate_sqw_tests_common.m
```diff
@@ -319,10 +319,14 @@ function test_gen_sqw(obj,varargin)
             obj.assertEqualToTolWithSave(w1a,'ignore_str',true,'tol',1.e-7);
         end
         %
-        function DISABLED_test_gen_sqw_sym(obj,varargin)
+        function test_gen_sqw_sym(obj,varargin)
             %-------------------------------------------------------------
+            if obj.save_output
+                return;
+            end
+            skipTest(fprintf('test_gen_sqw_sym_%s is disabled, Ticket #464',obj.test_pref));
```
nickbattam-tessella, line 327: you're skipping the test multiple times: here (always) and below (if bool)

nickbattam-tessella (changes requested):

_test/test_algorithms/test_duplicated_id.m
abuts: I wrote comment, explaining why this behavior is expected

_test/test_gen_sqw_workflow/gen_sqw_accumulate_sqw_tests_common.m
abuts: this test is currently failing and disabled, so the first sckip. I am starting to work on this ticket. In additional to that, it should be disabled on save as it does not currently save anything (and fails on skip in save mode)

The following skip is there for all tests and this will be there to inform if some tests become disabled on some platform due to some components missing (no mex, no parpool, etc....)

_test/test_algorithms/test_duplicated_id.m
nickbattam-tessella: you can use the test name to make that clear too, e.g. `test_does_nothing_if_all_data_unique`, `test_does_nothing_if_data_unique_and_inverse_order`

nickbattam-tessella (approved):

nickbattam-tessella (approved):

Issue #547, opened 2021-02-03
Use Warnings Next Gen plugin for CppCheck results on Jenkins


hsautessella:
The [Warnings-NG](https://github.com/jenkinsci/warnings-ng-plugin/blob/master/doc/Documentation.md) plugin is more modern and feature-rich than the `CppCheck` plugin. Warnings-ng can do things like: aggregate results with other warnings/static analyzers (e.g. GCC or mlint warnings), show the warnings in the context of code snippets and perform Git forensics (see docs). You can also set a "baseline" warning level, such that any builds that have over a given number of warnings will fail.

nickbattam-tessella:
Related to #474

Issue #474, opened 2020-12-03
Add Linting to MATLAB code


nickbattam-tessella:
The CppCheck is run during the C build process to perform static analysis on the C source and results reported in the Jenkins dashboard.

MATLAB provides the mlint tool which should be added to the Jenkins build pipeline to identify and report potential code issues and allow these issues to be monitored and tracked.

hsautessella:
We should use the [Warnings Next Gen](https://github.com/jenkinsci/warnings-ng-plugin) plugin to display the linter's results on Jenkins. Matlab's `checkcode` function (which is the recommended way of running `mlint`) generates a `struct` containing results; there's no option to output the report to an `xml`/`json` file that "Warnings Next Gen" can parse. Because there's no standard file format for `mlint` results, there's no standard parser that the Jenkins plugin can use.

A large part of this task is to convert the `struct` outputted by `checkcode` into the Warnings Next Gen native `xml` or `JSON` format - see https://github.com/jenkinsci/warnings-ng-plugin/blob/master/doc/Documentation.md#export-your-issues-into-a-supported-format

nickbattam-tessella:
Related to #547

PR #552, opened 2021-02-08
Add MLint warnings to build

oerc0122:
Run MLint on build (Horace #474)
Update warning to use Warnings Next Gen (Horace #547)


tools/build_config/build.sh
```diff
@@ -65,18 +65,10 @@ function run_tests() {
...
 
-    analysis_cmd="cppcheck --enable=all --inconclusive"
-    analysis_cmd+=" --xml --xml-version=2"
-    analysis_cmd+=" -I ${HORACE_ROOT}/_LowLevelCode/cpp"
-    analysis_cmd+=" ${HORACE_ROOT}/_LowLevelCode/"
-    analysis_cmd+=" 2> ${output_dir}/cppcheck.xml"
-    echo_and_run "${analysis_cmd}"
-  fi
+  echo_and_run "cd ${build_dir}"
+  echo_and_run "make analyse"
```
nickbattam-tessella, line 71: There is no Makefile in the source tree. Do we want to introduce make into the build? Can the same be achieved through CMake

tools/build_config/build.sh
```diff
@@ -173,7 +174,10 @@ function main() {
   fi
 
   if ((build)); then
-    run_build "${build_dir}"
+      if [ ! -e ${build_dir}/CMakeCache.txt ]; then
```
nickbattam-tessella, line 177: is there a way the boolean test here (`build && !-e CMakeCache.txt`) could be combined with the preceding `if ((configure))` test so the non-trivial `run_configure` call isn't duplicated?

oerc0122: Realistically, we could just replace `((configure))` with `if [ ! -e ${build_dir}/CMakeCache.txt ]; then`, I guess, because you can't do anything without the configure.

nickbattam-tessella: Removing the 'configure' flag would prevent the regeneration of configuration if anything were changed. To avoid that we'd need to add a 'clean' option to purge the build folder and CMake configuration.

nickbattam-tessella (approved):

PR #553, opened 2021-02-08
Bug in combine

abuts:
This is simple fix for the bug, being there for ages, and appearing if one have `nochange` mode and number of files is bigger then 8. Do not know how to write a unit test for it as our current tests are checking 6 files, 8 will be more expensive and in a way useless but separate test is difficult to derive.  Performance tests would be caching it.

### DIFF
--- a/horace_core/sqw/file_io/@combine_sqw_pix_job/private/read_pix_for_nbins_block_.m
+++ b/horace_core/sqw/file_io/@combine_sqw_pix_job/private/read_pix_for_nbins_block_.m
@@ -36,6 +36,16 @@ pix_tb=cell(nfiles,n_bin2_process);  % buffer for pixel information
 
 %
 bin_filled = false(n_bin2_process,1);
+if ischar(run_label)
+    if strcmpi(run_label,'nochange')
+        run_label = filenum; % will not be used, just to keep common 
+        % interface to split_pix_per_bin_
+    else
+        error('READ_PIXELS:invalid_argument',...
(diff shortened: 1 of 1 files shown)


horace_core/sqw/file_io/@combine_sqw_pix_job/private/read_pix_for_nbins_block_.m
```diff
@@ -36,6 +36,16 @@
 
 %
 bin_filled = false(n_bin2_process,1);
+if ischar(run_label)
```
nickbattam-tessella, line 39: Are there tests that run with/without `nochange` runlabel to confirm the change from passing the n-th letter of the word to the n-th filenum works?

abuts: I do not think so. I am actually not sure that there are separate tests for `write_nsqw_to_sqw` at all. Have to check

abuts: No separate tests -- only as part of gen_sqw

nickbattam-tessella (approved):
The change looks sensible and addresses the issue resulting from the char 'nochange' being interpreted as an array in later loop.

As there are no tests confirming that the char case is trapped, what consequence of the error being raised is, or that passing a filenumber (from the substituted array) rather than the character will give the desired outcome it's impossible to assess wider impact of the change in other edgecases.

PR #555, opened 2021-02-09
Update to make Herbert_ROOT match between Herbert and Horace

oerc0122:


abuts (approved):
looks good

hsautessella (approved):
Looks fine to me

cmake/FindHerbert.cmake
```diff
@@ -31,39 +34,55 @@ set(DIRS_TO_SEARCH
...
+endif()
+
+if (NOT Herbert_ROOT)
+  # Root is one dir above core
+  set(Herbert_ROOT "${Herbert_CORE}/.." CACHE PATH "Directory containing Herbert")
+else()
+  # If given through CLI
+  set(Herbert_ROOT "${Herbert_ROOT}" CACHE PATH "Directory containing Herbert")
 endif()
+mark_as_advanced(Herbert_ROOT)
```
nickbattam-tessella, line 64: why is this advanced?

nickbattam-tessella (commented):
The logic in the middle is hard to pick through. It's not clear what happens if `Herbert_CORE` and `Herbert_ROOT` are both defined on input and are inconsistent

cmake/FindHerbert.cmake
oerc0122: Because it should be set automatically, not generally by the user unless they know what they're doing.

cmake/FindHerbert.cmake
```diff
@@ -7,9 +7,12 @@ Looks for the Herbert Matlab software package.
 Variables defined by the module
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-``Herbert_ROOT``
+``Herbert_CORE``
```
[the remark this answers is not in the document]
oerc0122: They are independent. It will attempt to detect the `herbert_init.m` in `Herbert_CORE` and if that fails, it will error out. If it succeeds, it will add `Herbert_CORE` and `Herbert_ROOT` into the cache, `Herbert_ROOT` as the one defined by the user.

PR #562, opened 2021-02-12
Api discussion

oerc0122:


documentation/matlab-api.md
```diff
@@ -0,0 +1,381 @@
...
+```
+arrayfun(@()(), list, 'UniformOutput', true)
+runtests(pwd,'IncludeSubfolders',true)
+runtests(tests, 'LoggingLevel', matlab.unittest.Verbosity.VERBOSE)
+```
+
+### Python
+Python's main method of handling variable arguments is to use the standard `*args` and `**kwargs` arguments.
+
+```
```
hsautessella, line 29:
```suggestion
```python
```

For Python syntax highlighting

documentation/matlab-api.md
```diff
@@ -0,0 +1,381 @@
...
+def function_name(*args):
+    for arg in args:
+        print(arg)
+
+function_name(1,2,3,4)
+
+def function_name(**kwargs):
+    if kwargs.get('header', False):
+        print(header)
+        del kwargs['header']
```
hsautessella, line 39: A bit nit-picky since this is only documentation, but `header` is undefined. Did you mean something like:
```python
if kwargs.get('header', False):
    print(kwargs.pop('header'))
```

documentation/matlab-api.md
```diff
@@ -0,0 +1,381 @@
...
+```
+
+Any or all of these can be used in combination with each other.
+
+### MATLAB Argument Parsers
+
+### `inputParser`
+Built-in `inputParser` supports positional, optional and required or optional key-value arguments
+
+```
```
hsautessella, line 69:
```suggestion
```matlab
```
for Matlab syntax highlighting

documentation/matlab-api.md
```diff
@@ -0,0 +1,381 @@
...
+|:-----------:|:-------------:|:-----------------:|:---:|
+| Positional  |      Y        |        Y          |  Y  |
+| Required    |      Y        |        Y          |  Y  |
+| Optional    |      Y        |        Y          |  Y  |
+| Key-value   |      Y        |        Y          |  Y  |
+| Flag        |      N        |        Y          |  Y  |
+| Default val |      Y        |        Y          |  Y  |
+| Validate value  |  Y        |        N          |  Y  |
+| Validate type |    Y        |        N          |  Y  |
+| Abbreviation |     N        |        Y          |  Y  |
```
hsautessella, line 162: `inputParser` allows abbreviations (or [partial matching](https://www.mathworks.com/help/matlab/ref/inputparser.html#d122e664961)).

documentation/matlab-api.md
```diff
@@ -0,0 +1,381 @@
...
+else
+    error('Check number of input arguments')
+end
+
+(from disp2sqw)
+```
+
+#### Pros
+- Completely flexible handling of arguments specific to function
+
```
cmarooney-stfc, line 143: PRO: gives you no illusions that you don't have to comment it to understand it. It is quite clear that someone should lay out the input argument options, preferably at this point, and why this degree of complexity is required.

documentation/matlab-api.md
```diff
@@ -0,0 +1,381 @@
...
+#### Pros
+- Very flexible
+- Well documented with potential online help
+- Standard MATLAB method
+- Explicit argument status
+- Keyword validation
+
+#### Cons
+- May change in future
+- Not extensible if doesn't meet requirements
```
cmarooney-stfc, line 95: CON: there is a known issue that an optional character input followed by a keyword-value pair will not give what you think. This combination should be explicitly avoided.

documentation/matlab-api.md
```diff
@@ -0,0 +1,381 @@
...
+(from refine_crystal)
+```
+
+#### Pros
+- Very flexible
+- Well documented
+- Under PACE control
+
+#### Cons
+- Bespoke, requires maintenance
```
hsautessella, line 116: There seem to be some tests for `parse_char_options` and `parse_args_namelist`, but not for `parse_arguments`. It's a relatively complex function (~500-600 lines) that's not explicitly tested, which could lead to quite high maintenance costs.

Not sure if that's worth mentioning or not, but thought I'd flag it

documentation/matlab-api.md
```diff
@@ -0,0 +1,381 @@
...
+- Requires Python implementation similar to that as noted above
+- Replace `parse_arguments` flag parser temporarily with variant printing deprecation warning and allow both syntaxes to coexist for a period
+- Need to check for flags in public APIs, can't check user-scripts
+
+
+# Proposal
+
+- Use positional parameters for core data
+- Use optional parameters where appropriate
+- Use key-value arguments for remaining arguments, including  rather than flag/no-flag arguments
```
hsautessella, line 306: "including"?

cmarooney-stfc, line 306: I don't understand "including rather than" - looks as if a word is missing

documentation/matlab-api.md
```diff
@@ -0,0 +1,381 @@
...
+function_name(a, 'flag', 1)
+function_name(a, '-noflag')
+function_name(a, 'flag', 0)
+```
+- Python can just pass args directly into the MATLAB call in principle
+
+##### Cons
+
+- Raises questions about the separation between arguments and flags are additional arguments to be labelled as flags?
+- Not Pythonic way of doing things
```
cmarooney-stfc, line 239: I think we should be spelling out exactly what is "not Pythonic" in any given instance

documentation/matlab-api.md
```diff
@@ -0,0 +1,381 @@
...
+PYTHON: def function_name(arg1, arg2, *args)
+```
+Both would be called as
+```
+function_name(a, b, c, d, ...)
+```
+Where `a` maps to `arg1`, `b` to `arg2`, `c` and `d` to a cell array (`varargin`) in MATLAB and a tuple (`args`) in Python.
+
+### Keyword-Value Pairs
+Keyword value pairs are a common means of allowing a variable number of arguments without requiring that they be in any particular order. Typically they follow positional arguments.
```
cmarooney-stfc, line 342: In the call. In the function/def line it is different

hsautessella (commented):
Not sure if this is worth mentioning or not, but Matlab's `inputParser` also allows [struct expansion](https://www.mathworks.com/help/matlab/ref/inputparser.html#d122e665029) as well as keyword-value pairs.

documentation/matlab-api.md
oerc0122: Which call? What is different? Sorry, I don't understand.

documentation/matlab-api.md
oerc0122: I have tried to address the Pythonic way of doing things in the appendix

documentation/matlab-api.md
oerc0122: I'm not sure one should encourage code that requires comments to be understood as a pro. Comments should be for clarification, not comprehension.

documentation/matlab-api.md
oerc0122: Originally I had written
```python
if header := kwargs.get('header', False):
```
But thought that would add more confusion and removed it without replacing it

documentation/matlab-api.md
mducle: In Python I would prefer to use "keyword=True/False" instead of flags, instead of using a string which starts with a minus.

If we're writing (perhaps automatically generated) wrappers for each Matlab function, this could be handled without changing the Matlab code by parsing the Matlab arguments. If we're going to just use generic wrappers this would require changing the Matlab code.

For example, the `-nopix` flag of `cut_sqw` can be specified in Python using the `pix` keyword which is `True` by default. E.g. `cut_sqw(..., '-nopix')` in Matlab would be `cut_sqw(..., pix=False)` in Python.

documentation/matlab-api.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Options
+
+### Dash flags `-(no)flagname`
+
+Allow flags with dashes in MATLAB
+
+#### MATLAB
+
+Currently, using argument parser is flexible in the way it handles flags and allows the same flag (or its negation) to be specified in multiple ways.
```
mducle, line 175: Is "argument parser" the Horace `parse_argument` function or the Matlab builtin `inputParser` class?

documentation/matlab-api.md
```diff
@@ -0,0 +1,385 @@
...
+#### Python
+
+Python keyword arguments are commonly passed through as:
+
+```python
+function_name(a, b, c, flagname = true)
+function_name(a, b, c, flagname = false)
+```
+
+If `call_MATLAB` is implemented along the lines of:
```
mducle, line 275: This is exactly how it is currently implemented: https://github.com/pace-neutrons/pace-python/blob/main/pace_python/MatlabProxyObject.py#L104

mducle (commented):
Thanks for writing this doc. I agree with the plan to use required and optional positional arguments and optional keyword arguments and to ditch the flags. 

But maybe you could add a note to say that Python is more flexible than Matlab with respects to positional and keyword arguments in that Python accepts keyword arguments *in the place* of positional arguments, whereas Matlab does not.

This is because Python functions accept two distinct data types as argument containers (`list` / cell arrays and `dict` / structures), whereas Matlab functions only accepts arguments in the form of a `list` / cell array (which *must* be called `varargin` whereas `args` and `kwargs` are named so in Python only by convention). 

For example, the function signature:

```
cut_sqw(data, proj, p1_bin, p2_bin, p3_bin, p4_bin, varargin)
```

could be called:

```
cut_sqw('file.sqw', myproj, p1_bin=[1,2,3], p2_bin=[], p3_bin=[4,5], p4_bin=[6,7,8])
```

in Python, but must be called:

```
cut_sqw('file.sqw', myproj, [1,2,3], [], [4,5], [6,7,8])
```

(unless we explicitly rewrite the function to accept:

```
cut_sqw('file.sqw', myproj, 'p1_bin', [1,2,3], 'p2_bin', [], 'p3_bin', [4,5], 'p4_bin', [6,7,8])
```

but I think such rewriting is too much effort and would be happy to have the difference in keyword arguments behaviour between Matlab and Python but this would mean the wrappers need to also be aware of this.

oerc0122:
- Add ToC
- Add section on `parse_char_options`
- Explore possibility of `parse_char_options` and `inputParser` combined
-  Discuss CLI style working

oerc0122:
ADR following meeting discussing this document has now opened:
https://github.com/pace-neutrons/Horace/pull/583

PR #565, opened 2021-02-12
Create 0021-errors-and-warnings ADR

oerc0122:
Create ADR for error handling

### DIFF
--- a/documentation/adr/0021-errors-and-warnings.md
+++ b/documentation/adr/0021-errors-and-warnings.md
@@ -110,6 +110,6 @@ end
 - Never perform a global catch. Specific catches should be used where relevant, preferably using `switch` for clarity.
 - Use a standard form for PACE thrown errors for easy identification, perhaps:
   ```matlab
-  error('PACE:CLASS:error', errorMessage)
+  error('(HORACE|HERBERT):(<CLASSNAME>|<FUNCTIONNAME>):error', errorMessage)
   ```
 - On warning messages, not only should the problem be reported, but where possible, the solution.
--- a/documentation/adr/0021-errors-and-warnings.md
+++ b/documentation/adr/0021-errors-and-warnings.md
@@ -38,6 +38,14 @@ Presently in PACE, there are several different mechanisms being used which obscu
 
   (@mfclass_tobyfit/set_bfun)
   ```
+- Use of an extra function to handle errors
+  ```matlab
+  if ~exist(file1,'file')
+     [err,message]=report_err(['first file to merge: ',file1,' do not exist'],throw);
+     return
+  end
+  (admin/merge_files.m)
+  ```
 
 ### Issue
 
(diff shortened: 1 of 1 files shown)


documentation/adr/0021-errors-and-warnings.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+## Recommendations
+
+- Replace all handling via args (`ok`, `mess`  syntax) with try-throw-catch process
+- Use `rethrow` over `throw(ME.message)` which
+  1. Destroys the identifier
+  2. Obscures the backtrace
+- Never perform a global catch. Specific catches should be used where relevant, preferably using `switch` for clarity.
+- Use a standard form for PACE thrown errors for easy identification, perhaps:
```
nickbattam-tessella, line 100: A related issue that a lot of (mainly legacy) error messages aren't including an `errId` at all at the moment.

We've standardised on `CLASS-NAME:error-type` for errors raised. There's no reason not to extend this to `PACE:CLASS-NAME:error-type`, which is consistent with the `errId` here: https://www.mathworks.com/help/matlab/ref/mexception.html

hsautessella: I thought at the moment we were doing `CLASS:function_name` at the moment. But +1 for `PACE:CLASSNAME:error-type`, although I might suggest we use `HERBERT`/`HORACE` rather than `PACE`.

Also, there's a (non-default) MLint warning for using an error ID 
image

documentation/adr/0021-errors-and-warnings.md
```diff
@@ -0,0 +1,104 @@
...
+end
+```
+
+## Recommendations
+
+- Replace all handling via args (`ok`, `mess`  syntax) with try-throw-catch process
+- Use `rethrow` over `throw(ME.message)` which
+  1. Destroys the identifier
+  2. Obscures the backtrace
+- Never perform a global catch. Specific catches should be used where relevant, preferably using `switch` for clarity.
```
hsautessella, line 99: There's another (non-default) MLint setting we could enable for this 
image

oerc0122: The issue here is that I don't think MATLAB's catch will check for type (as in python)
```python
try:
   raise NameError()
except NameError:
except TypeError:
...
```
I suspect the linter will know whether the subsequent catch checks the identifier. I suspect it's only as smart as "Oh, `catch ME`", that's fine, then. 

I spent some time trying to test this, but the code analyser just shows a blank web browser for whatever reason.

hsautessella: I think you're right, but at least we'll get warnings for a bare `try-catch`. If there is a situation where we capture the error, but don't do anything with it, hopefully the "unused variable" warnings can alert us that something's wrong.

oerc0122: True, but there are cases of:
```
try
   ...
catch ME
   warning(ME.message)
end
```
I suppose regardless we take all the help we can get.

nickbattam-tessella (approved):

PR #570, opened 2021-02-15
Make Horace linting ignore Herbert root

oerc0122:
In conjunction with https://github.com/pace-neutrons/Herbert/pull/300 exclude Herbert

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,10 @@ include(PACE_AddMex)
 include(PACE_FindMatlab)
 include(horace_FindHDF5)
 include(PACE_Version)
+
+# Make analysis ignore Herbert in case of nesting
+set(PACE_MLINT_IGNORE ${PACE_MLINT_IGNORE} "${Herbert_ROOT}/**/*.m")
+
 include(PACE_CodeAnalysis)
 if(${BUILD_TESTS})
     include(CTest)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -36,6 +36,10 @@ include(PACE_AddMex)
 include(PACE_FindMatlab)
 include(horace_FindHDF5)
 include(PACE_Version)
+
+# Make analysis ignore Herbert in case of nesting
```
nickbattam-tessella, line 40: why is this set here rather than with the other sets at the top? I assume this must be set before `PACE_CodeAnalysis` is imported?

oerc0122: It needs to be below FindHerbert and above CodeAnalysis.

nickbattam-tessella: would be worth adding that in the comment to stop someone like me reordering for tidyness

nickbattam-tessella:
If you add the `Herbert_567_lint_ignore_herbert` tag to this PR it will build against that branch rather than `master`.

hsautessella:
Looks like Herbert warnings are still being picked up (including [a new one](https://anvil.softeng-support.ac.uk/jenkins/job/PACE-neutrons/job/Horace/job/PR-Scientific-Linux-7-2018b/880/mlint/new/source.d5ffa106-9cd3-4bf7-b6e0-dc5b51cc06c4/#47) in Herbert's `lint_json`).

nickbattam-tessella (approved):

cmarooney-stfc (commented):
Looks OK to me.

PR #582, opened 2021-02-17
Add skip CI based on labels

oerc0122:
Add the ability to skip CI based on the `do-not-build` and `do-not-test` PR labels

### DIFF
--- a/tools/build_config/Jenkinsfile
+++ b/tools/build_config/Jenkinsfile
@@ -235,6 +235,8 @@ if (env.BRANCH_NAME) {
   currentBuild.description = "Git-SHA: ${env.PR_COMMIT_SHA.take(7)}"
 }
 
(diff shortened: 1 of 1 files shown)


tools/build_config/Jenkinsfile
```diff
@@ -487,6 +508,11 @@ pipeline {
         ]
       )
 
+      archiveArtifacts(
```
nickbattam-tessella, line 511: now both this and the `cppcheck.xml/mlint.json` archiveArtifacts calls are `allowEmptyArchive` do they need to be separate, or is there clarity in keeping them the analysis and test outputs separate?

oerc0122: Windows doesn't run analysis, needs those to be able to be empty.
`do-not-build/test` need to allow `testing.xml` to be empty. 

Separation is because they're calculated on different stages and are different things.

cmarooney-stfc (approved):

Issue #577, opened 2021-02-16
Identify functions for which nargin determines the meaning of positional arguments


nickbattam-tessella:
Identify all functions where the length of the input argument list (`nargin`) determines the meaning and expected type of the positional arguments.

All functions should have:
- required arguments (fixed order and type)
- optional arguments (fixed order and type, must be specified if previous arg included)
- keyword arguments (any order)

Output:
- record of functions with this behaviour
- new issue tickets created for each function including the proposed solution (new keyword args, new functions, ...)

Issue #576, opened 2021-02-16
Convert all MATLAB flag arguments to use `-` prefix


nickbattam-tessella:
For consistency across the MATLAB API and enable integration with Python, all flag arguments to functions must be updated to use the currently optional `-` prefix.

The prefix is specified through the `parse_arguments` `prefix` (default: '-') and `prefix_req` (default: `true`) arguments

Output
- documentation updated to remove references to `-` flags
- references to `-` name flags updated

Issue #556, opened 2021-02-10
Document current and proposed MATLAB / Python API


nickbattam-tessella:
A number of different API syntaxes are used for optional and flag arguments in MATLAB.

These should be reviewed and documented as part of a discussion of how these map to the Python API

cmarooney-stfc:
@nickbattam-tessella I have now realised that ticket #575 actually requests the ADR and links to PR 583. I'm presuming that this is the same issue - please correct if not, but for completeness it would be good if this could - if not deleted - be linked through to the other items. Thanks

oerc0122:
This issue was regarding developing the docs required the meeting and discussion, not the ADR that came out of it. Have "done"d and added the appropriate linkback from the (now closed) PR

PR #583, opened 2021-02-17
ADR for Python/MATLAB API conventions

nickbattam-tessella:
Proposal arising from discussion of creation of a consistent API syntax for MATLAB and Python

Issues #576 and #577 created to implement changes to Horace and Herbert required if this proposal is approved.

Addresses #556


documentation/adr/0021-use-keyword-args.md
```diff
@@ -0,0 +1,73 @@
...
+Currently, we don’t have an established syntax for Python. Because we have a clean slate it makes sense to only have a Pythonic syntax using required positional, optional positional, and optional keyword/value pairs
+
+```python
+def function_name(a, b, *args, **kwargs):
+	# body
+```
+
+ e.g. `function_name(a, b, c, d, keyword_1=val_1, keyword_2=val_2, …)`
+
+Python does not natively support flag arguments, and they are not used in the core language. They are supported in a small number of Python libraries, notably `matplotlib`, which was modelled on the MATLAB API, and may be implemented on any  function as optional string arguments with custom handling.
```
oerc0122, line 25: Worth noting `matplotlib` has removed many of these?

documentation/adr/0021-use-keyword-args.md
```diff
@@ -0,0 +1,73 @@
...
+The key-value pair syntax is equivalent to the Pythonic syntax, e.g. `function_name(a, b, c, d, ‘keywordA’, valA, ‘keywordB’, true, …)`.
+
+The mix of `-flagname` and `flagname` introduces complexity into argument parsing - specifically identifying if a string argument `flagname` a flag or a keyword argument?
+
+
+## Decision
+
+The MATLAB APIs will support positional, optional flag and keyword arguments:
+
+- flag arguments will be prefixed with `-`
```
oerc0122, line 47: Replace "will" with "must"?

documentation/adr/0021-use-keyword-args.md
```diff
@@ -0,0 +1,73 @@
...
+- if the same parameter is passed as both a flag and keyword argument an error will be raised if their values are inconsistent
+- flag arguments may be negated, specified as `-noflagname` (`false`) or `-flagname` (`true`). An error will be raised if this prefix would lead to an ambiguity with another flag argument e.g. defining`-nodes` (use nodes) and `-des`
```
oerc0122, line 49: Prefer only one option? Otherwise you're guaranteed to have a null flag along with more complicated parsing. 
i.e. Define `-nopix` exclusively, no ambiguity, no added complexity, no NOP flag?
Of course this does lead to also having the key-val argument `'nopix', true/false`

documentation/adr/0021-use-keyword-args.md
```diff
@@ -0,0 +1,73 @@
...
+
+The mix of `-flagname` and `flagname` introduces complexity into argument parsing - specifically identifying if a string argument `flagname` a flag or a keyword argument?
+
+
+## Decision
+
+The MATLAB APIs will support positional, optional flag and keyword arguments:
+
+- flag arguments will be prefixed with `-`
+- if the same parameter is passed as both a flag and keyword argument an error will be raised if their values are inconsistent
```
oerc0122, line 48: Define behaviour otherwise (i.e. OK)
e.g.
`function(a, b, '-flag', 'flag', true)`

documentation/adr/0021-use-keyword-args.md
```diff
@@ -0,0 +1,73 @@
...
+- flag arguments may be negated, specified as `-noflagname` (`false`) or `-flagname` (`true`). An error will be raised if this prefix would lead to an ambiguity with another flag argument e.g. defining`-nodes` (use nodes) and `-des`
+- flag arguments may be abbreviated
```
oerc0122, line 50: And non-flag arguments?

documentation/adr/0021-use-keyword-args.md
```diff
@@ -0,0 +1,73 @@
...
+
+The Python API will use positional, optional and keyword arguments.
+
+All optional arguments must appear between the positional and keyword arguments.
```
oerc0122, line 54: Suggest defining explicit order:
mandatory (positional) arguments, optional (positional) arguments, keyword arguments, output/modified arguments(?)

oerc0122 (changes requested):

documentation/adr/0021-use-keyword-args.md
oerc0122: By which I mean do we want to disallow negation for clarity and to avoid some flags doing nothing, but convincing users that they have some purpose? As mentioned, without special handling, this may require negation of the relevant keyword arg by default which may lead to more confusion.

mducle: I would agree with this, meaning that the flag value is `false` by default but is switched to true if the user specifies it. So in the `cut` case the flag name would actually be "nopix" rather than "pix" - e.g. users can specify `cut(..., '-nopix')` as they do now, but `cut(..., '-pix')` which currently does nothing would in future yield an error (flag not defined)... 

A quick look shows that the other negative flags in use are `-noplot` and `-nodisplay` but most other flags adhere to the rule that they are default `false` and switched to `true` if specified... except for `dispersion_plot`.

documentation/adr/0021-use-keyword-args.md
```diff
@@ -0,0 +1,73 @@
...
+- optional parameters interspersed with positional parameters in some functions
+
+The key-value pair syntax is equivalent to the Pythonic syntax, e.g. `function_name(a, b, c, d, ‘keywordA’, valA, ‘keywordB’, true, …)`.
+
+The mix of `-flagname` and `flagname` introduces complexity into argument parsing - specifically identifying if a string argument `flagname` a flag or a keyword argument?
+
+
+## Decision
+
+The MATLAB APIs will support positional, optional flag and keyword arguments:
```
mducle, line 45:
```suggestion
The MATLAB APIs will support positional, optional, flag and keyword arguments:
```

mducle (approved):
Looks good.

documentation/adr/0021-use-keyword-args.md
nickbattam-tessella: The mandatory order of `positional`, `optional`, `keyword` is implicit but can be stated explicitly. 

Any arguments to the right of keyword arguments must be keyword arguments in Python.

Modified output arguments aren't really part of MATLAB (with the exception of handle-classes that may be updated) or Python (everything passed by value), however both languages support multiple output arguments.

documentation/adr/0021-use-keyword-args.md
cmarooney-stfc: (1) Flag arguments may be abbreviated by truncation at the rear to at most the minimum unambiguous string. Other abbreviations will not be supported.
(2) Keyword arguments may not be abbreviated as they would otherwise interfere with the keyword matching in python, which is not an argument-parsing process. (Discuss...)

documentation/adr/0021-use-keyword-args.md
```diff
@@ -0,0 +1,73 @@
...
+- flag arguments may be negated, specified as `-noflagname` (`false`) or `-flagname` (`true`). An error will be raised if this prefix would lead to an ambiguity with another flag argument e.g. defining`-nodes` (use nodes) and `-des`
+- flag arguments may be abbreviated
+
+The Python API will use positional, optional and keyword arguments.
+
+All optional arguments must appear between the positional and keyword arguments. 
+
+## Consequences
+
+- The Python-MATLAB wrapper can be simply implemented, passing the Python `*args` and `**kwargs`directly to MATLAB as an array
```
cmarooney-stfc, line 58: An array? or arrays?

documentation/adr/0021-use-keyword-args.md
```diff
@@ -0,0 +1,73 @@
...
+- The simple wrapper allows flags to be passed from Python, although this is not a Pythonic syntax. This ability not be "officially" supported.
+- The following calls are all equivalent:
+```python
+wrapped_function_name(a, b, flagname=true)
+wrapped_function_name(a, b, 'flagname', 'true')
+wrapped_function_name(a, b, '-flagname')
+```
```
cmarooney-stfc, line 65: this explicitly debars flagname from becoming a (Python) positional argument in the future. Is there a way of making in clear in the function definition?

documentation/adr/0021-use-keyword-args.md
```diff
@@ -0,0 +1,73 @@
...
+```matlab
+function_name(a, b, '-flagname');
+function_name(a, b, 'flagname', 'true');
+```
+
+- MATLAB argument parsing will be solely responsible for process of arguments
```
cmarooney-stfc, line 71: process -> the processing of

documentation/adr/0021-use-keyword-args.md
```diff
@@ -0,0 +1,73 @@
...
+- MATLAB argument parsing will need to manage flags passed in both 'flag' and 'keyword' form
+- All MATLAB functions which currently support optional arguments out of order must be updated, either creating a new function name with the alternate signature (e.g. `cut` with no projection) or additional keyword arguments (e.g. `lattice` in `dispersion_plot`)
```
cmarooney-stfc, line 73: I think I understand the point here from our discussion but it will be quite cryptic to the outside reader. I think you want to say "where alternative function signatures have the optional arguments in different orders, with parsing done by inspecting numbers of arguments, this must be updated so that all function overloads have the same order of optional arguments, and where this is not possible, a function with a new name must be created."

cmarooney-stfc (changes requested):

documentation/adr/0021-use-keyword-args.md
oerc0122: I imagine it would actually be expanded as an argument stream via unpacking.

documentation/adr/0021-use-keyword-args.md
oerc0122: I'm not sure what you mean by positional? Both `'flagname', 'true'` and `'-flagname'` are positional form. If you mean a named positional, depending on the definition of `wrapped_function_name`, it doesn't. Given, however, that the "python signature for each MATLAB function" version is the "dream" situation, that is unlikely in the long run. Most likely the wrapper will be (and I believe is) of the form:
```python
call_matlab(function_name, *args, **kwargs)
```
Making this moot.

cmarooney-stfc:
@nickbattam-tessella @oerc0122 Nick, Jacob - pointing out that this PR creates ADR-21 for matlab/python API but PR 565 also creates ADR-21 for errors and warnings protocol. As Jacob was first, suggest we make this one ADR-22. thanks

cmarooney-stfc (changes requested):
Blocking this one as it will create a second ADR-21, first one was PR 565

documentation/adr/0021-use-keyword-args.md
nickbattam-tessella: if they are consistent there is nothing to define: the flag will take the value assigned via both syntaxes

oerc0122: Exactly. Do you think it's worth mentioning?

oerc0122 (approved):

documentation/adr/0022-use-keyword-args.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Consequences
+
+- The Python-MATLAB wrapper can be simply implemented, passing the Python `*args` and `**kwargs`directly to MATLAB as an array by unpacking the arguments
+- The simple wrapper allows flags to be passed from Python, although this is not a Pythonic syntax. This ability will not be "officially" supported
+- The following calls are all equivalent:
+```python
+wrapped_function_name(a, b, flagname=true)
+wrapped_function_name(a, b, 'flagname', 'true')
+wrapped_function_name(a, b, '-flagname')
```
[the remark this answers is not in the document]
hsautessella:
```python
wrapped_function_name(a, b, 'flagname', 'true')
```
makes me uneasy. I'd expect that `args = ("flagname", "true")` and `kwargs = {}` whereas, what you're effectively getting is `args = ()` and `kwargs = {"flagname": "true"}`.

hsautessella (commented):
Mostly seems sensible, but I have a few questions about supporting flags on the Python side.

documentation/adr/0022-use-keyword-args.md
```diff
@@ -0,0 +1,92 @@
...
+
+The Python API will use ordered (mandatory) positional, optional positional and keyword arguments in the function definition.
+
+All optional arguments must appear between the positional and keyword arguments.
+
+
+## Consequences
+
+- The Python-MATLAB wrapper can be simply implemented, passing the Python `*args` and `**kwargs`directly to MATLAB as an array by unpacking the arguments
+- The simple wrapper allows flags to be passed from Python, although this is not a Pythonic syntax. This ability will not be "officially" supported
```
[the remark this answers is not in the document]
oerc0122: No, it would be a side effect of using the MATLAB to parse the arguments regardless.

documentation/adr/0022-use-keyword-args.md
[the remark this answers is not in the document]
oerc0122: Again, these are side effects of using the MATLAB. But yes, it shouldn't be a string there.

documentation/adr/0022-use-keyword-args.md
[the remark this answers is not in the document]
nickbattam-tessella: I can't see any simple way to dependably prevent flags being passed in this form from Python; we would need to remove any string parameter starting with an `-`, but that could accidentally remove string valued arguments

documentation/adr/0022-use-keyword-args.md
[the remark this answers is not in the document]
nickbattam-tessella: Whether that data is args or kwargs is invisible and managed by the Python wrapper: by the time they are pushed to MATLAB they are just a list of `arg1, arg2, arg3, key, value, ...`

rebeccafair (approved):

documentation/adr/0022-use-keyword-args.md
```diff
@@ -0,0 +1,92 @@
...
+- flag arguments will be prefixed with `-`
+- flag arguments will default to `false` if not passed
+- if the same parameter is passed as both a flag and keyword argument with inconsistent values an error will be raised, otherwise the parameter is set to the passed value
+- support for negated flags, prefixed with `no`, will be removed. All flags will map to `true` when specified. Flag names may include the `no` prefix to make their meaning clear, e.g. `-nopix` will suppress pixel handling in that function
+- flag arguments may be truncated to the minimum unambiguous string for all flags defined on that function e.g. `-flagname`, `-flag`, `-f` are all valid. No other abbreviations will be supported
+- keyword arguments may not be abbreviated.
+
+The Python API will use ordered (mandatory) positional, optional positional and keyword arguments in the function definition.
+
+All optional arguments must appear between the positional and keyword arguments.
```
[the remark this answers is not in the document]
nickbattam-tessella: This is a feature of the `inputParser` rather than the native named arg vs varargin syntax in MATLAB.

I disagree that it means that optional positional arguments aren't required to appear before keyword arguments, it just means that optional positional arguments may also be passed using keyword syntax.

hsautessella (approved):
Happy for this to go in. I can't think of a clear way to disambiguate the flag arguments in Python, maybe one will present itself when we implement it.

PR #584, opened 2021-02-18
Clarify which pipelines vars affect

oerc0122:


tools/build_config/Jenkinsfile
```diff
@@ -135,25 +135,25 @@ properties([
     ),
     string(
       defaultValue: '3.19',
-      description: 'The version of CMake to run the build with.',
+      description: 'The version of CMake to run the build with. Not relevant on current Windows pipelines.',
```
nickbattam-tessella, line 138: Better to say where it is used than where it is not used.

tools/build_config/Jenkinsfile
```diff
@@ -135,25 +135,25 @@ properties([
...
       defaultValue: '6.3.0',
-      description: 'The version of GCC to build with.',
+      description: 'The version of GCC to build with. Not relevant on current Windows pipelines.',
       name: 'GCC_VERSION',
       trim: true
     ),
     string(
       defaultValue: '2017',
-      description: 'The year of the version of Visual Studio to build with.',
+      description: 'The year of the version of Visual Studio to build with.  Not relevant on current pipelines.',
```
nickbattam-tessella, line 150: this is used in build.ps1 so is used in Windows pipelines

nickbattam-tessella (changes requested):

tools/build_config/Jenkinsfile
oerc0122: I preferred to say where it wasn't in case new pipelines were added. Can update.

cmarooney-stfc: Just adding my agreement with Nick. The "Not relevant" looks odd. Maybe it would be better to centralise scopes for these comments in another document rather than piecemeal through the descriptions,

oerc0122: Ok, will fix. This is what shows up on the Jenkins "Build with Parameters", so it's documenting it there for user awareness, as well as within the Jenkins script.

tools/build_config/Jenkinsfile
nickbattam-tessella: If it's genuinely not required it should probably be deleted!

nickbattam-tessella (approved):

PR #587, opened 2021-02-18
Set defaults in Jenkinsfile to allow easy addition of new params

oerc0122:
- Set defaults via `get_param` function
- Move printing of `PR_LABELS` for general use and consistency


nickbattam-tessella:
The "consequence" of this change is that the versions of things used by a specific build aren't visible in the Jenkins UI.

Is everything logged? 

The builds scripts include a "write versions" step logging `cmake`, `matlab`, `g++` and cppcheck (Linux) / `cmake`, `matlab` and `vs_version` (Windows). 
Are the other variables all logged from the Jenkinsfile output?

oerc0122:
@nickbattam-tessella  I could add logging into the `get_param` function and report the final values, which would handle that if preferred.

nickbattam-tessella:
That's a good idea; the `get_param` function is only called setting up the build environment and logging in that would guarantee the build log included full information about the build parameters used in the the Jenkins file. These aren't necessarily the same as those passed to and reported by the build scripts.

tools/build_config/Jenkinsfile
```diff
@@ -239,14 +241,30 @@ Boolean run_build_stage = (! env?.PR_LABELS =~ "do-not-build")
 Boolean run_test_stage = (! env?.PR_LABELS =~ "(do-not-build|do-not-test)")
 
 pipeline {
+
   agent {
     label env.AGENT
   }
 
+  environment {
+    MATLAB_VERSION = get_param('MATLAB_VERSION', get_matlab_release(env.JOB_BASE_NAME))
```
nickbattam-tessella, line 250: a comment here or in the `get_param` function definition would make the behaviour clear: if the Jenkins parameter is passed that value is used, otherwise the default is applied

nickbattam-tessella (approved):

cmarooney-stfc (commented):
Is there a ticket for this PR. Apologies if it's there but I can't see it. Nice if we had an issue to link it to. thanks

oerc0122:
@cmarooney-stfc Sorry, forgot to tag it.

nickbattam-tessella (approved):

PR #588, opened 2021-02-19
Fix failing docs due to paths and permissions

oerc0122:
Update to fix the failing docs
As part of this a new configure stage is defined per comments from @nickbattam-tessella 
A corresponding update has been performed in [Herbert #311](https://github.com/pace-neutrons/Herbert/pull/311) to keep both Jenkins pipelines in line


cmake/PACE_Docs.cmake
```diff
@@ -68,6 +90,10 @@ else()
     COMMENT "Docs require sphinx and sphinx-rtd-theme to build"
     )
 
+  add_custom_target(docs-pack
+    COMMENT "Docs require sphinx and sphinx-rtd-theme to build"
+    )
```
cmarooney-stfc, line 95: Looks like a duplicate of the next block? As it isn't - docs-pack rather than manual, may be worth commenting what the difference is

oerc0122: This is the error message which is reported if you don't have the required packages and try to build. Because docs-pack is explicitly just "build the HTML docs and pack them (in tar or zip), I would say the error message is more helpful like this. Manual, however, is building a latex document

cmarooney-stfc: I was more thinking that when a maintainer comes along sometime in the future and has to distinguish between these things, some cue to help them along could be good. You could do it as a Groovy comment outside the add-custom-target block, or you could just extent the COMMENTs slightly with "Latex docs require .... " and "HTML docs require..."

oerc0122: I can't do it as a Groovy comment, as this is CMake, but I have extended the comments to clarify what is being built.

cmake/PACE_Docs.cmake
```diff
@@ -10,34 +10,56 @@ set(MANUAL_OUTPUT_DIR "${DOCS_WORK_DIR}/latex" CACHE FILEPATH "Directory to put
...
+      DEPENDS docs
+      )
+
   else()
     add_custom_command(TARGET docs POST_BUILD
       COMMAND sed -i -r "/\[NULL\]/d" "${DOCS_OUTPUT_DIR}/*html"
       DEPENDS build-docs
       )
+
+    set(DOCS_PACK_OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/docs.tar.gz" CACHE FILEPATH "Directory containing built HTML documentation")
```
hsautessella, line 49: Can we prefix cache variables with `Horace_`?

cmake/PACE_Docs.cmake
```diff
@@ -10,34 +10,56 @@ set(MANUAL_OUTPUT_DIR "${DOCS_WORK_DIR}/latex" CACHE FILEPATH "Directory to put
...
-                 "Foreach($f in Get-ChildItem -Path '${DOCS_OUTPUT_DIR}' -Filter *.html) {
-                      (Get-Content $f.FullName) | Where-Object {$_ -notmatch '\\[NULL\\]'} | Set-Content $f.FullName
+                 "Foreach($f in Get-ChildItem -Path '${DOCS_OUTPUT_DIR}' -Filter *.html) { \
+                      (Get-Content $f.FullName) | Where-Object {$_ -notmatch '\\[NULL\\]'} | Set-Content $f.FullName \
                   }"
       DEPENDS build-docs
       VERBATIM
       )
+
+    set(DOCS_PACK_OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/docs.zip" CACHE FILEPATH "Directory containing built HTML documentation")
```
hsautessella, line 34: Ditto

tools/build_config/Jenkinsfile
```diff
@@ -355,15 +372,18 @@ pipeline {
       }
       steps {
         script {
+
           if (isUnix()) {
             sh '''
+              export PATH=${PATH}:~/.local/bin &&
```
hsautessella, line 378: What's the path addition here for? Do we need them for the pip installed packages? A comment may be useful

tools/build_config/Jenkinsfile
```diff
@@ -329,15 +329,32 @@ pipeline {
       steps {
         script {
           if (isUnix()) {
+            if (env.BUILD_DOCS?.toBoolean()) {
```
nickbattam-tessella, line 332: why does analyse need to be aware of whether the docs are being built? We're not performing any static analysis on the documents

tools/build_config/Jenkinsfile
```diff
@@ -355,15 +372,18 @@ pipeline {
       }
       steps {
         script {
+
           if (isUnix()) {
             sh '''
+              export PATH=${PATH}:~/.local/bin &&
+              module load python/3.6 &&
```
nickbattam-tessella, line 379: why is Python required for the build step? Isn't this is only required in the `Build-Docs` stage

tools/build_config/Jenkinsfile
nickbattam-tessella: if we need to explicitly include this path in multiple build steps can you add it as a constant to help with maintainability.

There is also an inconsistency in the ordering of the `PATH` updates and the `module load`s : before in 'build' and 'analyse' steps, after in the 'build-docs' step

cmake/PACE_Docs.cmake
```diff
@@ -10,34 +10,56 @@ set(MANUAL_OUTPUT_DIR "${DOCS_WORK_DIR}/latex" CACHE FILEPATH "Directory to put
 find_program(sphinx-build NAMES sphinx-build)
```
nickbattam-tessella, line 10: There's no file-level docstring documenting which (if any) variables the PACE_Docs function requires to be set and which is sets and which build targets it adds (`docs`, `manual`, `docs-pack`)

tools/build_config/Jenkinsfile
oerc0122: This was more about not wanting to have to define a new configure stage. Because configure tries to find `sphinx-build` here, if it is not installed it will set that to be not found. This is a problem when it comes to build the docs.

tools/build_config/Jenkinsfile
oerc0122: Again, because of the configure happening here on Windows. Also for consistency in case the analyse stage is ever skipped (which it is for Windows).

tools/build_config/Jenkinsfile
nickbattam-tessella: the complexity and unnecessary configuration this has introduced suggests extracting a separate configure stage with knowledge of whether the doc stage is required would be clearer

cmake/PACE_Docs.cmake
```diff
@@ -1,55 +1,120 @@
...
+
+``Horace_DOCS_PACK_OUTPUT``
+Output filename (including extension) for compressed docs file
+e.g. ``${CMAKE_CURRENT_BINARY_DIR}/docs.zip``
+
+``Horace_MANUAL_WORK_DIR``
+Working directory where LaTeX docs are to be built
+
+``Horace_MANUAL_OUTPUT_DIR``
+Output directory where LaTeX docs are placed
```
nickbattam-tessella, line 30: is this an output or working directory? or is this where the built `.pdf` or `.dvi` is placed?

cmake/PACE_Docs.cmake
```diff
@@ -1,55 +1,120 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``docs``
+Build a HTML webpage variant of the user documentation in ``Horace_DOCS_OUTPUT_DIR``
+
+``docs-pack``
+Build HTML docs and compress
+
+``manual``
+Build a LaTeX manual variant of the user documentation
```
nickbattam-tessella, line 42: as a pdf or dvi or uncompiled sources?

tools/build_config/Jenkinsfile
```diff
@@ -342,6 +342,47 @@ pipeline {
...
+            } else {
+              powershell '''
+             pip install --user sphinx
+             pip install --user sphinx_rtd_theme
+            '''
+            }
+          }
+          if (isUnix()) {
+            sh '''
+          export PATH=${PATH}:~/.local/bin &&
```
nickbattam-tessella, line 364: does Python need to be on the path here so `--print_versions` can log the Python version along with the other versions?

nickbattam-tessella (changes requested):

cmake/PACE_Docs.cmake
```diff
@@ -1,55 +1,120 @@
...
+
+#]=======================================================================]
+
+set(Horace_DOCS_ROOT_DIR "${Horace_ROOT}/documentation/user_docs")
+set(Horace_DOCS_SOURCE_DIR "${Horace_DOCS_ROOT_DIR}/docs")
+set(Horace_DOCS_WORK_DIR "${Horace_DOCS_ROOT_DIR}/build" CACHE FILEPATH "Directory to put in-progress docs")
+set(Horace_DOCS_OUTPUT_DIR "${Horace_DOCS_WORK_DIR}/html" CACHE FILEPATH "Directory containing built HTML documentation")
+set(Horace_MANUAL_WORK_DIR "${Horace_DOCS_WORK_DIR}/latex" CACHE FILEPATH "Directory to build LaTeX sources")
+set(Horace_MANUAL_OUTPUT_DIR "${Horace_DOCS_WORK_DIR}/latex" CACHE FILEPATH "Directory to put compiled LaTeX manual")
```
hsautessella, line 51: In the interest of having fewer options crowding the CMake GUI, do all these need to be cache variables? Would just having a `Horace_DOCS_BUILD_DIR` cache variable be sufficient? The HTML would always be built in `Horace_DOCS_BUILD_DIR/html` and the LaTeX always built in `Horace_DOCS_BUILD_DIR/latex`. Unless I'm missing a use-case for generating the html/latex in separate places (and in separate places to their in-progress docs).

I'm also not sure what "in-progress" docs are, are these the `make` scripts?

hsautessella:
We're not running the doc builds on this PR, should we be?

oerc0122:
> We're not running the doc builds on this PR, should we be?

I have been intermittently running the build on the branch pipeline, but yes, I haven't done it more recently

nickbattam-tessella (approved):

PR #604, opened 2021-03-01
Set default values to avoid error

oerc0122:
Add default values to the Jenkinsfile to avoid assigning `null` in the environment


nickbattam-tessella (approved):

hsautessella (approved):
Hopefully this fixes things, it certainly shouldn't make it worse

Issue #550, opened 2021-02-07
Implement sqw_eval such that is can be performed on SQW objects whose pixels cannot all be stored in memory.


cmarooney-stfc:
Implement sqw_eval such that is can be performed on SQW objects whose pixels cannot all be stored in memory.

Child of #434

PR #610, opened 2021-03-04
Refactor npix chunking logic in cut

hsautessella:
A function has been written that splits an `npix` array into chunks whose sums are less than a given value. This refactors out the logic, originally written for cut, so that the functionality can be used elsewhere.

You'll notice there's a return value in `split_npix_full_bins` that is not used in `cut_accumulate_data_`. That return value is there as it is used in my work on #550.


horace_core/sqw/@sqw/private/split_npix_full_bins.m
```diff
@@ -0,0 +1,83 @@
...
+% that value will comprise its own sub-array.
+%
+% Input:
+% ------
+% npix          An array of integers (usually sqw.data.npix)
+% page_size     An integer specifying the maximum sum for each sub-array
+% num_pix       The sum of npix [optional].
+%               If this argument is not given it is calculated on-the-fly.
+%               In most cases, this value is cached for sqw objects in
+%               sqw.data.pix.num_pixels.
```
oerc0122, line 16: If it's stored, is this attached to the object? Can it just be retrieved?

horace_core/sqw/@sqw/private/split_npix_full_bins.m
```diff
@@ -0,0 +1,83 @@
...
+%               calculated this quantity (defines upper bounds of bin edges).
+%               This should be a vector.
+%
+% Output:
+% -------
+% npix_chunks  Cell array of vectors. Concatenation of these arrays will be
+%              equal to a flattened version the inputted npix array, i.e.
+%              npix(:).
+% idxs         The indices at which the input npix array was "split". Has size
+%              [2, n], where n is numel(npix_chunks). Each idxs(i, :) are the
```
oerc0122, line 29: Each "is"

oerc0122 (approved):

horace_core/sqw/@sqw/private/split_npix_full_bins.m
nickbattam-tessella, line 29: Ordering is important here is `(1, n)` the upper or lower index?

The description looks to be transposed from where it is used above: there `sub_bin_idxs(1, iter):sub_bin_idxs(2, iter)`
so `Each idxs(:, i) are the lower and upper indices ...`

horace_core/sqw/@sqw/private/split_npix_full_bins.m
```diff
@@ -0,0 +1,83 @@
...
+% >> npix = [3, 2, 0, 6, 0, 5, 3, 1, 1, 4, 2, 3, 0];
+% >> page_size = 11;
+% >> [chunks, idxs] = split_npix_full_bins(npix, page_size)
+%   chunks =
+%       { [3, 2, 0, 6, 0], [5, 3, 1, 1], [4, 2, 3, 0] }
+%   idxs =
+%       1     6    10
+%       5     9    13
+%
+if ~exist('num_pix', 'var')
```
nickbattam-tessella, line 43: Are there use-cases for `num_pix` and `npix_cum_sum` to be optional parameters? There's a lot of additional complexity introduced by potentially needing to calculate them

horace_core/sqw/@sqw/private/cut_accumulate_data_.m
```diff
@@ -148,7 +128,7 @@
         % Retain only the pixels that contributed to the cut
         pix_retained{iter} = candidate_pix.get_pixels(ok);
         pix_ix_retained{iter} = ix;
-    end
+        end
```
nickbattam-tessella, line 131: the code between `else` and this `end` needs indenting too

horace_core/sqw/@sqw/private/split_npix_full_bins.m
hsautessella: Ah, well spotted, cheers

horace_core/sqw/@sqw/private/split_npix_full_bins.m
hsautessella: It's just a relic of me originally deriving the values in this function, then realizing I already had them calculated where I was calling it.

Since this is a private SQW method and not a general "split this array of counts" utility, you're probably right, we can make the optional args compulsory. Unless we think this function is better as a utility - still not really sure where the best place for it is

horace_core/sqw/@sqw/private/split_npix_full_bins.m
hsautessella: Not in the case of cut. The output of  `proj.get_nbin_range` in [`cut_accumualte_data_`](https://github.com/pace-neutrons/Horace/pull/610/files#diff-b1ef44eea1c119f36c058dfaedcf05f4b743713d5fca2d1b905424a7e6ee982cR43) corresponds to bin edges for the output cut. So we have an `npix` array, but it's not attached to an object yet

abuts (commented):
All looks useful, but where is the unit test for this function?

hsautessella:
> _All looks useful, but where is the unit test for this function?_

It's a private method on `sqw`, so not accessible for unit testing. We could consider extracting the function and generalizing it as a utility. That way it would be directly testable. I'm not sure where the best place for it would be in that case, just `horace_core/utilities`?

horace_core/utilities/split_counts.m
```diff
@@ -0,0 +1,111 @@
...
+%              counts(:).
+% idxs         The indices at which the input counts array was "split". Has size
+%              [2, n], where n is numel(chunks). Each idxs(:, i) is the
+%              upper and lower index into counts of chunks{i}.
+%
+% Example:
+% --------
+% >> counts = [3, 2, 0, 6, 0, 5, 3, 1, 1, 24, 4, 2, 3, 0];
+% >> max_counts = 11;
+% >> [chunks, idxs] = split_counts(counts, max_counts, sum(counts), cumsum(counts))
```
nickbattam-tessella, line 31: This is a clumsy API: `data, maxsize, [f(g(data)), [g(data)]]` 

The only place this function is used (currently) doesn't use the total and cumulative counts data it calculates for anything else other this function call.

Removing the two optional args will simplify the implementation too as you are guaranteed the calculated values are correct and consistent with the `counts` array argument.

nickbattam-tessella: If future use-cases need total/cumsum data it could be returned from this function as optional additional return values.

nickbattam-tessella (changes requested):

nickbattam-tessella (approved):

_test/test_utilities/test_split_counts.m
```diff
@@ -0,0 +1,86 @@
...
+        assertExceptionThrown(f, 'MATLAB:expectedScalar');
+    end
+
+    function test_error_if_max_counts_is_zero(~)
+        counts = ones(1, 10);
+        f = @() split_counts(counts, 0);
+        assertExceptionThrown(f, 'MATLAB:expectedPositive');
+    end
+
+    function test_chunking_correct_if_a_count_gt_max_counts(~)
```
oerc0122, line 54: May seem silly, but what happens if counts is not an int? Error? Truncating if float?

horace_core/utilities/split_counts.m
```diff
@@ -0,0 +1,82 @@
+function [chunks, idxs] = split_counts(counts, max_counts)
```
oerc0122, line 1: Is this explicitly splitting counts now, even if that's its only current use case, or is it a general `split_chunks`?

horace_core/utilities/split_counts.m
```diff
@@ -0,0 +1,82 @@
...
+%SPLIT_COUNTS Split the given array of counts such that the sum of each
+%  sub-array has a maximum of max_counts or has length 1.
+%
+% The counts counts are not split across bin boundaries - i.e. each sub-array
```
oerc0122, line 5: Counts counts?

oerc0122 (approved):

_test/test_utilities/test_split_counts.m
hsautessella: I don't see a reason why the algorithm shouldn't just work, whether `counts` are ints or not. I can add a test to this effect.

Perhaps the function should be renamed to something like `split_vector_on_sum` or just `split_vector`?

oerc0122: I didn't mean counts, I had misunderstood what the function was doing. Ignore me.

horace_core/utilities/split_counts.m
```diff
@@ -0,0 +1,82 @@
...
+if max_chunks == 1
+    % Only one chunk of data, return it
+    chunks = {counts};
+    idxs = [1; numel(counts)];
+    return
+end
+
+chunks = cell(1, max_chunks);
+idxs = zeros(2, max_chunks);
+iter = 1;
```
cmarooney-stfc: It would be a bit more transparent if iter started at zero and was incremented at the start of the while loop. Then you wouldn't have to say iter-1 at lines 81,82, just iter.

horace_core/utilities/split_counts.m
```diff
@@ -0,0 +1,82 @@
...
+
+    if start_idx > end_idx
+        % Happens where bin size > page size
+        end_idx = start_idx;
+    end
+
+    chunks{iter} = counts(start_idx:end_idx);
+    idxs(:, iter) = [start_idx, end_idx];
+
+    % Decrement the cumulative sum with the number of values we've allocated
```
cmarooney-stfc: add comment "so that the find for end_idx will look for the next count > max_counts".

horace_core/utilities/split_counts.m
```diff
@@ -0,0 +1,82 @@
...
+    cumulative_counts = cumulative_counts - cumulative_counts(end_idx);
+    iter = iter + 1;
+end
+
```
cmarooney-stfc: Presuming this crops the end of chunks if iter<max_chunks - good to say so

horace_core/utilities/split_counts.m
```diff
@@ -0,0 +1,82 @@
...
+    % Find first index where cumulative sum is greater than max_counts
+    end_idx = ...
+        (start_idx - 1) + find(cumulative_counts(start_idx:end) > max_counts, 1) - 1;
+    if isempty(end_idx)
+        % If no elements of cumulative sum > max_counts, this is the final chunk
+        end_idx = numel(cumulative_counts);
+    end
+
+    if start_idx > end_idx
+        % Happens where bin size > page size
```
cmarooney-stfc: I do not understand this - more explanation will be good

horace_core/utilities/split_counts.m
```diff
@@ -0,0 +1,82 @@
...
+    idxs = zeros(2, 0);
+    return;
+end
+
+validateattributes(counts, {'numeric'}, {'vector', 'nonnegative'});
+validateattributes(max_counts, {'numeric'}, {'scalar', 'positive'});
+
+cumulative_counts = cumsum(counts);
+total_counts = cumulative_counts(end);
+max_chunks = ceil(total_counts/max_counts);
```
cmarooney-stfc: While this value for max_chunks is vaguely intuitive, I keep thinking there may be cases where it fails. Is it possible to say why it will always work?

cmarooney-stfc (changes requested):

horace_core/utilities/split_counts.m
hsautessella: You're right. In general the max number of iterations is `numel(counts)` - each sub-array has length 1. We'd probably be over-allocating (by quite a lot) more often than not if we set `max_chunks` to `numel(counts)`. I'll have to think about this a bit more

horace_core/utilities/split_vector.m
```diff
@@ -0,0 +1,94 @@
...
+    idxs = [1; numel(vector)];
+    return
+end
+
+chunks = cell(1, max_num_chunks);
+idxs = zeros(2, max_num_chunks);
+iter = 0;
+end_idx = 0;
+while end_idx < max_num_chunks
+    iter = iter + 1;
```
nickbattam-tessella, line 59: why `iter` not `chunk`?

horace_core/sqw/@sqw/private/cut_accumulate_data_.m
```diff
@@ -46,71 +49,42 @@
...
 
-% Get the cumulative sum of pixel bin sizes and work out how many
-% iterations we're going to need
-cum_bin_sizes = cumsum(bin_ends - bin_starts);
 block_size = obj.data.pix.base_page_size;
-max_num_iters = ceil(cum_bin_sizes(end)/block_size);
+% Get indices in order to split the candidate bin ranges into pixel page sized
+% chunks
+[~, sub_bin_idxs] = split_vector(bin_ends - bin_starts, block_size);
+num_iters = size(sub_bin_idxs, 2);
```
nickbattam-tessella, line 56: is this `num_chunks` or `num_iters`? The method is described as "paged sized chunks"

oerc0122 (approved):

horace_core/utilities/split_vector.m
```diff
@@ -0,0 +1,100 @@
...
+
+% We need the number of chunks we're to return so we can allocate output arrays.
+% Calculating the real number of chunks required is non-trivial. We can get the
+% minimum number using `sum(vector)/max_chunk_sum`, but this will result in us
+% under-allocating the output arrays.
+% If we under-estimate the number of chunks, we will end up re-allocating the
+% output arrays as they grow, so we'll need two instances of each array. Hence
+% it's more memory efficient to over-allocate to the maximum number of chunks
+% and then crop.
+max_num_chunks = numel(vector);
```
nickbattam-tessella, line 50: I would extend this to say "... so initialize assuming all chunks will be contain one index" or something to explain why `numel(vector)` is the right value

horace_core/utilities/split_vector.m
```diff
@@ -0,0 +1,100 @@
+function [chunks, idxs] = split_vector(vector, max_chunk_sum)
```
nickbattam-tessella, line 1: is `vector` a good variable name? vector of what? it's a keyword in a lot of languages which may lead to confusion.

horace_core/utilities/split_vector.m
```diff
@@ -0,0 +1,100 @@
...
+% its own sub-vector.
+%
+% Input:
+% ------
+% vector       A vector of non-negative values.
+% max_chunk_sum      A positive value specifying the maximum sum for each sub-vector.
+%
+% Output:
+% -------
+% chunks       Cell array of vectors. Concatenation of these vectors will be
```
nickbattam-tessella, line 16: it's not an array of vectors, it's a cell array of arrays; if you're using vector as a 'type' here that makes the argument name even more confusing

oerc0122: Again, given you can do `validateattributes(x, {'numeric'}, {'vector'}`, I think this is well defined. Perhaps label that it's a numeric vector, though.

hsautessella: A "vector" in Matlab is a [1, n] or [n, 1] array. So it's not a type (or a reserved name), it's a commonly used name for a special case of `array`. See the Matlab function `isvector`.

PR #611, opened 2021-03-08
484 parse args branch

oerc0122:
Write `is_cut_equal` to use `inputParser` object


abuts (commented):
projection may be structure and may be instance of aProjection class. 
All cut parameters may be empty. 
projection can be missing (not sure about that, if cut_sqw accepts empty projection or you need to go directly to cut (IMHO, it should accept empty) 
tol can be negative (defines relative-absolute), not sure if it can be 0

I have no doubt that parameter parsing can be done through parser and this way, but this kind of testing needs a unit test of its own while Toby test route takes 2 rows (ok  + 3 to throw error)

I think we need to use Toby's projection

_test/common_functions/is_cut_equal.m
```diff
@@ -12,48 +14,85 @@
...
+    if ~iscell(f2), f2={f2}; end
+
+    proj = p.Results.proj;
+    pN = {p.Results.p1, p.Results.p2, p.Results.p3, p.Results.p4};
+    pN = pN(~cellfun(@isempty, pN));
+    tol = p.Results.tol;
+    cutArgs = p.Results.cutArgs;
+end
+
+function ok = validate_sqw(inp)
```
nickbattam-tessella, line 87: This is validating an SQW object or file path

_test/common_functions/is_cut_equal.m
```diff
@@ -12,48 +14,85 @@
...
+    check = @(x)(isa(inp, 'sqw') || ischar(inp));
+    ok = check(inp) || (iscell(inp) && all(cellfun(check, inp)));
+end
+
+function ok = validate_proj(inp)
+    ok = isfield(inp, 'u') && isfield(inp, 'v');
+    if ok
+        validateattributes(inp.u, {'numeric'}, {'nonempty','vector','numel',3});
```
nickbattam-tessella, line 95: if `validateattributes` fails `ok` is not set to false

_test/common_functions/is_cut_equal.m
```diff
@@ -12,48 +14,85 @@
...
+    end
+    for i=1:numel(f2)
+        w2(i)=cut_sqw(f2{i},proj,pN{:},cutArgs{:});
+    end
+
+    w1tot=combine_cuts(w1);
+    w2tot=combine_cuts(w2);
+
+    % To check equality, see if npix, s, e arrays are the same
+    if equal_to_tol(w1tot.data.npix,w2tot.data.npix,tol) &&...
```
nickbattam-tessella, line 33: this would be easier to read with some whitespace between the args

_test/common_functions/is_cut_equal.m
```diff
@@ -12,48 +14,85 @@
...
         else
-            ok = false;
-            mess=sprintf(['Pixels parameters of two cuts are different:\n',...
-                '   npix1=%d and npix2 = %d\n',...
-                'pix_range1 = [%f  %f %f %f; pix_range2=[%f  %f %f  %f;\n',...
-                '              %f  %f %f %f]             %f  %f %f  %f]'],...
-                w1tot.data.pix.num_pixels,w2tot.data.pix.pix_range,...
-                w1tot.data.pix.pix_range(1,:),w2tot.data.pix.pix_range(1,:),...
-                w1tot.data.pix.pix_range(2,:),w2tot.data.pix.pix_range(2,:));
+            if equal_to_tol(w1tot.data.pix.pix_range,w2tot.data.pix.pix_range,tol) && ...
```
nickbattam-tessella, line 41: whitespace would help here too

_test/common_functions/is_cut_equal.m
```diff
@@ -12,48 +14,85 @@
...
+                              'pix_range1 = [%f  %f %f %f; pix_range2=[%f  %f %f  %f;\n',...
+                              '              %f  %f %f %f]             %f  %f %f  %f]'],...
+                             w1tot.data.pix.num_pixels,w2tot.data.pix.pix_range,...
+                             w1tot.data.pix.pix_range(1,:),w2tot.data.pix.pix_range(1,:),...
+                             w1tot.data.pix.pix_range(2,:),w2tot.data.pix.pix_range(2,:));
+            end
         end
+    else
+        ok=false;
+        mess='One or more of npix, s, e are not the same';
```
nickbattam-tessella, line 58: the detail of the two not-equal messages is wildly different: array elements vs one of these four variables.
Can more detail be added here to name which is/are not equal?

_test/common_functions/is_cut_equal.m
oerc0122: In a spout of MATLAB genius, `validateattributes` raises if it's not met, it cannot set ok. I guess, I should probably catch and set `ok` to not `ok`.

nickbattam-tessella:
Having `proj` as an optional mid-list parameter is specifically prohibited by ADR 0022.

If we need support for proj and or the pis being omitted we can rewrite the API using keyword arguments or create a "with-" and "without-" proj argument function: this is covered by https://github.com/pace-neutrons/Horace/issues/577

oerc0122:
> projection may be structure and may be instance of aProjection class.
> All cut parameters may be empty.
> projection can be missing (not sure about that, if cut_sqw accepts empty projection or you need to go directly to cut (IMHO, it should accept missing, so parameters will look p1,p2,p3,p4)
> tol can be negative (defines relative-absolute), not sure if it can be 0
> 
> I have no doubt that this type of parameter parsing can be done through Matlab parseInput, but this kind of testing needs a unit test of its own, while Toby test route takes 2 rows (ok + 3 to throw error) -- I can not believe you can make a lot of errors there.
> 
> I think we need to use Toby's parser -- it is much easier

I would somewhat disagree on these points. Part of the reason it is set up the way it is which does not match your expectations is that those expectations are not clear from its current definition, but you were able to see immediately that they did not match. 

Those parsing restrictions were based on the description I was given of cut's inputs. 

`tol` should now be defined as a vector, passing it as a scalar is considered backwards compatibility so we should move away from it for consistency. Strictly this new arrangement allows scalars passing still, as they are scalar, 1 element rank-1, 1x1 element rank-2 in MATLAB.  
```
%           is accepted.
%             Examples:
%               [1e-4, 1e-6]    absolute 1e-4 or relative 1e-6 required
%               [1e-4, 0]       absolute 1e-4 required
%               [0, 1e-6]       relative 1e-6 required
%               [0, 0]          equality required
%               0               equivalent to [0,0]
%
%           For backwards compatibility, a scalar tolerance can be given
%           where the sign determines absolute or relative tolerance
%               +ve : absolute tolerance  abserr = abs(a-b)
%               -ve : relative tolerance  relerr = abs(a-b)/max(abs(a),abs(b))
%             Examples:
%               1e-4            absolute tolerance, equivalent to [1e-4, 0]
%               -1e-6           relative tolerance, equivalent to [0, 1e-6]
```

nickbattam-tessella (approved):

cmarooney-stfc (approved):

PR #613, opened 2021-03-08
Update clone_herbert_branch.ps1

oerc0122:
Add configure to Herbert build on Horace Jenkins


cmarooney-stfc (approved):
does what it says on the tin

PR #614, opened 2021-03-08
Add PixelData.set_data method

hsautessella:
As part of my work on #550 I need a `set_` method on pixels that can set field data across page boundaries. This method is the inverse of `PixelData.get_data` so they share a fair amount of logic.

refs #550

### DIFF
--- a/horace_core/sqw/PixelData/@PixelData/private/check_pixel_fields_.m
+++ b/horace_core/sqw/PixelData/@PixelData/private/check_pixel_fields_.m
@@ -1,0 +1,28 @@
+function check_pixel_fields_(obj, fields)
+%CHECK_PIXEL_FIELDS_ Check the given field names are valid pixel data fields
+% Raises error with ID 'HORACE:PIXELDATA:invalid_field' if any fields not valid.
+%
+%
+% Input:
+% ------
+% fields    A cellstr of field names to validate.
+%
+num_bad_fields = 0;
+bad_fields = cell(1, numel(fields));
+for i = 1:numel(fields)
+    field = fields{i};
+    if ~obj.FIELD_INDEX_MAP_.isKey({field})
+        bad_fields{num_bad_fields + 1} = field;
+        num_bad_fields = num_bad_fields + 1;
+    end
+end
+
+if num_bad_fields > 0
+    valid_fields = obj.FIELD_INDEX_MAP_.keys();
+    error( ...
+        'HORACE:PIXELDATA:invalid_field', ...
+        'Invalid pixel field(s) {''%s''}.\nValid keys are: {''%s''}', ...
+        strjoin(bad_fields(1:num_bad_fields), ''', '''), ...
+        strjoin(valid_fields, ''', ''') ...
+    );
+end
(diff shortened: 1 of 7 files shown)


horace_core/sqw/PixelData/@PixelData/private/check_pixel_fields_.m
```diff
@@ -0,0 +1,28 @@
+function check_pixel_fields_(obj, fields)
```
oerc0122, line 1: Would this be a useful extension to general objects? I.e. make this a public function and replace references to "pixeldata" with `class(obj)` ?

horace_core/sqw/PixelData/@PixelData/private/logical_to_normal_index_.m
```diff
@@ -0,0 +1,15 @@
+function abs_pix_indices = logical_to_normal_index_(obj, logical_indices)
+%ABS_PIX_INDICES Convert the given logical indices to normal indices
```
oerc0122, line 2: What does this mean? Example?

oerc0122: Is it converting:
`[true, true, false, false, true]`
to
`[1,3,5]`?

horace_core/sqw/PixelData/@PixelData/set_data.m
```diff
@@ -0,0 +1,112 @@
...
+    NO_INPUT_INDICES = -1;
+
+    validateattributes(pix_fields, {'cell', 'char', 'string'}, {'nonempty'});
+    validateattributes(data, {'numeric'}, {});
+
+    parser = inputParser();
+    parser.addOptional( ...
+        'abs_pix_indices', ...
+        NO_INPUT_INDICES, ...
+        @(x) validateattributes(x, {'numeric', 'logical'}, {'integer', 'nonnegative'}) ...
```
oerc0122, line 64: Question more for my own sake, does the check function only apply IF the argument is present?

horace_core/sqw/PixelData/@PixelData/set_data.m
```diff
@@ -0,0 +1,112 @@
...
+        obj.advance();
+        set_page_data(obj, field_indices, data, abs_pix_indices);
+    end
+else
+    if abs_pix_indices == NO_INPUT_INDICES
+        obj.data_(field_indices, 1:end) = data;
+    else
+        obj.data_(field_indices, abs_pix_indices) = data;
+    end
+end
```
oerc0122, line 48: Indentation?

oerc0122 (approved):

horace_core/sqw/PixelData/@PixelData/private/check_pixel_fields_.m
hsautessella: This function in particular is for `PixelData` data fields. These are tracked as the keys in a `containers.Map`, the map's values being the corresponding index/indices into the underlying data array. So I'm checking something more specific in this case.

horace_core/sqw/PixelData/@PixelData/private/logical_to_normal_index_.m
hsautessella: I've added an example to the function help - hopefully that clarifies things

oerc0122 (approved):

horace_core/sqw/PixelData/@PixelData/private/logical_to_normal_index_.m
```diff
@@ -0,0 +1,15 @@
...
+%
```
cmarooney-stfc, line 23: Not clear why it would be ok to have this array larger than expected as long as the extras aren't logicals. It sounds as if it should be an error to get here in the first place

Issue #578, opened 2021-02-16
Use a single argument parsing approach across the code


nickbattam-tessella:
The Horace code uses a number of different argument parsing routines:
- raw `varargin`/`nargin` processing
- `parse_arguments`
- `parse_args_namelist`
- `parse_char_options`
- `parse_keywords` (deprecated, no longer used)
- `inputParser`

The codebase should be updated to use a consistent argument parsing scheme and all other parsers deleted.
An ADR will define which argument parser to use.

There will be a small number of cases where the standard parser may not be applicable. In these limited cases an alternative may be used and comments must be included to justify the deviation. This is likely to be performance on low-level frequently called functions where it is not possible to restructure the code in such a way as to make that parsing unnecessary.

PR #619, opened 2021-03-10
fix sqw refs to allow sqw loader to work

cmarooney-stfc:
Additional changes to enable removal of skipTest from tests where the sqw object would not load. Tests where the dnd object would not work have had their skipTest relabelled or parts of the test commented out with a %skipTest("New dnd.... comment to identify them; they remain skipped but for different reasons.

### DIFF
--- a/_test/test_sqw_file/create_testdata.m
+++ b/_test/test_sqw_file/create_testdata.m
@@ -59,9 +59,12 @@ else
     save(sqw2d_arr(1),sqw2d_name{1})
     save(sqw2d_arr(2),sqw2d_name{2})
     
+    %{
+    skipTest("New dnd objects not yet supported");
     save(d1d_arr(1),d1d_name{1})
     save(d1d_arr(2),d1d_name{2})
     
     save(d2d_arr(1),d2d_name{1})
     save(d2d_arr(2),d2d_name{2})
+    %}
 end
--- a/_test/test_sqw_file/test_file_input.m
+++ b/_test/test_sqw_file/test_file_input.m
@@ -21,7 +21,6 @@ classdef  test_file_input < TestCase
 
     methods
         function obj = test_file_input(varargin)
-            %skipTest("New sqw loader not implemented");
             if nargin > 0
                 name = varargin{1};
             else
@@ -128,12 +127,10 @@ classdef  test_file_input < TestCase
         end
         %
         function obj = test_normal_file_cut(obj)
-            %skipTest("Infinite Loop");
             obj=obj.input_operations();
         end
         %
         function obj = test_crossbuf_file_cut(obj)
-            %skipTest("Infinite Loop");
             hc = hor_config;
             mem_chunk_size = hc.mem_chunk_size;
             clob = onCleanup(@()set(hor_config,'mem_chunk_size',mem_chunk_size));
--- a/_test/test_sqw_file/test_file_input.m
+++ b/_test/test_sqw_file/test_file_input.m
@@ -21,7 +21,7 @@ classdef  test_file_input < TestCase
 
     methods
         function obj = test_file_input(varargin)
-            skipTest("New sqw loader not implemented");
+            %skipTest("New sqw loader not implemented");
             if nargin > 0
                 name = varargin{1};
             else
@@ -128,10 +128,12 @@ classdef  test_file_input < TestCase
         end
         %
         function obj = test_normal_file_cut(obj)
+            %skipTest("Infinite Loop");
             obj=obj.input_operations();
         end
         %
         function obj = test_crossbuf_file_cut(obj)
+            %skipTest("Infinite Loop");
             hc = hor_config;
             mem_chunk_size = hc.mem_chunk_size;
             clob = onCleanup(@()set(hor_config,'mem_chunk_size',mem_chunk_size));
@@ -210,17 +212,21 @@ classdef  test_file_input < TestCase
 
             % Cut of dnd objects or files
             % ---------------------------
+            %{
+            skipTest("New dnd object not yet supported");
             d1_d=cut(obj.d2d_arr(2),[0.5,0,1.2],[170,180]);
             d1_d_h=cut(obj.d2d_arr(2),[0.5,0,1.2],[170,180]);
             d1_d_d=cut_dnd(obj.d2d_arr(2),[0.5,0,1.2],[170,180]);
             d1_f_h=cut(obj.d2d_name{2},[0.5,0,1.2],[170,180]);
             d1_f_d=cut_dnd(obj.d2d_name{2},[0.5,0,1.2],[170,180]);
-
+            %}
             function call_cut_sqw(w)
                 % We want to call cut_sqw with an output arg, so no lambda
                 d1_d_s=cut_sqw(w,[0.5,0,1.2],[170,180]);
             end
             
+            %{
+            skipTest("New dnd object not yet supported");
             assertExceptionThrown(@() call_cut_sqw(obj.d2d_arr(2)), 'HORACE:cut_sqw');
             assertExceptionThrown(@() call_cut_sqw(obj.d2d_name{2}), 'HORACE:cut_sqw');
 
@@ -228,6 +234,7 @@ classdef  test_file_input < TestCase
             if ~equal_to_tol(d1_d,d1_d_d), assertTrue(false,'Error in functionality'), end
             if ~equal_to_tol(d1_d,d1_f_h,'ignore_str',1), assertTrue(false,'Error in functionality'), end
             if ~equal_to_tol(d1_d,d1_f_d,'ignore_str',1), assertTrue(false,'Error in functionality'), end
+            %}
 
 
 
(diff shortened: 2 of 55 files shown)


nickbattam-tessella (commented):
Do the changes here mean that the loader will now no longer work with old SQW objects?
If so are there additional code changes (e.g. removing the class converting shim on the Herbert sqw_refactor_branch) required?

@nickbattam-tessella I am assuming that all .mat files in the _test folder contain objects labelled sqw saved as old-style objects. These are now being loaded as new-style objects - this is done by passing them to a sqw(S) constructor call which converts them to new sqw objects. The object actually extracted from the file is a struct with the required data fields. A conversion operator for the resulting loaded objects has not been needed since the loadobj method was added to new sqw. Hence further management of this load does not appear necessary. We may need to upgrade the load when we add more to new sqw but it is ok for the moment. I think the additional question you are asking is - have any sqw_old objects been explicitly saved with that label, and would we want to retrieve them. I'm guessing that the easy thing is to wait for something to fail with that case, and fix it then.

_test/test_sqw_file/create_testdata.m
```diff
@@ -59,9 +59,12 @@
     save(sqw2d_arr(1),sqw2d_name{1})
     save(sqw2d_arr(2),sqw2d_name{2})
     
+    %{
```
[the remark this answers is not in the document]
cmarooney-stfc: @nickbattam-tessella As noted in the description comment, this is a label to group all similar issues so they can be found and tested. They will be removed when this phase of updating is complete.

horace_core/sqw/file_io/sqw_formats_factory.m
```diff
@@ -45,10 +45,10 @@
         %
         % Rules to load/save different classes:
         % sqw2 corresponds to sqw file in indirect mode with varying efixed
-        written_types_ = {'sqw_old','sqw2','dnd','d0d_old','d1d_old','d2d_old','d3d_old','d4d_old'};
+        written_types_ = {'sqw', 'sqw_old','sqw2','dnd','d0d_old','d1d_old','d2d_old','d3d_old','d4d_old'};%,'sqw'};
```
nickbattam-tessella, line 48: This has support for `sqw` and `sqw_old`, but the `if` on line 105 below only supports `sqw`.
Ia `sqw_old` support required at all?

horace_core/sqw/file_io/@dnd_binfile_common/dnd_binfile_common.m
```diff
@@ -134,7 +134,7 @@
                 error('SQW_FILE_IO:runtime_error',...
                     'dnd_binfile_common:init_from_sqw_obj method should be invoked with at least an existing sqw or dnd object provided');
             end
-            if isa(varargin{1},'sqw_old') || is_sqw_struct(varargin{1})
+            if isa(varargin{1},'sqw') || is_sqw_struct(varargin{1})
```
nickbattam-tessella, line 137: do these checks need to be `sqw_old` || `sqw`?

horace_core/sqw/@sqw/section.m
```diff
@@ -0,0 +1,159 @@
...
+
+for n=1:numel(win)
+    [ndim,sz]=dimensions(win(n));   % need to get sz array specific for each element in array win
+    % Get section parameters and axis arrays:
+    % The input sectioning arguments refer to the *display* axes; these must be converted to the relevant plot axes in the algorithm
+    irange = zeros(2,ndim);
+    array_section = cell(1,ndim);
+    p=win(n).data.p;   % extract bin boundaries
+    for i=1:nargs
+        if isempty(varargin{i}) || (isscalar(varargin{i}) && isequal(varargin{i},0))
```
nickbattam-tessella, line 69: what is `varargin{1}`

horace_core/sqw/@sqw/section.m
```diff
@@ -0,0 +1,159 @@
...
+            else
+                error (['No data along axis ',num2str(i),' in the range [',num2str(varargin{i}(1)),',',num2str(varargin{i}(2)),']'])
+            end
+        else
+            error (['Limits parameter for axis ',num2str(i),' must be zero or a pair of numbers'])
+        end
+    end
+
+    % Section signal, variance and npix arrays
+    wout(n).data.s = win(n).data.s(array_section{:});
```
nickbattam-tessella, line 95: as `section` is a dnd method too; changing these to `data_` would allow this method to move to the SQWDnDBase class

horace_core/sqw/@sqw/private/cut_sqw_sym_main_single.m
```diff
@@ -138,6 +138,72 @@
...
+% Look only at irun, idet, ien, as the pix coordinates may have been altered by
+% the symmetrisation algorithm, depending on where that is done.
+nend = cumsum(npixtot);
+nbeg = nend - npixtot + 1;
+pixind = zeros(npixtot_all,3);
+fields = {'run_idx', 'detector_idx', 'energy_idx'};
+% pix_range = PixelData.EMPTY_RANGE_;
+for i=1:nw
+    pixind(nbeg(i):nend(i),:) = varargin{i}.data.pix.get_data(fields)';
+    %     loc_range = varargin{1}.data.img_range;
```
nickbattam-tessella, line 175: is this commented by you or historic?

horace_core/horace_function_utils/horace_function_call_method.m
```diff
@@ -93,19 +93,19 @@
 end
 
 if all(w.sqw_type)
-    dummy_obj=sqw_old;
+    dummy_obj=sqw;
 else
     ndims=w.ndims(1);
     if ndims==0
-        dummy_obj=d0d_old;
+        dummy_obj=d0d;
```
nickbattam-tessella, line 100: do these work?

This function (horace_function_call_method) should be disappearing as part of the DnD migration

nickbattam-tessella (commented):
Tests of the sqw (class) object should be in test_sqw_class rather than test_sqw folder.

@nickbattam-tessella I think that distinction is out of date. Now so much of the code is testing the new sqw object, it is testing that class. I would say that test_sqw_class should be reserved for tests specifically testing the functionality of the class methods. Uses of new sqw "in the wild" will simply be tests of sqw, doing exactly the same for sqw_old's replacement as their originals did for sqw_old. At the end of the upgrade we will expect test_sqw to do the same for new sqw as it did for sqw_old before we started. I am not aware of any new tests violating that distinction.

horace_core/sqw/@sqw/private/cut_sqw_sym_main_single.m
cmarooney-stfc: Historic

horace_core/horace_function_utils/horace_function_call_method.m
cmarooney-stfc: I think this was just a tidy up on my part. I did not test functionality outside their use (if any) in the existing tests. As I made the change, it must have fixed one or more tests.

horace_core/sqw/@sqw/section.m
cmarooney-stfc: It says varargin{i} not varargin{1} - it is the relevant [lo,hi] as described in the args after win

horace_core/sqw/@sqw/section.m
cmarooney-stfc: I'm sure this is generally true, but in fact the dnds differ from sqw in explicitly returning the win argument if nargin==0, with d0d doing that regardless due to the dimensionality. I would rather leave this to a dedicated exercise. This PR is just about maximising the number of working tests subject to making sqw_old into sqw now the changes for multifit have been done.

horace_core/sqw/file_io/@dnd_binfile_common/dnd_binfile_common.m
cmarooney-stfc: We have basically moved a lot of the code so it is using sqw rather than sqw_old. The more sensible thing at this point would be to detect sqw_old as an error condition. But as this change was made following a test failure, this functionality already tests for sqw_old as a fail state, and it should be enough just to leave this code as-is.

horace_core/sqw/@sqw/section.m
```diff
@@ -0,0 +1,159 @@
...
+for n=1:numel(win)
+    [ndim,sz]=dimensions(win(n));   % need to get sz array specific for each element in array win
+    % Get section parameters and axis arrays:
+    % The input sectioning arguments refer to the *display* axes; these must be converted to the relevant plot axes in the algorithm
+    irange = zeros(2,ndim);
+    array_section = cell(1,ndim);
+    p=win(n).data.p;   % extract bin boundaries
+    for i=1:nargs
+        if isempty(varargin{i}) || (isscalar(varargin{i}) && isequal(varargin{i},0))
+            pax=win(n).data.dax(i);
```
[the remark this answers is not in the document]
cmarooney-stfc: See further reply on section.m below. As I'm not sure we are in a position to put section in the base class yet, I would leave this to a wholesale refactor of this functionality.

horace_core/sqw/file_io/sqw_formats_factory.m
cmarooney-stfc: I think you mean line 195. My approach was that if an if was testing that the object was sqw_old, we should change it to sqw as that is what we want to treat going forward - that is what line 195 does. Here the aim was to support sqw_old when it occurs correctly so nothing breaks for no good reason. All the changes are the result of fixing a broken test. So sqw_old is retained in this list. When we remove sqw_old this can go, but I'm not breaking any tests at the moment to initiate a clearout.

nickbattam-tessella: I wanted to confirm that the rest of the file and file handling routines were able to correctly handle both sqw and sqw_old if both options were included here; i.e are there other places where we have replaced sqw_old with sqw rather than sqw||sqw_old

Thinking here about changes in other files e.g. dnd_binfile_common as well as elsewhere in this one

horace_core/sqw/@sqw/section.m
nickbattam-tessella: ok, this raw varargin handling is slightly opaque but will be resolved as part of #578

_test/test_sqw_file/test_file_input.m
```diff
@@ -21,7 +21,7 @@
 
     methods
         function obj = test_file_input(varargin)
-            skipTest("New sqw loader not implemented");
+            %skipTest("New sqw loader not implemented");
```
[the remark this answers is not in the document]
cmarooney-stfc: I am removing them

nickbattam-tessella:
PR build notes this has added 200 MLint warnings (+330/-129)

PR #621, opened 2021-03-11
Disable unstable tests

oerc0122:
Disabled:
test_gen_sqw_workflow/test_gen_sqw_accumulate_sqw_herbert
test_gen_sqw_workflow/test_gen_sqw_accumulate_sqw_parpool

Until such a time as they can be stabilised.

### DIFF
--- a/_test/disabled_tests.md
+++ b/_test/disabled_tests.md
@@ -2,7 +2,7 @@
 
 - test_change_crystal_1a.m
         - test_u_alighnment_tf_way (no ticket) -- Why disabled?
-	- test_u_alighnment (no ticket)
+    - test_u_alighnment (no ticket)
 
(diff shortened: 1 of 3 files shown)


nickbattam-tessella (approved):
Disabled tests should be added to `_test/disabled_tests.md`

hsautessella (approved):
I say we merge this. I've not seen that `test_cut` test fail before, if it starts failing more regularly we can disable that as well.

Is there an issue to fix and re-enable these tests? If not, could you open one and link it here?

nickbattam-tessella:
> I say we merge this. I've not seen that `test_cut` test fail before, if it starts failing more regularly we can disable that as well.

Reran the build and `test_cut` passed.

PR #623, opened 2021-03-12
Refactor npix page splitting

hsautessella:
This PR refactors some common code out of pixel paging operations and adds tests for it. Notice that the new `split_vector_fixed_sum` function is different to `split_vector_max_sum` (previously `split_vector`) in that every chunk's sum is equal to the given `chunk_sum` value (except for the last chunk which holds the remainder).

The logic in this function was written many months ago, and can be hard to follow (it was for me on a re-read), but I've tried to comment it as best I can and have left a "run-through" of the algorithm at the bottom of the function. Let me know if things are unclear, or you can think of a clearer way of attacking the problem.


horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -0,0 +1,217 @@
+function [chunks, idxs] = split_vector_fixed_sum(numeric_vector, chunk_sum, cumulative_sum)
+%SPLIT_VECTOR_FIXED_SUM Split the given vector into sub-vectors such that each
+% sub-vector has sum 'chunk_sum'. If chunk sum does not divide the sum of the
```
oerc0122, line 3: Exact or at least (most)?

horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -0,0 +1,217 @@
...
+    return;
+end
+
+validateattributes(numeric_vector, {'numeric'}, {'vector', 'nonnegative'});
+validateattributes(chunk_sum, {'numeric'}, {'scalar', 'positive'});
+
+if ~exist('cumulative_sum', 'var')
+    cumulative_sum = cumsum(numeric_vector);
+end
+vector_sum = cumulative_sum(end);
```
oerc0122, line 52: Is there an easy check you can do to verify cumulative_sum is reasonable? Or do we rely on good faith?

oerc0122: e.g. 
```matlab
split_vector_fixed_sum([1,2,4,4], 3, [1])
```

horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -0,0 +1,217 @@
...
+            chunk(end) = leftover_end;
+        end
+        chunks{chunk_num} = chunk;
+    end
+    idxs(:, chunk_num) = [start_idx; end_idx];
+    allocated = allocated + sum(chunks{chunk_num});
+end
+
+
+% Algorithm run through:
```
oerc0122: Not 100% convinced of the value of this, but it is doing no harm.

oerc0122 (approved):

nickbattam-tessella (commented):
Would `split_vector_fixed_size` and `split_vector` be clearer names?

Does the explanation of the algorithm belong in the code or in the documentation? There is a lot of comment in this split function, to the extent that is is obfuscating the code.

A lot of comments are describing WHAT is being done (e.g. "subtract sum of values we've allocated so that ....") rather than WHY it's being done (e.g. "offset cum_sum array so next positive index is in bucket of interest").

horace_core/utilities/split_vector_fixed_sum.m
hsautessella: The intention is that, generally, you do not use the argument. But there were some cases where the cumulative sum was used elsewhere in the calling function, which made it advantages to be able to just pass in here, rather than calculate it again.

horace_core/utilities/split_vector_fixed_sum.m
hsautessella: I found it very useful writing it out, but perhaps reading it doesn't have the same effect.

horace_core/sqw/PixelData/@PixelData/private/binary_op_double_.m
```diff
@@ -57,8 +57,7 @@
     % object.
     %
     [npix_chunks, idxs] = split_vector_fixed_sum(npix(:), obj.base_page_size, npix_cum_sum);
-    page_number = 1;
-    while true
+    for page_number = 1:numel(npix_chunks)
```
oerc0122, line 60: This is fundamentally changing the operation, was page_number not updating before?

horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -1,7 +1,7 @@
 function [chunks, idxs] = split_vector_fixed_sum(numeric_vector, chunk_sum, cumulative_sum)
 %SPLIT_VECTOR_FIXED_SUM Split the given vector into sub-vectors such that each
-% sub-vector has sum 'chunk_sum'. If chunk sum does not divide the sum of the
-% input vector, the final chunk's sum will be the remainder.
+% sub-vector's sum is exactly 'chunk_sum'. If chunk sum does not divide the sum
```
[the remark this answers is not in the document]
hsautessella: >_Values in the input vector may be split between chunks._

```matlab
>> split_vector_fixed_sum([1,3,100,10], 50)
    ans =

  1×3 cell array

    {[1   3  46]}    {[50]}    {[4  10]}
```

horace_core/sqw/PixelData/@PixelData/private/binary_op_double_.m
hsautessella: Yep, I'd forgotten to increment `page_number` when I refactored the function

horace_core/utilities/split_vector_fixed_sum.m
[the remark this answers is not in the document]
oerc0122: Ah, I'd missed that. Thanks

horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -0,0 +1,113 @@
...
+allocated = 0;
+for chunk_num = 1:num_chunks
+    remaining_sum = vector_sum - allocated;
+    chunk_sum = min(chunk_sum, remaining_sum);
+
+    if chunk_num == 1
+        % For the first iteration, we need >= to catch leading zeros
+        % This index may contribute its full value, or only part of its value
+        start_idx = (end_idx - 1) + find(cumulative_sum(end_idx:end) >= 0, 1);
+    else
```
cmarooney-stfc, line 67: Should be start_idx=1 as the find logical is always true

horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -0,0 +1,113 @@
...
+            leftover_end = 0;
+        end
+        chunks{chunk_num} = min(chunk_sum, numeric_vector(start_idx) - leftover_end);
+    else
+        % Add an empty element to the end, which will be used to hold a part of
+        % the end index's value. If there's no remainder on the end index's
+        % value, the empty element will be removed
+        chunk = [ ...
+            leftover_begin, ...
+            reshape(numeric_vector(start_idx + 1:end_idx - 1), 1, []), ...
```
cmarooney-stfc, line 98: Unclear why reshape is needed

horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -0,0 +1,113 @@
...
+            0 ...
+        ];
+
+        leftover_end = chunk_sum - sum(chunk);
+        if leftover_end == 0 && chunk_num ~= num_chunks
+            chunk(end) = [];
```
cmarooney-stfc, line 104: unclear how [] will be removed

cmarooney-stfc (changes requested):
Repeating briefly comments I made on #624

horace_core/utilities/split_vector_fixed_sum.m
nickbattam-tessella: @oerc0122 raises a good point.

We could easily add a check that the final value in cumsum is equal to sum - at a performance cost. How critical is the performance of this function? At least part of the reason for taking cumsum as an argument is to stop its unnecessary recalculation.

An alternative is to add a note to the function header that cumsum argument isn't validated and the behaviour if it's not correct is undetermined.

_test/test_utilities/test_split_vector_max_sum.m
```diff
@@ -35,39 +35,39 @@ function test_outputs_have_length_1_if_max_counts_eq_total_counts(~)
...
     end
 
     function test_error_if_max_counts_is_negative(~)
         vector = ones(1, 10);
-        f = @() split_vector(vector, -1);
+        f = @() split_vector_max_sum(vector, -1);
         assertExceptionThrown(f, 'MATLAB:expectedPositive');
     end
 
     function test_chunking_correct_if_a_count_gt_max_counts(~)
```
nickbattam-tessella, line 67: "chunking correct if count gt max counts" doesn't tell me what "correct is": does it give me a chunk with more than max elements, or multiple chunks with max elements?

horace_core/utilities/split_vector_fixed_sum.m
hsautessella: I ran some benchmarks for cumsum/summing a large array:
```matlab
>> x = rand(1, 1e8);
>> median(benchmark_function(@() sum(x), 10, 10))

ans =

    0.0249

>> median(benchmark_function(@() cumsum(x), 10, 10))

ans =

    0.2215
```

Looks like it's very cheap to get the sum, but that calls into question the value of providing the `cumulative_sum` argument anyway.

The callers all use `cumulative_sum(end)` to verify that the length of a pixel array matches the sum of the elements in an `npix` array. Perhaps the cleanest solution is to remove the `cumulative_sum` argument from `split_vector_` and just use a `sum` to perform validation in the calling functions (e.g. `do_binary_op_sigvar_`).

_test/test_utilities/test_split_vector_max_sum.m
hsautessella: Is `test_vector_value_gt_max_chunk_sum_comprises_its_own_chunk` a better name? It describes the test's purpose, but maybe doesn't describe that it's actually verifying that the rest of the chunking is "correct" as well.

horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -0,0 +1,104 @@
...
+    chunks = {};
+    idxs = zeros(2, 0);
+    return;
+end
+
+validateattributes(numeric_vector, {'numeric'}, {'vector', 'nonnegative'});
+validateattributes(chunk_sum, {'numeric'}, {'scalar', 'positive'});
+numeric_vector = make_row(numeric_vector);
+
+if ~exist('cumulative_sum', 'var')
```
nickbattam-tessella, line 50: The optional `cumulative_sum` argument has been removed so this `if` will always be true and so can be dropped

nickbattam-tessella (approved):

PR #624, opened 2021-03-12
Implement file-backed sqw_eval

hsautessella:
~Some of the commits on this branch are also on https://github.com/pace-neutrons/Horace/pull/623. I'll rebase once https://github.com/pace-neutrons/Horace/pull/623 is merged.~

~There appears to be a failing test when not using mex (I believe it's a problem with `compute_bin_data`), so I've marked this as a draft for now, until the issue is resolved.~

Adresses: #550 

~**Edit:**~
~_Any comments regarding the `split_npix_max_sum` function should be made in PR #623. That PR will have the most up-to-date version of the function. This PR will be rebased on top of #623 when #623 is merged._~

- I've updated `sqw_eval` such that it can work with file-backed pixels.
- I've added a `load_sqw_dnd` method, which takes in an array of sqw/dnd objects or a cell array of sqw/dnd files and returns loaded objects.
    - The logic from this function was previously present in `algorithms/func_eval`
    - The key part of this function is that it can return objects with file-backed pixels
    - I expect that all our file-backed algorithms will use this function to parse its inputted data source(s)
- I've added a `put_image` method to `dnd_binfile_common`.
    - This was useful for the file-backed implementation of `sqw_eval` as, to minimize the number of loops over pixels, it was required that a "placeholder" image be written to the file and then overwritten once the new image was calculated.
- I've added the function `gen_unique_file_paths`
    - Similar logic as is in this function was already present in `cut` and in `func_eval`
    - This function will generate a random unique path (using  UUID) within `tmp_dir()`
    - I've used a Java function to generate the UUID, is this OK?

Currently, we're performing two loops over pixels in `do_sqw_eval_average_filebacked`, one for applying the input function to the pixels, one for recomputing the bin/image data. It should be possible to perform both these operations in a single loop. If time allows, I'm happy to revisit this, but I think that should probably be a new PR - this one is big enough already.


hsautessella:
Blocked waiting on merge of #623

horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -0,0 +1,113 @@
...
+% sub-vector's sum is exactly 'chunk_sum'. If chunk sum does not divide the sum
+% of the input vector, the final chunk's sum will be the remainder.
+%
+% Values in the input vector may be split between chunks.
+%
+% Input:
+% ------
+% numeric_vector   A vector of numeric, non-negative, values.
+% chunk_sum        A positive value specifying the sum for each sub-vector.
+% cumulative_sum   The cumulative sum of 'numeric_vector'. (optional)
```
cmarooney-stfc, line 12: This is an output in the other split_vector function. It may be simpler to calculate it locally in the same way rather than rely on an optional input which may not have been correctly set up

horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -0,0 +1,113 @@
...
+    cumulative_sum = cumulative_sum - chunk_sum;
+    end_idx = (start_idx - 1) + find(cumulative_sum(start_idx:end) > 0, 1);
+
+    if isempty(end_idx)
+        % No end index found, so we must be at the end of the vector
+        end_idx = numel(numeric_vector);
+    end
+
+    if start_idx == end_idx
+        % All vector values in single value for this chunk
```
cmarooney-stfc, line 87: I think this means, all the values for this chunk will be in the single current item in numeric_vector, but I cannot make any sense of this comment, I just work out what the start_idx==end_idx condition is. There is no comment to identify leftover_end at this point; it would be good to say what it is now rather than waiting for further use lower down

horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -0,0 +1,113 @@
...
+        if ~exist('leftover_end', 'var')
+            leftover_end = 0;
+        end
+        chunks{chunk_num} = min(chunk_sum, numeric_vector(start_idx) - leftover_end);
+    else
+        % Add an empty element to the end, which will be used to hold a part of
+        % the end index's value. If there's no remainder on the end index's
+        % value, the empty element will be removed
```
cmarooney-stfc, line 95: I am not clear why reshape is required. A quick experiment at the comment line seems to say that [begin, nv[start+1,end-1],0] produces a 1xn vector fine

cmarooney-stfc (changes requested):
Partial review while I wrestle with the algorithm

horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -0,0 +1,113 @@
...
+    chunk_sum = min(chunk_sum, remaining_sum);
+
+    if chunk_num == 1
+        % For the first iteration, we need >= to catch leading zeros
+        % This index may contribute its full value, or only part of its value
+        start_idx = (end_idx - 1) + find(cumulative_sum(end_idx:end) >= 0, 1);
+    else
+        % We ignore leading zeros on subsequent iterations as they will have
+        % been assigned to the previous chunk
+        start_idx = (end_idx - 1) + find(cumulative_sum(end_idx:end) > 0, 1);
```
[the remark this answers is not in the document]
hsautessella: Sorry, I should have been more explicit in the PR description, but there's a separate PR open for this function which may be slightly ahead of this version. #623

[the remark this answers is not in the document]
hsautessella: A lot of these comment still apply though, but I'll make the updates to 623

horace_core/utilities/split_vector_fixed_sum.m
hsautessella: The reason for having it as an input argument was because in functions like [`binary_op_double_`](https://github.com/pace-neutrons/Horace/blob/4967048545f7bc12862c25c253f1d7be0ab6442a/horace_core/sqw/PixelData/%40PixelData/private/binary_op_double_.m#L92) we are already calculating the (cumulative) sum for argument validation. So it's more efficient to pass that in, rather than calculate it again.

cmarooney-stfc (commented):
Review part #2, noting you have said there is another  PR slightly ahead of this version

horace_core/utilities/split_vector_fixed_sum.m
```diff
@@ -0,0 +1,113 @@
...
+        % value, the empty element will be removed
+        chunk = [ ...
+            leftover_begin, ...
+            reshape(numeric_vector(start_idx + 1:end_idx - 1), 1, []), ...
+            0 ...
+        ];
+
+        leftover_end = chunk_sum - sum(chunk);
+        if leftover_end == 0 && chunk_num ~= num_chunks
+            chunk(end) = [];
```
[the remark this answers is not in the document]
hsautessella: To delete elements in arrays (also works for cell arrays), you assign an empty value to it. This is equivalent to
```matlab
chunk = chunk(1:end - 1);
```

It's much cheaper to allocate one extra element to the chunk and remove it, than it is to append an element (which results in re-allocating the whole array).

horace_core/utilities/split_vector_fixed_sum.m
hsautessella: The `reshape` is important, otherwise the function will fail if the input is a column vector.
```matlab
>> row_vec = [1, 2, 3, 4];
>> col_vec = row_vec';
>> size(row_vec(1:end - 1))
ans =

     1     3

size(col_vec(1:end - 1))
ans =

     3     1
```

hsautessella: Another (probably cleaner) solution is to reshape the whole vector at the start of the function. I'll do that instead

horace_core/sqw/file_io/load_sqw_dnd.m
```diff
@@ -0,0 +1,112 @@
...
+                    sqw_dnd_obj = ldr.get_sqw('pixel_page_size', pixel_page_size);
+                else
+                    sqw_dnd_obj = ldr.get_sqw();
+                end
+            else
+                % In contrast to the above case, we can use the loader to get the dnd
+                % as no extra constructor arguments are required.
+                sqw_dnd_obj = ldr.get_dnd(source);
+            end
+        catch ME
```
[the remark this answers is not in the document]
hsautessella: We're not _expecting_ an error from anywhere here. This is essentially a `try`-`finally` block that makes sure that, if there's an error reading the file, the file is closed.

horace_core/sqw/@sqw/sqw_eval.m
```diff
@@ -1,102 +1,348 @@
...
-if ~ok
-    error('SQW_EVAL:invalid_argument',mess);
+%   wout        If `filebacked` is false, an sqw object or array of sqw objects.
+%               If `filebacked` is true, a file path or cell array of file paths.
+%               Output argument must be specified if `outfile` not given.
+%
+[sqwfunc, pars, opts] = parse_arguments(win, sqwfunc, pars, varargin{:});
+if isempty(opts.outfile) || opts.filebacked
+    nargoutchk(1, 1);
 end
```
cmarooney-stfc, line 81: I'm not clear about this. I can't even formulate a comment. Guessing we're ensuring that wout has been specifed at the call (a) in the case that it is filebacked, so wout is an obj/array of file paths (b) in the case that outfile has not been specified, in which case we must have specified some form of wout to receive the result. What is the alternative case? Under what circumstances do we not need wout?

horace_core/sqw/@sqw/sqw_eval.m
```diff
@@ -1,102 +1,348 @@
...
 wout = copy(win);
-if ~iscell(pars), pars={pars}; end  % package parameters as a cell for convenience
+if ~iscell(pars)
+    pars = {pars};  % package parameters as a cell for convenience
+end
+
+for i = 1:numel(win)
 
-for i=1:numel(win)
     if is_sqw_type(win(i))   % determine if sqw or dnd type
```
cmarooney-stfc, line 90: Checking if we should say `has_pixels` - may have to await Nick's new sqw

horace_core/sqw/@sqw/disp2sqw_eval.m
```diff
@@ -78,25 +78,30 @@
 all_bins=false;
 ave_pix=false;
 if exist('opt','var')  % no option given
-    if ischar(opt) && ~isempty(strmatch(lower(opt),'all'))    % option 'all' given
+    if ischar(opt) && startsWith(opt,'-all', 'IgnoreCase', true)
```
cmarooney-stfc, line 81: checking comment, looks as if option is given

horace_core/utilities/split_vector_fixed_sum.m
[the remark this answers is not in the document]
hsautessella: Matlab doesn't allow "nested" arrays in this sense; it has multidimensional arrays. Elements in arrays cannot be empty.

You can store an empty array within a _cell_ array using curly braces `{ }` (which provide element access):
```matlab
>> c = cell(1, 2);
>> size(c)

ans =

     1     2

>> c{2} = [];
>> size(c)

ans =

     1     2

>> c{2}

ans =

     []
```

If you use smooth braces `( )` (which provide slice access), the similar syntax will delete the element:
```matlab
>> c = cell(1, 2);
>> size(c)

ans =

     1     2

>> c(2) = [];
>> size(c)

ans =

     1     1
```

horace_core/sqw/@sqw/sqw_eval.m
hsautessella: `has_pixels` doesn't exist on `master`, so we'd probably need to wait for Nick.

horace_core/sqw/@sqw/sqw_eval.m
```diff
@@ -1,102 +1,348 @@
...
 %              full sqw object.
 %
-%   'average' Requests that the calculated sqw be computed for the
-%              average values of h,k,l of the pixels in a bin, not for each
+%   average    Requests that the calculated sqw be computed for the
+%              average values of h, k, l of the pixels in a bin, not for each
 %              pixel individually. Reduces cost of expensive calculations.
-%              Applies only to the case of sqw object with pixel information 
+%              Applies only to the case of sqw object with pixel information
 %             - it is ignored if dnd type object.
```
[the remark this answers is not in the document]
hsautessella: I've added that we expect logical arguments. 

In [ADR 22](https://github.com/pace-neutrons/Horace/blob/master/documentation/adr/0022-use-keyword-args.md) we decided that we would use `-flags` and boolean `'name', value` pairs interchangeably. The usage example at the top of the help should demonstrate this.

Can I suggest we open an issue to update the `dnd` function helps separately? I know that Nick has re-factored the `dnd` implementation into the base class, so it would make more sense to do it there.

horace_core/sqw/file_io/@dnd_binfile_common/put_image.m
```diff
@@ -0,0 +1,36 @@
...
+bytes_in_file = obj.e_pos_ - obj.s_pos_;
+
+if bytes_to_write ~= bytes_in_file
+    error( ...
+        'HORACE:DND_BINFILE_COMMON:invalid_length', ...
+        ['Image signal size in file does not match the size of the input.\n' ...
+         '''%i'' elements in file, ''%i'' elements in input array.'], ...
+        bytes_in_file/BYTES_IN_SINGLE, numel(bytes_to_write) ...
+    );
+end
```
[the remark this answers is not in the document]
hsautessella: It can be overwriting an existing file, or writing a new one. When you initialize a loader you initialize with an object and a file path (existing or new). I think `obj.s_pos_` and `obj.e_pos_` are derived from the object, and hence so is the allowed image array size.

I've tried to clear this up in the function help

_test/common_functions/gen_tmp_file_path.m
```diff
@@ -0,0 +1,18 @@
+function tmp_file_path = gen_tmp_file_path(suffix)
```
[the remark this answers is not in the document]
hsautessella: The functions do different things. `gen_tmp_file_path` is a test utility, the generated path contains the name of the calling function, which makes it easy to identify exactly where the file came from.

`gen_unique_file_paths` focuses on being random (using UUIDs), with a counter, so you can generate sets of file paths. It is a public `horace_core` function.

horace_core/sqw/file_io/load_sqw_dnd.m
```diff
@@ -0,0 +1,113 @@
+function [data, unmatched_args] = load_sqw_dnd(sources, varargin)
```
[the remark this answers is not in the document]
hsautessella: It _uses_ the existing loaders. I'm not sure if the name of the function here is appropriate. The purpose of the function is to parse (and load) data sources passed to Horace algorithms.

It takes files/objects (or combinations thereof) as input and spits out corresponding objects (whose pixels may be file-backed). The idea is that all Horace algorithms that can run file-backed have a top-level overload, these overloads should use this function to parse/load their data source.

_test/test_utilities/test_gen_unique_file_paths.m
```diff
@@ -0,0 +1,108 @@
...
+            prefix = 'MY_PREFIX';
+            nfiles = 2;
+            paths = gen_unique_file_paths(nfiles, prefix, '');
+
+            for i = 1:numel(paths)
+                assertTrue(endsWith(paths{i}, 'tmp'));
+            end
+        end
+
+        function test_separate_calls_generate_different_UUIDs(~)
```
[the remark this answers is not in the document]
hsautessella: I am testing that UUIDs are different. I'm extracting them with a regex, then comparing them.

A quick test suggest to me that, yes, UUIDs will be different when the function is called in a `parfor` loop:

```matlab
parfor i = 1:8
    uuid = char(java.util.UUID.randomUUID());
    f = fopen(num2str(i), 'w');
    fprintf(f, uuid);
end
```
produces 8 files that all contain different UUIDs

_test/test_sqw/test_sqw_dnd_eval.m
```diff
@@ -207,7 +207,7 @@ function test_sqw_eval_aver(obj)
             %
             err_message = '';
             try
-                ds = sqw_eval(obj.sqw_4_test,@test_sqw_dnd_eval.sqw_eval_tester,[],'av');
+                ds = sqw_eval(obj.sqw_4_test,@test_sqw_dnd_eval.sqw_eval_tester,[],'-av');
```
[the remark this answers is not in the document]
hsautessella: From the `@sqw/sqw_eval` help:

```matlab
% Note: all optional string input parameters can be truncated up to minimal
%       difference between them e.g. routine would accept 'al' and
%       'av', 'ave', 'aver' etc....
```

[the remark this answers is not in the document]
nickbattam-tessella: It's totally valid, but I'm not sure I see any benefit in truncating options in "code" (vs interactive terminal session) unless it adds clarity.

_test/test_utilities/test_gen_unique_file_paths.m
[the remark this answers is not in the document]
nickbattam-tessella: It's a question about the test name -- the requirement on "get_unique_file_paths" is that it gives you unique paths, not that it has unique uuids.

_test/test_sqw/test_sqw_dnd_eval.m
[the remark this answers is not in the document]
oerc0122: I agree with nick here, in code, I'd always prefer the verbose option. It keeps all the information in one place without having to dive into another file's help.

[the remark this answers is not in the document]
nickbattam-tessella: `disp2sqw_eval` uses `-ave`.

horace_core/sqw/file_io/load_sqw_dnd.m
```diff
@@ -0,0 +1,113 @@
...
+%LOAD_SQW_DND attempt to load the given sources into sqw/dnd objects
+%
+% Input:
+% ------
+% sources       Can be:
+%                 - sqw object (scalar, array or cell array)
+%                 - dnd object (scalar, array or cell array)
+%                 - char array (path to an sqw/dnd file)
+%                 - cellstr (cell array of sqw/dnd file paths)
+%               You cannot mix types within cell arrays (e.g. a cell array of
```
nickbattam-tessella, line 11: Is this restriction necessary for the function use-cases or a result of the structure of the return values?
Returning array of object limits you with the current sqw/dnd old-style classes, but the new SQW objects would all fit in an array of SQWDnDBase.

horace_core/sqw/file_io/load_sqw_dnd.m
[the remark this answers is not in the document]
nickbattam-tessella: that's not obvious from the function name, but the `sources` argument description in the function comment is clear.

horace_core/sqw/@sqw/private/cut_accumulate_data_.m
```diff
@@ -73,11 +73,22 @@
 end
 
 for iter = 1:num_chunks
+    sub_bin_starts = bin_starts(sub_bin_idxs(1, iter):sub_bin_idxs(2, iter));
+    sub_bin_ends = bin_ends(sub_bin_idxs(1, iter):sub_bin_idxs(2, iter));
+
+    num_pix_in_chunk = sum(sub_bin_ends - sub_bin_starts);
+    if num_pix_in_chunk > block_size
+        warning( ...
```
nickbattam-tessella, line 81: do we want one warning per too big chunk or one warning that _these_ chunks (maybe multiple) are too big.

I'd prefer to see a single warning

horace_core/sqw/file_io/load_sqw_dnd.m
nickbattam-tessella: Update: SQWDnDBase does not extend the Heterogeneous mixin so it's not possible to create a mixed array of DnD and SQW class objects.
The restriction in the comments will remain valid

cmarooney-stfc (approved):

nickbattam-tessella (approved):
Approved subject to the Windows builds completing successfully

admin/validate_horace.m
```diff
@@ -114,6 +114,10 @@
     hec.log_level = -1; % turn off informational output
 end
 
+hec
```
oerc0122, line 117: Short comment on why these are here? Use `disp` to explicitly print (Personally, I find just missing semi-colon non-obvious given half a dozen languages have optional semi-colons)

horace_core/sqw/@sqw/disp2sqw_eval.m
```diff
@@ -78,25 +78,30 @@
 all_bins=false;
 ave_pix=false;
 if exist('opt','var')  % no option given
```
oerc0122, line 80: Do we want to use `parse_char_options` here?

horace_core/sqw/@sqw/disp2sqw_eval.m
oerc0122, line 81: This would detect the option:
`-allTheseOptionsAreStupid`, do we want this behaviour?

oerc0122 (approved):

horace_core/sqw/@sqw/disp2sqw_eval.m
oerc0122: Would strcmpi be better?

nickbattam-tessella: it's `startswith` , but the other way i.e. `-a`, `-al` and `-all` should match, but not `-allTheseOptionsAreStupid`

admin/validate_horace.m
hsautessella: These are just here temporarily for debugging purposes. The builders keep timing out on different tests, and I'm trying to replicate the errors. No luck so far though.

I could add `disp` if we think it's useful to print the config at the start of each test folder run?

Issue #625, opened 2021-03-15
Mex/Matlab implementations of compute_bin_data returning different output


hsautessella:
`PixelData.compute_bin_data` is returning different values depending on whether `hor_config.use_mex` is set to `true` or `false`. 

There are tests in `test_PixelData_operations` that check that `compute_bin_data` return the same thing with mex on and off (for 2 different datasets), but perhaps there are cases these tests do not cover.

I'll try and add more detail to the issue once I've investigated further.

Tests to show the behaviour are below (note that the ability to run these tests depends on the new `sqw_eval` API written in #624). The test with mex on passes, the test with mex off does not. I stepped through the debugger, and it seems as though the pixel data is equivalent when mex is on and when it is off, the discrepancy comes in the image signal, which suggest that `compute_bin_data` is not doing its job.

```matlab

classdef test_sqw_eval < TestCase

    properties (Constant)
        FLOAT_TOL = 1e-5;
    end

    properties
        sqw_2d_obj;
        sqw_2d_file_path = '../test_sqw_file/sqw_2d_1.sqw';
        sqw_2d_sqw_eval_ref_obj;
        sqw_2d_sqw_eval_ref_file = 'test_sqw_eval_gauss_ref.sqw';
        sqw_2d_pix_pg_size = 3e5; % Gives us 6 pages

        gauss_sqw;
        gauss_params;
    end

    methods

        function obj = test_sqw_eval(~)
            obj = obj@TestCase('test_sqw_eval');

            % Sum of the gaussian of each coordinate
            obj.gauss_sqw = ...
                @(u1, u2, u3, dE, pars) ...
                    sum(arrayfun(@(x) gauss(x, pars), [u1, u2, u3, dE]), 2);
            obj.gauss_params = [10, 0.1, 0.05];

            obj.sqw_2d_obj = sqw(obj.sqw_2d_file_path);
            obj.sqw_2d_sqw_eval_ref_obj = sqw(obj.sqw_2d_sqw_eval_ref_file);
        end

        function test_with_filebacked_pix_matches_reference_object_with_mex(obj)
            conf_cleanup = set_temporary_config_options( ...
                hor_config, 'pixel_page_size', obj.sqw_2d_pix_pg_size, ...
                'use_mex', true ...
            );

            out_sqw = sqw_eval( ...
                obj.sqw_2d_file_path, obj.gauss_sqw, obj.gauss_params, ...
                'filebacked', true ...
            );

            assertEqualToTol( ...
[...]

PR #627, opened 2021-03-17
Mex and Matlab implementations of compute_bin_data returning different output

hsautessella:
In this PR I've re-written and (hopefully) simplified `compute_bin_data_matlab_`. I've also extracted all the `compute_bin_data` tests out of `test_PixelData_operations` and put them in a new file: `test_PixelData_compute_bin_data`.

It was hard to pin down the exact problem that was causing the test failure described in #625. This meant I wasn't able to replicate the failure in a simpler version of the test given in #625. But with this change, the existing `compute_bin_data` tests pass, and the test on the branch for #624 now passes.

In the interest of getting #624 done, I didn't want to spend any more time on this.

Addresses #625


nickbattam-tessella (commented):
It's not obvious whether the test case data covers the handling of the NaNs or not, otherwise looks good

nickbattam-tessella (approved):
Can you add the before/after performance data you collected to the ticket to record this refactor hasn't slowed it

_test/test_sqw/test_PixelData_operations.m
```diff
@@ -4,195 +4,14 @@
...
-        
-        old_warn_state;
     end
     
     methods
         
         function obj = test_PixelData_operations(~)
             obj = obj@TestCase('test_PixelData_operations');
-            
-            addpath(fullfile(obj.this_dir, 'utils'));
```
cmarooney-stfc, line 36: Rather than cause a surprise if anyone looks at the file history, it may be worth saying where the tests were moved to.

cmarooney-stfc (approved):

hsautessella:
These benchmarks were taken using the Iron data that is part of the Horace tutorial. Pixels were all in memory, not file-backed. It has ~10 GB worth of pixels (as float64). The `benchmark_function` function can be found in `Horace/_test/common_functions`. 

On `master`:
```matlab
>> s = sqw('.\iron_data.sqw');
>> pix = s.data.pix; npix = s.data.npix;
>> set(hor_config, 'use_mex', false);
>> benchmark_function(@() pix.compute_bin_data(npix), 2, 3)

ans =

    9.0164    9.0993    9.0971

```

On `625_mex_matlab_implementations_of_compute_bin_data_returning_different_output`:
```matlab
>> s = sqw('.\iron_data.sqw');
>> pix = s.data.pix; npix = s.data.npix;
>> set(hor_config, 'use_mex', false);
>> benchmark_function(@() pix.compute_bin_data(npix), 2, 3)

ans =

    3.2785    3.2189    3.2682

```

oerc0122:
So what we thought was optimisation slowed it down 300%?

_test/test_sqw/test_PixelData_operations.m
hsautessella: I feel as though if someone's looking at the file history, then the commit message ("_Refactor compute_bin_data tests into own file_"), and the full commit diff, will make clear what's happened.

cmarooney-stfc: I would note that reviewing any PR through the file changes does not give the commit message history. It should be easy to review without having to second-guess what the developer did and go looking for other access points to the reasoning.

hsautessella: I'd argue this is the job of the PR description, not a comment within code. If the purpose of a comment is to guide a reviewer, it becomes redundant as soon as the PR is merged.

Issue #618, opened 2021-03-10
Write SMG for using PixelData


hsautessella:
Write some documentation for the `PixelData` API, and how to use it. This should include some code examples.

There's a question of the target audience for this documentation: developers or users? We may want a document for both.

PR #630, opened 2021-03-23
Write SMG for using PixelData

hsautessella:
I've written an SMG for using the `PixelData` class. I've written some context on the problem it aims to address, as well as some principles behind the object. There are also a couple of examples on how to implement file-backed algorithms, and a few notes on things to be aware of.

Addresses #618 

_Note: I expect there will be some additions following our meeting tomorrow (24/03/21)_

### DIFF
--- a/documentation/smg/17_pixel_data.md
+++ b/documentation/smg/17_pixel_data.md
@@ -1,4 +1,4 @@
-# Using Pixel Data
+# Pixel Data
 
 ## The Problem to Address
 
--- a/documentation/smg/17_pixel_data.md
+++ b/documentation/smg/17_pixel_data.md
@@ -406,3 +406,7 @@ end
 - Algorithms that transform pixel coordinates should update the **pixel range**.
   The pixel range is cached in file and on the object.
   This specifies the minimum and maximum value for each coordinate.
+
+- Pixel data values are stored in single-precision in file.
+  They are store in double-precision in memory.
+  See [ADR 15](../adr/0015-store-pixel-data-in-single-precision.md)
--- a/documentation/smg/17_pixel_data.md
+++ b/documentation/smg/17_pixel_data.md
@@ -188,7 +188,7 @@ column.
 
 All properties use [_relative indexing_](#Absolute-and-Relative-Indexing).
 
-### Getters/Methods
+### Methods
 
 Getter and setter methods are also provided.
 
--- a/documentation/smg/17_pixel_data.md
+++ b/documentation/smg/17_pixel_data.md
@@ -167,9 +167,13 @@ There also exists `sqw_binfile_common.put_bytes`
 which can be used to save arbitrary numeric data to an `sqw` file.
 You can position the `sqw_binfile_common` instance's file handle at the start
 of the `sqw` file's pixel array,
-and save pixels directory to file this way.
-This is more efficient than saving out to temporary files and then
-reading them back in later when calling `put_pix`.
+and save pixels directly to file this way.
+
+This can be useful if you are performing an action on an sqw object's pixels,
+and want to output directly to an sqw file,
+rather than dealing with temporary files.
+See the example on [avoiding temporary files](#avoiding-temporary-files)
+for an example of how and where this is useful.
 
 ### Getters/Setters
 
@@ -345,6 +349,63 @@ classdef PixelData < handle
     end
 end
 ```
+
+### Avoiding Temporary Files
+
+It can sometimes be possible to avoid using temporary files when operating on pixels.
+
+Take, the above [`minus` example](#simple-example) for the case of an sqw object.
+If our desired output is not a file-backed `PixelData` object,
+but a new SQW file, we can write the updated pixels directly to the output file,
+instead of first storing them in temporary files and then re-combining on save.
+
+```matlab
+function minus(sqw_obj, scalar, outfile_path)
+    %MINUS Subtract a scalar from every pixel in the input sqw object,
+    % and save the output sqw object to `outfile_path`
+    %
+    % The input sqw object is not to be modified.
+    %
+
+    % Initialize an faccess object for our output file
+    faccess_out = sqw_formats_factory.instance().get_pref_access(sqw_obj);
+    faccess_out = faccess_out.init(sqw_obj, outfile_path);
+
+    % Write everything up to pixels
+    % This will also move the file handle to the position of the start of the
+    % pixel array
+    faccess_out.put_sqw('-nopix');
+
+    % Get reference to PixelData object, note no copying as PixelData is a handle
+    pix = sqw_obj.data.pix;
+
+    % Make sure we're on the first page before beginning the loop
+    pix.move_to_first_page();
+
+    % Make the subtraction from the pixel's signal
+    pix.signal = pix.signal - scalar;
+    % Write the edited pixel cache directly to the output file
+    faccess_out.put_bytes(pix.data);
+
+    % Start loop over subsequent pixels
+    while pix.has_more()
+        % The `nosave` flag means changes to the cache are discarded on advance,
+        % hence no temporary files are written.
+        pix.advance('nosave', true);
+
+        pix.signal = pix.signal - scalar;
+        faccess_out.put_bytes(pix.data);
+    end
+end
+% Now all pixels are saved, file footers must be written
+faccess_out.put_footers();
+```
+
+The alternative to the above is performing two loops over the pixel data.
+The first loop would be making the subtraction and writing the cached changes
+to temporary files (on `pix.advance()`).
+The second loop would be reading in those temporary files,
+and then writing them to the output file.
 
 ### Managing Binning
 
--- a/documentation/smg/17_pixel_data.md
+++ b/documentation/smg/17_pixel_data.md
@@ -381,7 +381,7 @@ classdef PixelData < handle
                 % number of elements as we have pixels in the current page.
                 signal_to_add = repelem(img_chunk(:), npix_chunks{page_number});
 
-                % Increment the pixel signal and move the next page
+                % Increment the current page's signal and move to the next page
                 obj.signal = obj.signal + signal_to_add;
                 obj.advance()
             end
--- a/documentation/smg/17_pixel_data.md
+++ b/documentation/smg/17_pixel_data.md
@@ -419,8 +419,12 @@ end
   (e.g. signal), as these data are not contiguous in memory (or on disk).
 
 - Algorithms that transform pixel coordinates should update the **pixel range**.
+  This range is a property on the PixelData instances (`pix_range`),
+  it should be updated with the new coordinate limits.
+  Preferably, the update should take place within the algorithm.
+  This avoids the use of the `recalc_pix_range` method,
+  which performs a loop over pixels.
   The pixel range is cached in file and on the object.
-  This specifies the minimum and maximum value for each coordinate.
 
 - Pixel data values are stored in single-precision in file.
   They are store in double-precision in memory.
--- a/documentation/smg/17_pixel_data.md
+++ b/documentation/smg/17_pixel_data.md
@@ -164,7 +164,7 @@ This method, called when saving an `sqw` object,
 loops over pixel pages and saves them to the `sqw` file.
 
 There also exists `sqw_binfile_common.put_bytes`
-which can be used to save arbitrary data to an `sqw` file.
+which can be used to save arbitrary numeric data to an `sqw` file.
 You can position the `sqw_binfile_common` instance's file handle at the start
 of the `sqw` file's pixel array,
 and save pixels directory to file this way.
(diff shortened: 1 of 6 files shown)


documentation/smg/17_using_pixel_data.md
```diff
@@ -0,0 +1,394 @@
+# Using Pixel Data
```
oerc0122, line 1: I would probably say this is more "Pixel Data" than just using it. Especially given you start with a definition of why it's needed.

documentation/smg/17_using_pixel_data.md
```diff
@@ -0,0 +1,394 @@
...
+
+## The Problem to Address
```
oerc0122, line 3: Problems?

documentation/smg/17_using_pixel_data.md
```diff
@@ -0,0 +1,394 @@
...
+
+As pixel data is often large, it's advantageous to perform operations on the
+data using distributed memory parallelism.
+To aid in this, a simple way of "chunking" the data between workers was required.
+
+#### Parallelism - Solution
+
+The ability to move to specific pages of pixel data was proposed to solve this
+problem.
+Using the `move_to_page(page_num)` method allows a worker to position itself at
```
oerc0122, line 81: Is there a method to set page sizes such that each page is of such a size the data can be evenly distributed between N workers or similar? Might save some work
```matlab
pixData.page_size = pixData.distribute(nWorkers, maxPage)
```
Such that `page_size` will be some multiple of `nWorkers` and no bigger than `maxPage`

documentation/smg/17_using_pixel_data.md
```diff
@@ -0,0 +1,394 @@
...
+  - Changes are _never_ written to the SQW file used in construction.
+  - A logical array is held on the object that tracks which pages are "dirty".
+  - If a page is "clean" the data is loaded from the SQW file, if the page is
+    "dirty", it is loaded from a temporary file.
+  - Each page has its own temporary file.
+
+- It is a
+  [`handle` class](https://www.mathworks.com/help/matlab/matlab_oop/handle-objects.html).
+  - To copy a `PixelData` object, you must explicitly call the `copy` method.
+    This copy can be expensive, as all temporary files are also copied.
```
oerc0122, line 110: Are they kept as temporaries or are they made part of the core? Does it still refer to the original SQW?

documentation/smg/17_using_pixel_data.md
```diff
@@ -0,0 +1,394 @@
...
+  - There is no "lazy-copy-on-assignment" as exists for Matlab value classes.
```
oerc0122, line 111: This should in principle be possible using a doubly dirty notation? Copy the dirties as 1, if they're subsequently dirtied, they're set to 2 and new tmps are created. On copy set all 2s to 1.

i.e. 0 -> SQW, 1 -> Temps, 2-> MyTemps, but then I guess when you chain copy, this gets awkward?

documentation/smg/17_using_pixel_data.md
```diff
@@ -0,0 +1,394 @@
...
+  where pixels are stored as `single`.
+  See [ADR 15](../adr/0015-store-pixel-data-in-single-precision.md).
+
+- `PixelData` objects do not know about their binning.
+  - The binning is defined by the `sqw` object's image.
+
+### Getters/Setters
+
+All getters return data in-memory.
+All setters must set data that is in memory.
```
oerc0122, line 155: Is it an error to fail to do so?

documentation/smg/17_using_pixel_data.md
```diff
@@ -0,0 +1,394 @@
...
+
+All these properties work by slicing into the cached data array at the correct
+column.
+
+All properties use [_relative indexing_](#Absolute-and-Relative-Indexing).
+
+### Getters/Methods
+
+Getter and setter methods are also provided.
+You cn
```
oerc0122, line 190: can

documentation/smg/17_using_pixel_data.md
```diff
@@ -0,0 +1,394 @@
...
+whether that data is in memory or is in a file.
+They also have the advantage of being able to get/set multiple columns at once.
+This can be more efficient, as you're performing a single slice, instead of
+multiple.
+
+```matlab
+% Below examples are equivalent, but using the get_data method is more
+% efficient as it performs fewer slices
+
+% Performs one slice over the pixel array
```
oerc0122, line 222: Provided they are contiguous

documentation/smg/17_using_pixel_data.md
```diff
@@ -0,0 +1,394 @@
...
+            % Subtract scalar from every value in cache.
+            % If there is no data in the cache, this call will load that data
+            % from file and then perform the operation.
+            obj.signal = obj.signal - scalar;
+
+            % Check if there are any more pixels to load from file.
+            while obj.has_more()
+                % Move to the next page of data.
+                % This call dumps the previous page to a temporary file and
+                % marks the page as "dirty".
```
oerc0122, line 299: Does it always mark it as dirty, even if no changes have taken place? Is it possible to set the dirty marking to any setters/getters?

documentation/smg/17_using_pixel_data.md
```diff
@@ -0,0 +1,394 @@
...
+```
+
+### Managing Binning
+
+Some algorithms require that pixels in different bins are operated on in
+different ways.
+In most cases, bins will cross pixel page boundaries,
+so it takes some work to find which bins require which action.
+
+As an example, consider performing addition between an `sqw` and `dnd` object.
```
oerc0122, line 323: a `dnd` object

documentation/smg/17_using_pixel_data.md
hsautessella: No such method as yet. Changing page size once the object already exists is quite fiddly; what happens if there are dirty pages stored in temp files? For the time being, page size can only be set on construction.

oerc0122: I also realised that if you were keen on doing this, you could set it via `split_vector_into_chunks` (or whatever it's called)

documentation/smg/17_using_pixel_data.md
hsautessella: We did consider this, but thought that it would get a bit messy. Copying edited file-backed objects is not something we have done anywhere at the moment. If the need arises, we could look at optimizing this, but it may be premature to do so at this point.

documentation/smg/17_using_pixel_data.md
hsautessella: The values you set must be numeric, so it would be an error to pass any of the setters anything other than an in-memory array.

Although, no `set_pixels` method exists at the moment (`set_data` was only recently added), but that could in theory set pixel data from a file-backed `PixelData` object.

Should I remove the setters part of this in case a `set_pixels` method is ever required?

oerc0122: I think the getters bit is clarified below but useful. I just think it would be good to note what happens if you attempt to set pixels out of current page, or set pixels out of range? If it's an error, note "Anything else will result in an error"

documentation/smg/17_using_pixel_data.md
hsautessella: No, pages are only marked dirty if they're edited.

hsautessella: Not entirely sure what you mean by the second question

oerc0122: It looked like you were saying dirty was set on calling `advance`. If you set "dirty" on anything that modifies pixels (presumably the pixels' setter) then that would avoid that.

oerc0122 (commented):
Looks good, but I'll wait until we've had the meeting tomorrow to approve.

documentation/smg/17_pixel_data.md
```diff
@@ -0,0 +1,397 @@
...
+            [npix_chunks, img_idxs] = split_vector_fixed_sum(npix, obj.base_page_size);
+
+            for page_number = 1:numel(npix_chunks)
+                % Get the signal at the bins for which the currently loaded
+                % page of pixels contribute
+                img_chunk = sigvar_obj.s(img_idxs(1, page_number):img_idxs(2, page_number);
+
+                % "Smear" the sigvar image's signal, such that it has the same
+                % number of elements as we have pixels in the current page.
+                signal_to_add = repelem(img_chunk(:), npix_chunks{page_number});
```
cmarooney-stfc, line 371: Sorry, but I do not understand this. Vaguely it appears you are getting img_chunk out of sigvar_obj.s as the values for the current page. I do not understand why this is not added directly to obj.signal. I do not understand why img_chunk needs to be put through repelem. I do not understand what smearing means. It would be good if you could provide more explicit explanation here. Thanks

hsautessella: I've added more detail to the problem we're trying to solve in the function header. Hopefully that makes it more understandable, but let me know if it needs further clarification.

documentation/smg/17_pixel_data.md
```diff
@@ -0,0 +1,427 @@
...
+>> pix.num_pixels
+     ans =
+       100
+>> pix.move_to_page(3);
+>> all(pix.signal(1:6) == pix.get_data('signal', 21:26))
+     ans =
+       true
+```
+
+There exists a private utility function that can perform the conversion between
```
nickbattam-tessella, line 265: which function?

documentation/smg/17_pixel_data.md
```diff
@@ -0,0 +1,427 @@
...
+## Things to Be Aware Of
+
+- The underlying pixel data is stored in a 9xN array
+  (where N is number of pixels).
+  This means **it is efficient to slice full pixels** from the data,
+  as data points for each pixel are contiguous in memory.
+  **It is less efficient to extract data for pixel fields**
+  (e.g. signal), as these data are not contiguous in memory (or on disk).
+
+- Algorithms that transform pixel coordinates should update the **pixel range**.
```
nickbattam-tessella, line 421: how? is there a method to do that update?

documentation/smg/17_pixel_data.md
```diff
@@ -0,0 +1,427 @@
...
+the number of workers, the number of pixel, the page size.
+Each worker can receive its start and end page index from the head node,
+and operate on those pixels.
+
+## The PixelData Class
+
+### Principles
+
+- It operates with "pages" of cached data.
+  - The page size is specified on construction (in bytes).
```
nickbattam-tessella, line 93: is there a default page size? is there a property containing the page size to set in the object?

documentation/smg/17_pixel_data.md
```diff
@@ -0,0 +1,427 @@
...
+  - Only a page size worth of pixels is loaded at any one time.
+  - If all pixels fit in one page, all pixels are loaded.
+
+- Temporary files hold changes to pixel data.
```
nickbattam-tessella, line 97: It would be helpful to include information on how to save an SQW file to merge temp files back and have an updated file on disk.
Does that work if you save back over the original file or does it require a new file to be created?

documentation/smg/17_pixel_data.md
hsautessella: I was reluctant to put the name of the function here, since it's a private method and the name could change. If you think it would be better to include the name, I'm happy to do so

nickbattam-tessella (approved):

oerc0122 (approved):

documentation/smg/17_pixel_data.md
```diff
@@ -0,0 +1,449 @@
...
+
+>> head_run_idx = pix.run_idx(1:10);
+>> size(head_run_idx)
+    ans =
+      [1, 10]
+
+>> tail_coordinates = pix.coordinates((end - 10):end);
+>> size(tail_coordinates)
+    ans =
+      [4, 10]
```
cmarooney-stfc, line 202: Not sure this is right? If end==100, then end-10==90 and the range 90:100 has size 11.

documentation/smg/17_pixel_data.md
```diff
@@ -0,0 +1,449 @@
...
+whether that data is in memory or is in a file.
+The indexing argument attempts to mimic that of a normal Matlab array,
+i.e. logical indices are allowed and indices can be given out of order.
+However, it does not support indexing using `end`,
+you can use the `num_pixels` property to achieve the same effect.
+
+The getter methods have the advantage of being able to get/set multiple columns
+at once.
+This can be more efficient, as you're performing a single slice, instead of
+multiple.
```
cmarooney-stfc, line 246: Instead of multiple what?

documentation/smg/17_pixel_data.md
```diff
@@ -155,6 +155,22 @@ and operate on those pixels.
...
+This method, called when saving an `sqw` object,
+loops over pixel pages and saves them to the `sqw` file.
+
+There also exists `sqw_binfile_common.put_bytes`
+which can be used to save arbitrary data to an `sqw` file.
+You can position the `sqw_binfile_common` instance's file handle at the start
+of the `sqw` file's pixel array,
+and save pixels directory to file this way.
+This is more efficient than saving out to temporary files and then
+reading them back in later when calling `put_pix`.
```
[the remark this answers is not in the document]
hsautessella: `directory` was a typo, it was meant to say `directly`.

I've added an example of what's being described here. Hopefully that addresses any confusion.

oerc0122 (approved):

cmarooney-stfc (approved):

PR #633, opened 2021-03-24
Add ADR 23

oerc0122:
Create ADR 23 outlining decisions made around error identifiers in PACE.

Also update surrounding ADRs to match format and link correctly.

### DIFF
--- a/documentation/adr/0023-error-format.md
+++ b/documentation/adr/0023-error-format.md
@@ -1,0 +1,74 @@
+[<-previous](./0022-use-keyword-args.md) |
+next->
+
+# 23 - Error Format
+
+Date: 2021-Mar-24
+
+## Status
+
+Accepted
+
+## Context
+
+Currently throughout PACE, different error identifiers are used to reference the same error.
+This makes it difficult in cases of specific try-catch (as described in [ADR 21](./0021-errors-and-warnings.md)) to know what to compare against.
+
+Examples include:
+```
+herbert_core/applications/multifit/@mfclass/fit.m                                             : Herbert:mfclass:invalid_argument
+herbert_core/classes/data_loaders/@rundata/private/gen_runfiles_.m                            : GEN_GRUNFILES:invalid_arguments
+herbert_core/utilities/general/catstruct.m                                                    : catstruct:InvalidArgument
+herbert_core/classes/data_loaders/parse_arg.m                                                 : PARSE_ARG:wrong_arguments
+
+herbert_core/classes/data_loaders/@rundata/get_par.m                                          : RUNDATA:invalid_argument
+herbert_core/classes/data_loaders/@rundata/get_rundata.m                                      : RUNDATA:invalid_arguments
+
+herbert_core/utilities/maths/noisify.m                                                        : HERBERT:noisify
+herbert_core/utilities/misc/objdiff.m                                                         : YMA:OBJDIFF:NotEnoughInputs
+herbert_core/utilities/xml_io_tools/xmlwrite_xerces.m                                         : xml:FileNotFound
+herbert_core/admin/matlab_nbits.m                                                             : MATLAB:NBITS
+
+_test/shared/matlab_xunit_ISISextras/@TestCaseWithSave/private/get_ref_dataset_.m             : TestCaseWithSave:invalid_argument
+_test/shared/matlab_xunit_ISISextras/@TestCaseWithSave/private/instantiate_methods_to_save_.m : TEST_CASE_WITH_SAVE:invalid_argument
+
+erbert_core/classes/MPIFramework/@iMessagesFramework/iMessagesFramework.m                     : iMESSAGES_FRAMEWORK:invalid_argument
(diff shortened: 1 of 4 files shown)


documentation/adr/0023-error-format.md
```diff
@@ -0,0 +1,74 @@
...
+
+Following discussions it was decided that it would be useful to have a standard format with 3 components, in the following format
+expanding upon that intially outlined in [ADR 21](./0021-errors-and-warnings.md).
+
+
+```matlab
+error('(HORACE|HERBERT):(function_name|ClassName):error_identifier')
+```
+
+This scheme is based on the current (2020b) MATLAB error identifiers common to many functions.
```
nickbattam-tessella, line 54: May be worth linking to MATLAB documentation (in the context Section) on which this template is based: https://uk.mathworks.com/help/matlab/ref/mexception.html definition of `errId`

documentation/adr/0023-error-format.md
```diff
@@ -0,0 +1,74 @@
...
+
+erbert_core/classes/MPIFramework/@iMessagesFramework/iMessagesFramework.m                     : iMESSAGES_FRAMEWORK:invalid_argument
+herbert_core/classes/MPIFramework/@iMessagesFramework/iMessagesFramework.m                    : MESSAGES_FRAMEWORK:invalid_argument
+herbert_core/classes/MPIFramework/@iMessagesFramework/private/mix_messages_.m                 : iMESSAGES_FRAMEWOR:invalid_argument
+```
+From which it should be obvious is not a consistent scheme for error identification, even within the same set of routines.
+
+## Decision
+
+As MATLAB standards do not enforce a particular style of error statement,
```
nickbattam-tessella, line 43: isn't this context?

documentation/adr/0023-error-format.md
```diff
@@ -0,0 +1,74 @@
...
+| `invalid_output`   | Insufficient outputs                                                                          |
+| `not_implemented`  | Called function is virtual/abstract                                                           |
+| `array_mismatch`   | Array dimensions are incompatible (will usually be identified by MATLAB's error)              |
+| `file_not_found`   | File not found on system                                                                      |
+| `io_error`         | Issues with opening, reading or writing files                                                 |
+| `system_error`     | Issues when calling system (shell) functions from MATLAB                                      |
+
+## Consequences
+
+In future error identifiers should be written to follow these guidelines and in cases where surrounding code is being updated an effort should be made to
```
nickbattam-tessella, line 73: There are more consequences:
- errors will be easier to identify in code and at runtime
- legacy errors will need to be updated
- defined standard will help future developers

documentation/adr/0023-error-format.md
```diff
@@ -0,0 +1,74 @@
...
+
+# 23 - Error Identifier Format
+
+Date: 2021-Mar-24
+
+## Status
+
+Accepted
+
+## Context
```
nickbattam-tessella, line 12: Alex observed during the discussion that MATLAB has changed its standard errors in the past; that may be worth noting here.

nickbattam-tessella: How do the standard errors you list below relate to the (current) MATLAB standards?

documentation/adr/0023-error-format.md
oerc0122: I originally had it in context, and moved it because I felt it flowed better, but will consider moving it back.

cmarooney-stfc (approved):

documentation/adr/0023-error-format.md
```diff
@@ -0,0 +1,88 @@
...
+- `HORACE` or `HERBERT` should be in all caps.
+- The `function_name` and `ClassName` should exactly match that of the parent function/class, including casing.
+- The `error_identifier` should be in `lower_snake` case and should succinctly identify the error message, using one of
+     the following names if the error falls within their remit.
+     **NB.** This is in contrast to the usual MATLAB format which uses `lowerCamelCase`, but is in keeping with the majority
+     of identifiers currently used in PACE.
+
+| Identifier         | Issue                                                                                         |
+| :----------------- | :-------------------------------------------------------------------------------------------- |
+| `invalid_argument` | Argument fails validation, there are insufficient arguments or argument is an unexpected flag |
```
hsautessella, line 68: What happens if we're using something like `validattributes`? Should we be catching the errors and re-throwing with our ID? I'm not sure that I mind either way, but a convention on what to do in these cases would be good.

oerc0122: A very good question. I would probably say use MATLAB's throw. The advantage of throwing our own means we know exactly what it is and it is consistent. 

Eventually (you might say I'm a dreamer, but I'm not the only one), I can see either MATLAB's `inputParser`/`validateattributes` or PACE's `parse_args` being used globally and, either way, consistency will be achieved. Even if we end up slowly phasing `invalid_argument` or MATLAB's equivalent, that's really the key. 

`invalid_argument` will be useful for those times where parsing beyond that of validate is used (though admittedly if `inputParser` is used that argument becomes moot with the ability to define arbitrary validation.

PR #635, opened 2021-03-24
Fix erroneous print

oerc0122:
Fix `assertExceptionThrown` printing since https://github.com/pace-neutrons/Herbert/pull/334


nickbattam-tessella (approved):
Approved, assuming this has removed all unwanted exceptions from the build logs, which are truncated on Jenkins as there were no test failures.

hsautessella (approved):

cmarooney-stfc (approved):

Issue #639, opened 2021-03-26
Write package install script for users


hsautessella:
The installation process for users is a bit fiddly, and involves them directly editing variables in `.m` files.

We should write a Matlab `horace_install` script that will do all the necessary file editing/copying. A user should be able to use Horace (including mex and parallel) by performing the following steps:

1. Download the release package from GitHub
2. Extract the package anywhere on the file system they have read access
3. From within Matlab, run `cd <path_to_extracted_release>; horace_install>`
4. Now from any directory, users should be able to run `horace_on` to enable Horace.

cmarooney-stfc:
@abuts @oerc0122 Adding us as reviewers so we can close what looks like a done ticket

PR #640, opened 2021-03-26
Write package install script for users

hsautessella:
This PR adds an installation script (`install_horace.m`) that users can call to install a Horace release package.

The release package is structured as:
```txt
Horace-3.5.1-win64-R2019b/
  `-- Horace/
      `-- horace_init.m
       -- ...
  `-- Herbert/
      `-- herbert_init.m
       -- ...
  `-- herbert_on.m.template
  `-- horace_on.m.template
  `-- horace_install.m
  `-- worker_v2.m.template
  `-- LICENSE
  `-- README.md
```

The idea is that users download the package, extract it to some location, `cd` into that location in Matlab and run `horace_install`.
This will look for `userpath` (creating it if it doesn't exist) and copy over the template files (removing the `.template` extension), updating the template variables within each file.
Users should then be able to call `horace_on` from anywhere.

Please do build a release package (`./tools/build_config/build.ps1 -configure -package`) and see if the installation works on your machine.

~Note that this script will not work for an instance of Horace checked out of Git. This is due to the different relative locations of the `_init.m` files, particularly as a non-installed, non-package, version of Horace has no knowledge of where Herbert (and therefore `herbert_init`) is. It should be possible to adapt the script to run from a git checkout if required - you can pass some extra candidate paths to the `find_file` function.~

**Edit:**
The installation script should also work for an instance of Horace/Herbert checked out (from Git) into the same directory (e.g. `horace_root = .../pace/Horace` and `herbert_root = .../pace/Herbert`). If Horace/Herbert are not in the same directory, you can point the script to Herbert using the `herbert_root` name-value pair argument.

Addresses #639


admin/horace_install.m
```diff
@@ -0,0 +1,134 @@
+function horace_install()
+%HORACE_INSTALL install this instance of Horace
+%
+HORACE_ON_SUB_PATTERN = '${Horace_CORE}';
```
nickbattam-tessella, line 4: `xxx_SUB_PATTERN`? Reading how this is used it's a `_PLACEHOLDER`

admin/horace_install.m
```diff
@@ -0,0 +1,134 @@
...
+HERBERT_ON_SUB_PATTERN = '${Herbert_CORE}';
+
+install_root = fileparts(mfilename('fullpath'));
+
+% Find directories containing herbert/horace_init.m
```
nickbattam-tessella, line 9: This function is doing the same thing twice:
- find xxx_on
- reat content
- replace placeholder with xxx_init_dir
- save file to userpath

Rather than duplicating the code and interlacing it, two function calls to a single implementation would be easier to follow

admin/horace_install.m
```diff
@@ -1,40 +1,41 @@
...
+    herbert_on_path, ...
+    fullfile(user_path, 'herbert_on.m'), ...
+    {HERBERT_ON_PLACEHOLDER}, ...
+    {her_init_dir} ...
+);
 
 % Copy worker_v2 script (required by parallel routines) to userpath
 worker_path = find_file('worker_v2.m.template', {install_root});
-copy_file(worker_path, fullfile(user_path, 'worker_v2.m'));
+install_file(worker_path, fullfile(user_path, 'worker_v2.m'));
```
nickbattam-tessella, line 38: does redirecting this `copy()` though `install_file`  complicate things? It changes that from being a simple function into one with some logic

admin/horace_install.m
```diff
@@ -1,40 +1,41 @@
...
-new_hor_on_contents = replace(horace_on_contents, HORACE_ON_SUB_PATTERN, hor_init_dir);
-new_hor_on_contents = replace(new_hor_on_contents, HERBERT_ON_SUB_PATTERN, her_init_dir);
-
-% Find and update herbert_on
-herbert_on_path = find_file('herbert_on.m.template', {install_root});
-herbert_on_contents = fileread(herbert_on_path);
-new_her_on_contents = replace(herbert_on_contents, HERBERT_ON_SUB_PATTERN, her_init_dir);
+% Do this first before find_userpath, so we can error out before potentially
+% creating a directory
+hor_init_dir = find_directory('horace_init.m', {fullfile(install_root, 'Horace')});
```
nickbattam-tessella, line 12: why not pass `install_root` and `'Horace'` or `'Herbert'` arguments to `find_directory`? e.g. 
```
xxx_init_dir = find_directory('xxx_init.m', 'Xxxxx', install_root);
```

These are the only uses of `find_directory` so the `candidate_dirs` argument only ever contains one path

nickbattam-tessella:
"Note that this script will not work for an instance of Horace checked out of Git."

Is there documentation for setting up paths for users with a clone of the source?

admin/horace_install.m
```diff
@@ -0,0 +1,164 @@
...
+    fullfile(user_path, 'herbert_on.m'), ...
+    {HERBERT_ON_PLACEHOLDER}, ...
+    {her_init_dir} ...
+);
+
+% Copy worker_v2 script (required by parallel routines) to userpath
+worker_path = find_file('worker_v2.m.template', {install_root});
+install_file(worker_path, fullfile(user_path, 'worker_v2.m'));
+
+disp('Horace successfully installed.')
```
oerc0122, line 40: Should we run some simple tests (i.e. calling horace_on, herbert_on and a quick parallel init) before claiming it's installed?

hsautessella:
Note that this PR depends on https://github.com/pace-neutrons/Herbert/pull/339. However, merging https://github.com/pace-neutrons/Herbert/pull/339 will likely break Horace builds until _this_ PR (#640) is merged. So they should be merged at the same time.

admin/CMakeLists.txt
```diff
@@ -55,27 +30,13 @@ configure_file("version.h.template"
...
-        FILES "${_template_file}"
-        DESTINATION "."
-        RENAME "${_new_file_name}"
-    )
-endfunction()
-
 set(TEMPLATE_FILES
     "horace_on.m.template"
     "worker_v2.m.template"
-    "worker_4tests.m.template"
```
oerc0122, line 75: Is this no longer used?

oerc0122 (approved):

admin/CMakeLists.txt
hsautessella: Yup, no use in having it in a release package. See https://github.com/pace-neutrons/Horace/pull/640/commits/038e70e0da95dd1d18525e07cfd5a93ca7825b73

Issue #637, opened 2021-03-25
Write installation instructions for users


hsautessella:
Write/update instructions for users on how to install the Herbert/Horace release packages.

PR #641, opened 2021-03-30
Write installation instructions for users

hsautessella:
I've updated the user docs with installation instructions based on #640.

I've removed a large chunk of the instructions at the bottom, as they were either no longer relevant (`hpc_config().accum_in_separate_process` is not an existing option), or because that information belongs elsewhere and should be re-written so that it is more up-to-date with current options. There's no mention of choosing a parallel framework for example.

Addresses #637 

This PR should not be merged until #640 is merged


documentation/user_docs/docs/Download_and_setup.rst
```diff
@@ -1,311 +1,126 @@
...
+ Download and setup
+####################
+
+*********************
+ System Requirements
+*********************
+
+-  Windows 64-bit, or Linux 64-bit operating system
 
+-  Plenty of free disk space in order to cope with the combined data files (> 30GB)
```
oerc0122, line 11: I wouldn't say this is part of the requirements for downloading and setting up Horace

oerc0122 (approved):

cmarooney-stfc (commented):
The document itself is quite clear and as such I do not have a problem. The changes from the old one are more problematic. This is a different way of installing. For those accustomed to the old way, a comment about why this is so different would be useful. It would also benefit newcomers who have seen the previous install guide and have doubts about whether they are doing the right thing.

hsautessella:
@cmarooney-stfc 
>The document itself is quite clear and as such I do not have a problem. The changes from the old one are more problematic. This is a different way of installing. For those accustomed to the old way, a comment about why this is so different would be useful. It would also benefit newcomers who have seen the previous install guide and have doubts about whether they are doing the right thing.

I'm not sure that I agree that the old installation method should be referenced, is it relevant to users installing the new versions of Horace? As a user, all I want from an installation guide is "how do I start using this software?", talking about the old method of installation could be confusing and is noise in a document that should be easy to follow.

> It would also benefit newcomers who have seen the previous install guide and have doubts about whether they are doing the right thing.

I'm not sure that I follow, why would a newcomer be reading the old version of the guide?

PR #642, opened 2021-03-30
Remove references to _old classes

nickbattam-tessella:
Remove all references to _old from the NON xxx_old class folders and update tests.

Some additional functions migrated from _old to new classes


nickbattam-tessella:
Removed Herbert branch tag: we're no-longer dependent on the sqw_old class shim

PR #643, opened 2021-03-31
SQW class rewrite

nickbattam-tessella:
Feature branch rewrite of SQW and DnD as newstyle matlab classes.


nickbattam-tessella:
Blocked until #642 is merged

PR #650, opened 2021-04-07
Replace build tests with build testing

oerc0122:


CMakeLists.txt
```diff
@@ -44,15 +41,15 @@ include(PACE_Version)
 set(PACE_MLINT_IGNORE ${PACE_MLINT_IGNORE} "${Herbert_ROOT}/**/*.m")
 
 include(PACE_CodeAnalysis)
-if(${BUILD_TESTS})
-    include(CTest)
+include(CTest)
+if(BUILD_TESTING)
```
cmarooney-stfc, line 45: Presuming that BUILD_TESTING is not required to be a $variable but checking - I can see it comes from the -D options below but so did $BUILD_TESTS

oerc0122: If is an odd cookie and doesn't quire the expansion manually. Purely stylistic, but it's the way I usually see ifs. 
https://cmake.org/cmake/help/latest/command/if.html#variable-expansion

abuts:
have no opinion on this -- may be will have after cmake course

abuts (commented):
have no opinion on this -- may be will have after cmake course

oerc0122:
Essentially the issue is that `BUILD_TESTS` sets `BUILD_TESTING` to on anyway (`enable_testing()` sets this flag as it's the standard CMake/CTest flag) this just means that we are using the CMake standard flag people will expect rather than one of our own designation.

PR #666, opened 2021-04-26
disp2sqw eval bug

abuts:
Simple solution for simple bug. `varargin` is empty if no optional arguments are available, so simple fix to the issue


rebeccafair (approved):
Fix confirmed! https://anvil.softeng-support.ac.uk/jenkins/job/PACE-neutrons/job/pace-integration/job/Branch-Scientific-Linux-7-2018b/32/console, thanks :)

PR #668, opened 2021-04-26
Fix issue with Horace Jenkins breaking

oerc0122:
Issue due to conflict between new Jenkins shared lib and modifications made prior


abuts (approved):
well, do not understand why the build dos not run without these changes but if it runs, I am happy

oerc0122:
The original error was just the conflict in the Jenkinsfile between the shared library version and this which hadn't updated to using it yet.

 Then the windows system couldn't find where its sphinx-build was. This is a really hacky way of getting around it, by including a copy in the branch. It still requires the sphinx-build packages, but I will try to construct a better way of detecting whether those exist.

Issue #670, opened 2021-04-28
Jenkins dying due to push docs


oerc0122:
Erroneous call in Jenkinsfile following push-docs update causes Jenkins to fail on Linux & Windows R2019b.

PR #671, opened 2021-04-28
Hotfix cleanup

oerc0122:
Fix issue with pushing docs, and make CMake check for sphinx correctly.


cmarooney-stfc:
@oerc0122 Hi Jacob, does this mean that ticket #670 can now be moved from in progress to in review?

Issue #111, opened 2020-01-29
Fix and re-enable failing Matlab tests


hsautessella:
Several Matlab tests were failing on the build servers (some also locally), to get the servers up and running these were disabled in #110. These need to be fixed/re-written and re-enabled. Below is a `.txt` file containing details about the various tests that were failing.

The commits disabling these tests were: bea56145a279b33188768a4f400fc4949dfa8b82 85eaec5b1801b20f822ea3304f84d77bc75a0b48

The disabled tests are:
- [x] `test_gen_sqw_cylinder::test_gen_sqw_cyl` -  single assertion disabled https://github.com/pace-neutrons/Horace/issues/155
- [x] `test_gen_sqw_powder::test_powder_cuts` - test function disabled https://github.com/pace-neutrons/Horace/issues/155
- [ ] `test_multifit::test_fit_single_or_array2` - test function disabled
- [x] `test_tobyfit::test_tobyfit_refine_crystal_1` - test script disabled https://github.com/pace-neutrons/Horace/issues/188
- [x] `test_tobyfit::test_tobyfit_1` - test script disabled
- [x] `test_tobyfit::test_tobyfit_let_2` https://github.com/pace-neutrons/Horace/issues/172
- [x] `test_tobyfit` - https://github.com/pace-neutrons/Horace/issues/172
- [x] `test_sqw` - suite disabled https://github.com/pace-neutrons/Horace/issues/139
- [x] `test_mslice_utilities` - suite disabled #182

[matlab_failing_tests.txt](https://github.com/pace-neutrons/Horace/files/4127719/matlab_failing_tests.txt)

**Outputs:**
- Tests enabled and passing on builds servers and locally

Depends on: https://github.com/pace-neutrons/Horace/issues/126

abuts:
strange list. 
I have just run the tests on iDaaaS machine and see only test_multifit::test_fit_single_or_array2 failing.
is it possible that these test are running long time? The run on iDaaaS machine takes 48min.

hsautessella:
Only the `test_sqw` tests were timing out - and that was due to an error with the MPI workers. I've put the error messages for the failing tests in the `matlab_failing_test.txt` file

hsautessella:
[...]

PR #674, opened 2021-04-29
167 resolution covariance bug fixing

tgperring:
Resolves some bugs in resolution ellipsoid plotting, and adds plotting imporvements. Depends on Herbert changes made in response to Herbert ticket #111 for which there is a mirror pull request.


_test/test_tobyfit/test_tobyfit_resfun_2.m
```diff
@@ -0,0 +1,198 @@
...
+            
+            self.ulen = 2*pi*sqrt(1/self.alatt(1)^2 + 1/self.alatt(2)^2);
+            self.vlen = 2*pi/self.alatt(3);
+            
+            self.save()
+        end
+        
+        
+        %--------------------------------------------------------------------------
+        function test_indepOfSampleOrientation(S)
```
oerc0122, line 60: Use consistent naming, you used `self` earlier (though I believe `obj` is preferred in MATLAB) and `S` here.

horace_core/algorithms/resolution_plot.m
```diff
@@ -116,38 +143,46 @@
...
     plot_args = {};
+    newplot = true;
 end
 
 
-
 % Construct sqw object
 % --------------------
 % Create a two-dimensional sqw object with one pixel
+psi = psi_deg * (pi/180);
```
oerc0122, line 181: Do we have a constant or a function anywhere for this common operation?

horace_core/algorithms/resolution_plot.m
```diff
@@ -233,33 +275,71 @@
...
-data.iint = [-Inf,-Inf; Inf,Inf];
+%data.iint = [-Inf,-Inf; Inf,Inf];
+data.iint = 1e-10*[-1,-1; 1,1];
 data.pax = iax(1:2);
 data.p = {1e-10*(-3:2:3)'/3, 1e-10*(-3:2:3)'/3};    % something tiny
 data.dax = [1,2];
-data.s = zeros(3,3);
-data.e = zeros(3,3);
-data.npix = zeros(3,3);
+data.s = [0,0,0; 0,NaN,0; 0,0,0];
```
oerc0122, line 283: Why are you inserting `NaN`s?  This usually means something weird is going on and can very easily result in unexpected or undefined behaviour.

horace_core/sqw/@sqw/get_nearest_pixels.m
```diff
@@ -1,34 +1,65 @@
...
+        'Only a single instance of an sqw object can be passed.')
+end
+
 nd = numel(win.data.pax);
 sz = size(win.data.npix);
-np = size(xp_in,1);
-if size(xp_in,2)~=nd
-    error('Number of coordinates in the point does not match dimensions of sqw object')
+np = size(xp,1);
+if numel(size(xp)) ~= 2
```
oerc0122, line 44: `ndims` would be better here. Though this doesn't match the specification in the error perhaps
```matlab
~(isvector(xp) || ismatrix(xp))
```
to make the intent clearer.

horace_core/sqw/@sqw/get_nearest_pixels.m
```diff
@@ -1,34 +1,65 @@
...
+        'Coordinate array must be a row vector or two-dimensional array.')
+elseif size(xp, 2) ~= nd
+    error('HORACE:get_nearest_pixels:invalid_argument',...
+        'Number of coordinates in the point does not match the dimensions of the sqw object.')
+end
+
+% Catch case of no query points on input
+% ----------------------------------------
+if size(xp, 1)==0
+    ok = false(0,1);
```
oerc0122, line 55: You have `size(xp, 1)` as the variable `np`

horace_core/sqw/@sqw/resolution_plot.m
```diff
@@ -54,76 +60,91 @@
...
+% purpose resolution function tool called resolution_plot that plots without a
+% dataset.
+
 
 
-% Check input arguments and get defaults where necessary
-% ------------------------------------------------------
 % Check sqw object
+% ----------------
 if ~isscalar(w) || dimensions(w)~=2
```
oerc0122, line 99: Is `dimensions` `ndims`?

horace_core/sqw/@sqw/resolution_plot.m
```diff
@@ -54,76 +60,91 @@
...
+% Get point(s) at which to calculate the resolution covariance matrix
+% Points are given in units of the projection axes
 if numel(par)==0
-    xp = xp_cent;
-elseif numel(par)==1 && npixtot>0
+    p = w.data.p;
+    xp = [0.5*(p{1}(1)+p{1}(end)), 0.5*(p{2}(1)+p{2}(end))];    % mid-points of plot axes
+elseif numel(par)==1
     xd = par{1};
     if ~(isnumeric(xd) && size(xd,2)==2 && size(xd,1)>0 && all(isfinite(xd(:))))
```
oerc0122, line 135: could be described by:
```matlab
validateattributes(xd, {'numeric'}, {'ncols', 2, 'nonempty', 'finite'},'resolution_plot', 'xd')
```

horace_core/sqw/@sqw/private/resolution_plot_private.m
```diff
@@ -16,26 +16,27 @@ function resolution_plot_private (x0,C,iax,flip)
...
+%           For plotting if the display axes are reversed from plot axes.
 
 
 % Original author: T.G.Perring
-%
-% $Revision:: 1759 ($Date:: 2020-02-10 16:06:00 +0000 (Mon, 10 Feb 2020) $)
 
 
 % Check input arguments
 if ~(isnumeric(x0) && numel(x0)==2 && all(isfinite(x0)))
```
oerc0122, line 26: Could be replaced by:
```matlab
validateattributes(x0, {'numeric'}, {'vector', 'numel', 2, 'finite'}, 'resolution_plot_private', 'x0')
validateattributes(C, {'numeric'}, {'size', [4,4]}, 'resolution_plot_private', 'C')
```
Can't cleanly replace the third check due to variable size and isunique.

horace_core/sqw/@sqw/resolution_plot.m
```diff
@@ -190,42 +205,57 @@
 [inst, all_inst] = get_inst_class(w);
 if isempty(inst)
     if all_inst
-        error('The instrument type is not the same for all contributing raw data files')
+        error('HORACE:resolution_plot:invalid_argument',...
+            'The instrument type is not the same for all contributing raw data files')
```
oerc0122, line 209: Same error on this if regardless of path.

horace_core/sqw/@sqw/resolution_plot.m
```diff
@@ -190,42 +205,57 @@
...
-    lx('round')
-    ly('round')
-end
-
 
 % Return covariance matrix, if requested
 % --------------------------------------
-if nargout==1
-    varargout{1} = covariance_matrix;
+if nargout>=1
```
oerc0122, line 250: Given you know the argument count, would it not be better to just make these static arguments? Matlab ignores non-returned arguments anyway.

oerc0122 (approved):

_test/test_tobyfit/test_tobyfit_resfun_2.m
tgperring: I found self.this self.that confusing when I was passing lots of arguments to calls in the tests; I inherited the self. from the template I used to make the test. Happy to change elsewhere to S.

horace_core/algorithms/resolution_plot.m
tgperring: Not that I know of. it would be handy though.

_test/test_tobyfit/test_tobyfit_resfun_2.m
oerc0122: Sometimes I just use an unused character like `ŋ` (alt gr+g on my keyboard) and then find and replace with what it's meant to be at the end.

horace_core/algorithms/resolution_plot.m
tgperring: A signal of NaN is a valid numerical value for an sqw object, even if apparently strange. The function exploits a feature of plotting sqw objects with a valid sqw object. If s(2,2) == 0 then the sqw method sqw/resolution_plot which is called will plot a bin with signal zero. With s(2,2) == NaN then plotting the sqw object results in a blank plot as NaN bins are excluded from being plotted. This means that the sqw method sqw/resolution_plot can be called without it having to know anything special about the origin of the sqw object being passed to it.

THis change avoids what was previously an incorrect sqw object - it declared that all elements of npix were zero, but the pix array still had one pixel in it. One day this will be caught as an invalid object.

oerc0122: Perhaps a short comment to this effect?

horace_core/sqw/@sqw/get_nearest_pixels.m
oerc0122: Correction, just want 
```matlab
~ismatrix(xp)
```
`ismatrix` checks for vectors too, apparently.

horace_core/sqw/@sqw/get_nearest_pixels.m
tgperring: well spotted. I hadn't.

horace_core/sqw/@sqw/get_nearest_pixels.m
tgperring: Didn't know of the existence of ndims...

tgperring: Matlab says use ismatrix as opposed to ndims(xp)==2; actually isvector doesn't permit trhe valid though trivial  case of size(0,n) (n>0)

oerc0122: `ismatrix` does allow that.

horace_core/sqw/@sqw/resolution_plot.m
tgperring: They are different. dimensions is a method of sqw that returns the dimensionality of the sqw object rather than ndims for the signal array. A one dimensional sqw object will have signal array size (n,1) (n>1); ndims(w.data.signal) will be returned as 2. In tthis context the sqw dimensionality is needed.

horace_core/sqw/@sqw/resolution_plot.m
tgperring: Should be OK. Its the confusing return from get_inst_class. Return variable inst will be non-empty only if both of the following are true (1) an instrument was defined for all contributing nxspe files in the sqw object and (2) all those instrumnets are of the same class. If inst is empty, then this can be because instrumnets are defined for all nxspe datasets (all_inst==true) but they are not all of the same class, or not all nxspe files had an instrument defined (all_inst==false)

horace_core/sqw/@sqw/resolution_plot.m
tgperring: This is a little trick. If one has static return arguments, then typing at the matlab prompt
>> resolution_plot (...)
would result in the first return argument being printed at the screen unless the line was terminated with a semi-colon. Mostly I expect this will be used without the uyser expecting a return argument - it is a visualisation utility - but they are there in case they are wanted (e.g. as I have done in the past). The odd looking if nargin>=1 etc lines enable this to be implemented.

PR #679, opened 2021-05-04
Fix spinw interface twins loop

### LINKED ISSUE #660: SpinW interface twin bug
Original reporter: @russellewings 

The [`spinw_disp.m`](https://github.com/pace-neutrons/Horace/blob/master/horace_core/functions/spinw_disp.m) gateway function currently does not correctly handle twins. The twins calculations is actually carried out but when converting from SpinW matrix format to the vector-of-cells required by `disp2sqw` only the first twin is included. The problematic code is in line [199](https://github.com/pace-neutrons/Horace/blob/master/horace_core/functions/spinw_disp.m#L199)

mducle:
Changes the loop max to account for all the twins.

Addresses #660

### DIFF
--- a/horace_core/functions/spinw_disp.m
+++ b/horace_core/functions/spinw_disp.m
@@ -94,7 +94,7 @@ if nTwin==0
 end
 w = cell(1,2*nMagExt*nTwin);
 s = cell(1,2*nMagExt*nTwin);
-for jj = 1:2*nMagExt
+for jj = 1:2*nMagExt*nTwin
     w{jj} = zeros(nHkl,1);
     s{jj} = zeros(nHkl,1);
 end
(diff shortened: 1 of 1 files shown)


cmarooney-stfc (approved):

cmarooney-stfc:
@mducle Hi Duc - this has been approved so guessing you should merge. Noting that the merge check is hung or something so you may need to prod it

cmarooney-stfc:
@mducle Hi again Duc. Also, just to say, the way of linking PRs and issues has changed as we're using Zenhub. Instead of doing "fixes #NNN" can you say "addresses #NNN" as the fixes version will close the ticket and we want its status to change to Done (we will close it after the sprint review). Thanks

PR #680, opened 2021-05-04
Fix spaghetti_plot 1D if no data in 2D cut bug

### LINKED ISSUE #661: Bug in spaghetti_plot for handling 1D cuts when there is no data.
Original reporter: @russellewings 

[`spaghetti_plot`](https://github.com/pace-neutrons/Horace/blob/master/horace_core/spaghetti_plot.m) has the option to output a series of 1D cuts as well as the usual 2D color plots, if the user specifies two output arguments.

The 1D cuts are made from the 2D cuts, but there is no check to see if there is any data in the 2D cut first. So, in these cases an error occurs and `sphaghetti_plot` fails.

The problem is around line [277](https://github.com/pace-neutrons/Horace/blob/master/horace_core/spaghetti_plot.m#L277)

mducle:
Fix bug in generating 1D cuts from the 2D slices normally output by `spaghetti_plot`.

First, originally the 2D cuts were `d2d` objects, and 1D cuts were made from this, but now we use `sqw` to make things more accurate. Also, there is a check if there are no pixels in the 2D which previously caused an exception.

Fixes #661

### DIFF
--- a/horace_core/spaghetti_plot.m
+++ b/horace_core/spaghetti_plot.m
@@ -273,12 +273,19 @@ for i=1:nseg
     u2 = [u20-u2bin,u20+u2bin];
     u3 = [u30-u3bin,u30+u3bin];
     % Make cut, and save to array of d2d
-    wdisp(i) = cut_sqw(sqw,proj,u1,u2,u3,ebin,'-nopix');
     if nargout>1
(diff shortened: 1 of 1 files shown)


abuts (approved):
clear useful fix. Go for it

cmarooney-stfc:
@mducle Hi Duc you may wish to merge this fix now it is approved, thanks

PR #682, opened 2021-05-07
Fix issues with too many arguments

oerc0122:


horace_core/sqw/multifit_subclasses/mfclass_Horace_sqw_sqw.m
```diff
@@ -821,7 +821,7 @@
                 wrapfun.bp_wrap = append_args (wrapfun.bp_wrap, '-ave');
                 obj_tmp.wrapfun = wrapfun;
             end
-            [data_out, calcdata, ok, mess] = fit@mfclass (obj_tmp, varargin{:});
+            [data_out, calcdata] = fit@mfclass (obj_tmp, varargin{:});
```
cmarooney-stfc, line 824: ok,mess are return values for this function and will now not be set. If not removed from this function, they should be set true,"" presuming that any other values will now be dealt with by an error throw in the fit call

horace_core/sqw/multifit_subclasses/mfclass_Horace_sqw.m
```diff
@@ -819,7 +819,7 @@
                 wrapfun.p_wrap = append_args (wrapfun.p_wrap, '-ave');
                 obj_tmp.wrapfun = wrapfun;
             end
-            [data_out, calcdata, ok, mess] = fit@mfclass (obj_tmp, varargin{:});
+            [data_out, calcdata] = fit@mfclass (obj_tmp, varargin{:});
```
cmarooney-stfc, line 822: ok,mess are return values for this function and will now not be set. If not removed from this function, they should be set true,"" presuming that any other values will now be dealt with by an error throw in the fit call

cmarooney-stfc (changes requested):

PR #685, opened 2021-05-14
Additional exception-s documentation

abuts:
Small addition to documentation, describing additional exception codes

### DIFF
--- a/documentation/adr/0023-error-format.md
+++ b/documentation/adr/0023-error-format.md
@@ -72,6 +72,8 @@ This scheme is based on the current (2020b) MATLAB error identifiers common to m
 | `array_mismatch`   | Array dimensions are incompatible (will usually be identified by MATLAB's error)              |
 | `file_not_found`   | File not found on system                                                                      |
 | `io_error`         | Issues with opening, reading or writing files                                                 |
+| `runtime_error`    | Unspecified issues, caused by running valid code under specific circumstances                 |
+|                    | (e.g. out of memory, or can not start some parallel cluster due to network error)             |
 | `system_error`     | Issues when calling system (shell) functions from MATLAB                                      |
 
 ## Consequences
--- a/documentation/adr/0023-error-format.md
+++ b/documentation/adr/0023-error-format.md
@@ -72,8 +72,6 @@ This scheme is based on the current (2020b) MATLAB error identifiers common to m
 | `array_mismatch`   | Array dimensions are incompatible (will usually be identified by MATLAB's error)              |
 | `file_not_found`   | File not found on system                                                                      |
 | `io_error`         | Issues with opening, reading or writing files                                                 |
-| `runtime_error`    | Unspecified issues, caused by running valid code under specific circumstances                 |
-|                    | (e.g. out of memory, or can not start some parallel cluster due to network error)             |
 | `system_error`     | Issues when calling system (shell) functions from MATLAB                                      |
 
 ## Consequences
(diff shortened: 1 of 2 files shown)


documentation/adr/0023-error-format.md
```diff
@@ -68,9 +68,12 @@ This scheme is based on the current (2020b) MATLAB error identifiers common to m
 | `invalid_argument` | Argument fails validation, there are insufficient arguments or argument is an unexpected flag |
 | `invalid_output`   | Insufficient outputs                                                                          |
 | `not_implemented`  | Called function is virtual/abstract                                                           |
+| `not_available`    | Method or function is not available on current system or current Horace/Herbert configuration |
 | `array_mismatch`   | Array dimensions are incompatible (will usually be identified by MATLAB's error)              |
 | `file_not_found`   | File not found on system                                                                      |
 | `io_error`         | Issues with opening, reading or writing files                                                 |
+| `runtime_error`    | Unspecified issues, caused by running valid code under specific circumstances                 |
```
oerc0122, line 75: I would prefer not a generic "runtime error" there are all runtime errors. I think 
"OOM_error" and "parallel_error" might be good options though.

abuts: runtime error may be also issue with mex files or something really unexpected. e.g.

oerc0122: However, it gives no information. I'd rather less standardised errors relating to what went wrong than "runtime error".

abuts:
```            [ok,mess,ndet,en] = is_loader_valid(this);
            if ok<1
                error('A_LOADER:runtime_error',mess);
            end
```
incorrect loader selected

`'HERBERT:runtime_error','Can not find Matlab'`  -- clear -- not IO, 
```
error('JOB_DISPATCHER:runtime_error',...
        'Attempt to restart job when the cluster is not running');

error('PARALLEL_FRAMEWORK:runtime_error',...
          ' Can not start parallel cluster %s after %d attempts. Parallel 

error('HERBERT_CONFIG:runtime_error',...
        'herbert_init is not on Matlab search path');

     
% check and set valid property, verifying all connected fields are
% consistent
        [ok,mess] = obj(ibegw1(i)+j).check_joint_fields();
        if ok
            obj(ibegw1(i)+j).valid_ = true;
        else % can not ever happen, unless w2 is invalid
            error('IX_data_1d:runtime_error',mess);
        end
```
do not even know why that. may be of course its invalid argument, but...
```
      error('HERBERT:config_folder:runtime_error',...
            'Cannot create configuration directory %s; Error: 
```
usually incorrect permissions IO error is not really good here...
```
                       error('JETester:runtime_error',...
                            'simulated failure for lab N %d',obj.labIndex);
```
nice one -- how would you name this?

cmarooney-stfc: How about "unexpected error" - since we were not expecting it and it alerts the user to unusual behaviour"

oerc0122: I would argue for `parallel_error` personally.

cmarooney-stfc: "unexpected" was in response to previous comments not Alex's latest.

cmarooney-stfc: In Alex's examples, the message qualifying runtime error makes the situation clearer. Maybe that is a requirement for "runtime error" - definitely need more info in the message

abuts: have nothing against parallel_error -- let it be. 

`runtime_error` is often thrown in Mantid -- its convenient to use common language rather then unexpected error, which makes us sound like dummies

abuts: added parallel_error to suggestions.

Issue #688, opened 2021-05-24
bug in spagetty plot


abuts:
code 
```Matlab
clear all
data_path1='/instrument/MAPS/CYCLE20211/RB2010464/horace_code/';%directory where sqw files go
sqw_file=[data_path1,'Ei50meV_5K_sym.sqw'];
%%
K=[-0.75 1.25 0];
L1=[-1/2 3/2 -1/2];
U=[-1/4 1 -1/4];
X=[-1 1 0];
W=[-1/2 1 0];
G=[-1 1 -1];
L2=[-1.5 1.5 -0.5];

Xr=['X(',num2str(X(1)),',', num2str(X(2)),',', num2str(X(3)),')'];
Kr=['K(',num2str(K(1)),',', num2str(K(2)),',', num2str(K(3)),')'];
L1r=['L(',num2str(L1(1)),',', num2str(L1(2)),',', num2str(L1(3)),')'];
Ur=['U(',num2str(U(1)),',', num2str(U(2)),',', num2str(U(3)),')'];
Wr=['W(',num2str(W(1)),',', num2str(W(2)),',', num2str(W(3)),')'];
Gr=['\Gamma(',num2str(G(1)),',', num2str(G(2)),',', num2str(G(3)),')'];
L2r=['L(',num2str(L2(1)),',', num2str(L2(2)),',', num2str(L2(3)),')'];

%%
close all
qbin=0.08;
qw=0.14;
[wdisp1,cuts1] =  spaghetti_plot([G;L1;K],sqw_file,...
    'labels',{Gr,L1r,Kr}...
    ,'qbin',qbin,'qwidth',qw,'ebin',[-5,0.2,25]);
caxis([0,1.5])
colormap('jet')
keep
[wdisp2,cuts2] =  spaghetti_plot([L1;K;W;L1;U],sqw_file,...
    'labels',{L1r,Kr,Wr,L1r,Ur}...
    ,'qbin',qbin,'qwidth',qw,'ebin',[-5,0.2,25]);
caxis([0,1.5])
colormap('jet')
keep
spaghetti_plot([wdisp1,wdisp2(2:end)])
colormap('jet')
keep
[wdisp,cuts] =  spaghetti_plot([G;L1;K;W;L1;U],sqw_file,...
    'labels',{Gr,L1r,Kr,Wr,L1r,Ur}...
    ,'qbin',qbin,'qwidth',qw,'ebin',[-5,0.2,25]);
caxis([0,1.5])
[...]

PR #689, opened 2021-05-24
fixing issue with spagetti plot, producing non-orhtogonal projections

abuts:
small change, fixing the issue #688

### DIFF
--- a/horace_core/spaghetti_plot.m
+++ b/horace_core/spaghetti_plot.m
@@ -248,14 +248,14 @@ for i=1:nseg
             u2crt = cross(u1crt,b*[1;0;0]./norm(b*[1;0;0]));
         end
     else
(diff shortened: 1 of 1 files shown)


oerc0122 (approved):
If it fixes the bug

horace_core/spaghetti_plot.m
```diff
@@ -248,14 +248,14 @@
             u2crt = cross(u1crt,b*[1;0;0]./norm(b*[1;0;0]));
         end
     else
-        u2crt = u2crt0./norm(u2crt0);
+        u2crt = ortho_vec(u1crt);
```
cmarooney-stfc, line 251: wondering where ortho_vec is - it's not in this file or in the existing codebase

abuts: its another ticket in Herbert

PR #695, opened 2021-06-22
 Fix Horace GUI nxspe read check

abuts:
This is Duc's PR#694 released against master


horace_core/GUI/horace.m
```diff
@@ -2109,15 +2121,21 @@ function Rebin_rebin_pushbutton_Callback(hObject, eventdata, handles)
...
+    for i=stack_depth-3:-1:1
+        err{end+1} = sprintf('*** Row: %d, function: %20s file: %s',...
+            the_err.stack(i).line,the_err.stack(i).name,the_err.stack(i).file);
+    end
+    
+    set(handles.message_text_field,'String',char(err));
     guidata(gcbo,handles);
-    return;
+    
+    rethrow(the_err);
```
mducle, line 2138: Maybe all these error catch / printing to the GUI could be hived off into a separate function which is called rather than repeating the code >10 times?

mducle (changes requested):
Looks good to me... just that the error redirection code is repeated a lot of time needlessly and could be put into a separate function...

horace_core/GUI/horace.m
abuts: had such though. May be if its two, it worth implementing. Doing it only for master

Issue #692, opened 2021-06-21
Binary operations on sqw created from dnd doesn't work


mducle:
When a `dnd` object is promoted to an `sqw`, it has defined bins (fields `s`, `e`) but an empty `pix` fields - when a binary operation is applied to such an `sqw` object however, the bins seem to be ignored and only the `pix` field checked - this means that the resulting `sqw` has empty `s` and `e` fields and cannot be converted back into `dnd`. 

This particular bug manifests in the old-style fit functions (`*/fit_func`) which seem to require this conversion from `dnd` to `sqw` to access the correct method...

This old style fit function is used by the Horace GUI, which is needed for the Horace workshop *next week*.

A work around is to change the Hebert routine:

```diff
--- a/herbert_core/applications/multifit_legacy/private/multifit_func_eval.m
+++ b/herbert_core/applications/multifit_legacy/private/multifit_func_eval.m
@@ -101,6 +101,9 @@ if eval_back
                 else
                     if isempty(wout{i})
                         wout{i}=bkdfunc{1}(w{i},pars{:});
+                    elseif isempty(w{i}.data.pix)
+                        % Actually a dnd object
+                        wout{i} = sqw(dnd(wout{i}) + dnd(bkdfunc{1}(w{i}, pars{:})));
                     else
                         wout{i}=wout{i}+bkdfunc{1}(w{i},pars{:});
                     end
```

This is probably ok for a hotfix release for the Horace workshop *next week*...

(Note this is only a bug in release `3.5.2` - in master, you cannot create a `sqw` from a `dnd`, and the Horace GUI doesn't start).

To reproduce load this file: 
[example_d1d.zip](https://github.com/pace-neutrons/Horace/files/6686121/example_d1d.zip)

And run:

```matlab
wx = sqw(w2) + sqw(w2)
wx.data
```

And see that the `s` and `e` fields of `wx` is empty, and that `d1d(wx)` gives an error.

PR #696, opened 2021-06-24
bug in binary op

abuts:
this should fix the issue with binary operations, identified on ticket #692. The changes and unit tests are not compatible with master


mducle (approved):
Looks good to me and fixes the issue. :shipit:

Issue #691, opened 2021-06-14
Poor performance and memory managment of filebased cuts


abuts:
this ticket is closely related to #690. The development for this ticket has been started on branch `360_filebased_cut_bug`

Similarly to #690, if you run performance test on master, they run 40% slower then the code on `rel_3_5`. In addition to that, the cuts for 50 nxspe files almost exhaust 32Gb memory machine and the cuts on 150nxspe files fail even on 64Gb machine due to luck of memory. `rel_3_5` code barely occupy 13Gb for its operations under similar circumstances.

It is clear issue with the modified in Master code. 

As the poor performance have appeared because of appropriate tests are missing, one needs to write appropriate tests not to allow this in a future. Because of this, ticket #678 can be considered the part of this ticket

Issue #678, opened 2021-04-30
Performance and reliability of file-based cuts


abuts:
Recent ticket #669 have identified the issue with cuts not detected by existing unit tests. In addition to that, the performance of the fixture is not verified. This is the ticket to add the appropriate checks and the verification to the file-based cuts in addition to existing unit ans system tests.

abuts:
closed as this is the integral part of #691

PR #710, opened 2021-08-12
filebased cuts

abuts:
This PR does not completes the ticket #691 but creates set of tests which allow to verify and test the performance and highlight the errors. It may be considered as the completion of ticket #678, though it is combined with #691 because their close ties.


cmarooney-stfc (commented):
Hi Alex - sorry to be away when you submitted this. Just a few possible typos worth fixing.

For the odp presentation for performance, not clear if you intended to include it in github. If you did,  it could be useful to add a guide to the info present, as it will be unclear to anyone coming at it later. 

Thanks

PR #727, opened 2021-10-01
New sqw new instrument classes

cmarooney-stfc:
also sample


abuts (changes requested):
The PR contains committed local installation files in Horace/admin/ISIS -- the files are installed for local machine. It looks like they are on the search path before other installations which causes parallel workers to fail (hang-up)

This should be removed and master with new installation script merged into the branch too. 

Other review later, after tests are passing

abuts (changes requested):

horace_core/sqw/file_io/sqw_formats_factory.m
```diff
@@ -195,8 +197,15 @@
...
+                    header =sobj.my_header();
+                    if isa(header, 'Experiment')
+                        if isempty(header.expdata)
+                            loader = obj.supported_accessors_{1};
+                            return;
+                        else
+                            header = header.expdata(1);
+                        end
+                    elseif iscell(header)
                         header = header{1};
```
abuts, line 209: `sqw_formats` factory should not know anything about experiment and what is in the file. it should not rely on the header to identify this.

abuts (changes requested):

PR #757, opened 2021-11-10
Remove unnecessary build code for docs

oerc0122:
With the removal of `[NULL]` and the addition of the hidden doctrees, this workaround code is no longer necessary.


mducle (approved):
Since all the `[NULL]` sections have been removed this code is now redundant.

PR #761, opened 2022-01-12
Use null experiment components

cmarooney-stfc:
IX_inst and IX_samp, being the base classes of IX_inst_DGfermi (and other instruments) and IX_sample, have been deprecated as classes representing absence of instrument or sample data. IX_null_inst and IX_null_sample are used instead. This PR makes the change.

This PR parallels an equivalent one just made in Herbert. Individual PRs should fail, they have been tested together in a branch build and are OK.

### DIFF
--- a/_test/test_sqw_class/test_experiment.m
+++ b/_test/test_sqw_class/test_experiment.m
@@ -93,9 +93,9 @@ classdef test_experiment < TestCaseWithSave
             clear('expt');
             
             load(tmpfile, 'expt');
-            assertEqual(expt.instruments, IX_inst())
-            assertEqual(expt.samples, IX_samp())
-            assertEqual(expt.detector_arrays, [])
+            assertTrue( isempty(expt.instruments));
+            assertTrue( isempty(expt.samples));
+            assertEqual(expt.detector_arrays, []);
         end
         
         function test_instruments_setter_updates_value_for_valid_value(~)
--- a/_test/test_sqw_class/test_experiment.m
+++ b/_test/test_sqw_class/test_experiment.m
@@ -15,8 +15,8 @@ classdef test_experiment < TestCaseWithSave
             sample = IX_sample;
             expt = Experiment(detector_array, instrument, sample);
             
-            assertEqual(expt.samples, sample);
-            assertEqual(expt.instruments, instrument);
+            assertEqual(expt.samples{1}, sample);
+            assertEqual(expt.instruments{1}, instrument);
             assertEqual(expt.detector_arrays, detector_array);
         end
         
@@ -54,8 +54,8 @@ classdef test_experiment < TestCaseWithSave
                 [instrument, instrument], ...
                 [sample, sample]);
             
-            assertEqual(expt.samples, [sample, sample]);
-            assertEqual(expt.instruments, [instrument, instrument]);
+            assertEqual(expt.samples, {sample, sample});
+            assertEqual(expt.instruments, {instrument, instrument});
             assertEqual(expt.detector_arrays, [detector_array, detector_array]);
         end
         
@@ -63,7 +63,7 @@ classdef test_experiment < TestCaseWithSave
             tmpfile = fullfile([tempdir(), 'test_experiment', 'loadsave.mat']);
             clobR=onCleanup(@()self.delete_files(tmpfile));
             
-            instruments = [IX_inst_DGfermi(), IX_inst_DGdisk()];
+            instruments = {IX_inst_DGfermi(), IX_inst_DGdisk()};
             sample1 = IX_sample;
             sample1.name = 'sample1';
             sample2 = IX_sample;
@@ -76,10 +76,10 @@ classdef test_experiment < TestCaseWithSave
             clear('expt');
             
             load(tmpfile, 'expt');
-            assertEqual(expt.samples(1).name, 'sample1')
-            assertEqual(expt.samples(2).name, 'sample2')
-            assertTrue(isa(expt.instruments(1), 'IX_inst_DGfermi'));
-            assertTrue(isa(expt.instruments(2), 'IX_inst_DGdisk'));
+            assertEqual(expt.samples{1}.name, 'sample1')
+            assertEqual(expt.samples{2}.name, 'sample2')
+            assertTrue(isa(expt.instruments{1}, 'IX_inst_DGfermi'));
+            assertTrue(isa(expt.instruments{2}, 'IX_inst_DGdisk'));
             assertEqual(expt.detector_arrays, IX_detector_array);
         end
         
@@ -99,8 +99,8 @@ classdef test_experiment < TestCaseWithSave
         end
         
         function test_instruments_setter_updates_value_for_valid_value(~)
-            instruments = [IX_inst_DGfermi(),...
-                IX_inst_DGdisk()];
+            instruments = {IX_inst_DGfermi(),...
+                IX_inst_DGdisk()};
             expt = Experiment();
             expt.instruments = instruments;
             
@@ -121,9 +121,10 @@ classdef test_experiment < TestCaseWithSave
         function test_samples_setter_updates_value_for_valid_value(~)
             samples = IX_sample;
             expt = Experiment();
-            expt.samples = samples;
+            expt.samples = {samples};
             
-            assertEqual(expt.samples, samples);
+            assertEqual(expt.samples{1}, samples);
+            assertEqual(expt.samples,    {samples});
         end
         
         function test_samples_setter_raises_error_for_invalid_value(~)
--- a/_test/test_sqw_file/change_header_test.m
+++ b/_test/test_sqw_file/change_header_test.m
@@ -32,6 +32,7 @@ save(w,tmpsqwfile)
 if ~no_inst, set_instrument_horace(tmpsqwfile,inst); end
 if ~no_samp, set_sample_horace(tmpsqwfile,samp); end
 tmpfromfile=read_sqw(tmpsqwfile);
+newtmpfromfile = sqw(tmpsqwfile);
 [ok,mess]=equal_to_tol(wnew,tmpfromfile,'ignore_str',1); if ~ok, assertTrue(false,mess), end
 
 % Delete output file, if can
--- a/horace_core/sqw/@Experiment/Experiment.m
+++ b/horace_core/sqw/@Experiment/Experiment.m
@@ -2,9 +2,9 @@ classdef Experiment < serializable
     %EXPERIMENT Container object for all data describing the Experiment
     
     properties(Access=private)
-        instruments_ = IX_inst.empty;
+        instruments_ = {}; %IX_inst.empty;
         detector_arrays_ = []
-        samples_ = IX_samp.empty;
+        samples_ = {}; % IX_samp.empty;
         expdata_ = IX_experiment();
     end
     
@@ -57,22 +57,33 @@ classdef Experiment < serializable
                 end
             elseif nargin==3
                 obj.detector_arrays = S;
-                if isa(varargin{2}, 'IX_inst')
-                    obj.instruments =  varargin{2};
-                elseif isempty(varargin{2})
-                    % do nothing, instruments is already empty on
-                    % construction
+                if isempty(varargin{2})
+                    % do nothing, instruments is already empty on construction
                 else
-                    error('HORACE:Experiment:invalid_argument',...
-                        'bad type for instruments on construction');
+                    instruments = varargin{2};
+                    if ~iscell(instruments)
+                        instruments = num2cell(instruments);
+                    end
+                    if IX_inst.cell_is_class(instruments)
+                        obj.instruments =  instruments;
+                    else
+                        error('HORACE:Experiment:invalid_argument',...
+                            'bad type for instruments on construction');
+                    end
                 end
-                if isa(varargin{3}, 'IX_samp')
-                    obj.samples = varargin{3};
-                elseif isempty(varargin{3})
+                if isempty(varargin{3})
                     % do nothing, samples is already empty on construction
                 else
-                    error('HORACE:Experiment:invalid_argument',...
-                        'bad type for samples on construction');
+                    samples = varargin{3};
+                    if ~iscell(samples)
+                        samples = num2cell(samples);
+                    end
+                    if IX_samp.cell_is_class(samples)
+                        obj.samples = samples;
+                    else
+                        error('HORACE:Experiment:invalid_argument',...
+                            'bad type for samples on construction');
+                    end
                 end
             else
                 error('HORACE:Experiment:invalid_argument', ...
@@ -89,6 +100,9 @@ classdef Experiment < serializable
             % usage is deprecated and will be removed in a future.
             %
             samp = obj.get_unique_samples();
+            if iscell(samp)
+                samp = samp{1};
+            end
             if nargin == 2
                 oldhdrs = obj.expdata_(header_num).to_bare_struct();
                 oldhdrs.alatt = samp.alatt;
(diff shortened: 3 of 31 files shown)


horace_core/sqw/@Experiment/Experiment.m
```diff
@@ -2,9 +2,9 @@
     %EXPERIMENT Container object for all data describing the Experiment
     
     properties(Access=private)
-        instruments_ = IX_inst();
+        instruments_ = IX_inst.empty;
         detector_arrays_ = []
-        samples_ = IX_samp();
+        samples_ = IX_samp.empty;
         expdata_ = IX_experiment();
     end
```
abuts, line 9: This form (with empty arrays) was causing problem with previous code. Need check if the problem had been fixed when failure due to old herbert is resolved

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
```diff
@@ -1,37 +1,62 @@
...
-samples = repmat(IX_samp(),1,numel(headers));
 
-% convert old headers, resored differently from sqw and mat files into the
-% same format
+% Make arrays of the hew-style header class types to receive the data
+% coming from each run in headers
+expdata     = repmat(IX_experiment(), 1,numel(headers));
+instruments = repmat(IX_null_inst(),  1,numel(headers));
+samples     = repmat(IX_null_sample(),1,numel(headers));
+
```
abuts, line 27: we never had empty sample in our old Horace. It was always IX_samp (with lattice) . Empty instrument would be ok

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
```diff
@@ -1,37 +1,62 @@
...
-                instr.fermi_chopper);
-            instruments(i) = instrument;
+            instruments(i) = IX_null_inst();
+        % Struct may have data to make an instrument, delegate to factory
+        % method (which may reject the struct and throw)
         else
-            error('HORACE:Experiment:invalid_argument',...
-                'unknown structure tried to be recovered as instrument');
+            instruments(i) = make_instrument_from_struct(instr);
         end
```
abuts, line 57: dont we have instrument constructor for this?

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
```diff
@@ -40,45 +65,91 @@
...
-        sampl.alatt = alatt;
-        sampl.angdeg = angdeg;
-    elseif isa(sampl,'IX_samp') && isempty(sampl)
+    
+    % If sampl is a struct
+    if isstruct(sampl)
+        % Sample in header is an empty struct, replace with null sample and
+        % populate with the lattice parameters
+        if isempty(sampl) || isempty(fieldnames(sampl))
+            sampl = IX_null_sample();
```
abuts, line 76: null_sample means sqw object is broken. Have never happened and should not in a future.

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
```diff
@@ -88,5 +159,82 @@
...
+        'unknown structure tried to be recovered as instrument [2]');
 end
+
+end % function make_instrument_from_struct
+
+%-------------------------------------------------------------------------
+function sample = make_sample_from_struct(~) % when used, change ~ to sample_info)
+%
+% Temporary location for this sample factory which should really go in
+% Herbert instrument classes location
```
abuts, line 216: the same as for instrument, about sample

horace_core/sqw/file_io/@faccess_sqw_v3/private/get_instr_or_sample_.m
```diff
@@ -19,10 +19,10 @@
 end
 res_block = get_all_instr_or_samples_(obj,field_name);
 if strcmp(field_name,'instrument') && isstruct(res_block) && numel(fields(res_block))==0
-    res_block  = IX_inst();
+    res_block  = IX_null_inst();
 end
 if strcmp(field_name,'sample') && isstruct(res_block) && numel(fields(res_block))==0
-    res_block  = IX_samp();
+    res_block  = IX_null_sample();
 end
```
abuts, line 26: we never had `null_sample` -- `IX_samp` with lattice only have to be defined for anything to work.

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
oerc0122: Nonetheless, defending against it and being able to catch it is useful.

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
```diff
@@ -40,45 +65,91 @@
...
         sampl.alatt = alatt;
         sampl.angdeg = angdeg;
+            samples(i) = sampl;
+        % struct may have enough info to make a sample (though this is not
+        % defined yet - call to factory method will probably fail)
     else
-        sampl = IX_samp(sampl);
-        if isempty(sampl)
+            samples(i) = make_sample_from_struct(sampl);
+        end
```
oerc0122, line 84: Indentation

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
```diff
@@ -40,45 +65,91 @@
...
-        if isempty(sampl)
+            samples(i) = make_sample_from_struct(sampl);
+        end 
+    % Sample is actually a subclass of IX_samp, so keep it but overwrite
+    % the lattice parameters if they were not in the old version
+    elseif isa(sampl,'IX_samp')
+        if isempty(sampl.alatt)
             sampl.alatt = alatt;
+        elseif sampl.alatt ~= alatt
+            warning('HORACE:Experiment:invalid parameter',...
```
oerc0122, line 91: Does `invalid parameter` want to be underscore separated?

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
```diff
@@ -40,45 +65,91 @@
...
+                'incoming sample alatt and old header alatt do not match');
+        end
+        if isempty(sampl.angdeg)
             sampl.angdeg = angdeg;
+        elseif sampl.alatt ~= alatt
+            warning('HORACE:Experiment:invalid parameter',...
```
oerc0122, line 97: Again

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
```diff
@@ -88,5 +159,82 @@
...
+% Output
+% ------
+%   >> sample: the constructed sample, currently not implemented, so fails
+%
+% Errors
+% ------
+% If the struct does not sufficiently define a sample (which it won't, not
+% defined yet), an error will be thrown
+
+sample = IX_null_sample(); % suppress error, should be removed when proper
```
oerc0122, line 234: Is proper functionality to be added?

oerc0122 (commented):
There're some parts which look incomplete in this PR, are they to be fixed in later PRs or purged?

_test/test_sqw_class/test_experiment.m
```diff
@@ -93,9 +93,9 @@ function test_load_save_default_object_creates_default_object(self)
             clear('expt');
             
             load(tmpfile, 'expt');
-            assertEqual(expt.instruments, IX_inst())
-            assertEqual(expt.samples, IX_samp())
-            assertEqual(expt.detector_arrays, [])
+            assertTrue( isempty(expt.instruments));
+            assertTrue( isempty(expt.samples));
+            assertEqual(expt.detector_arrays, []);
```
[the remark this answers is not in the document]
cmarooney-stfc: At L.97, in Experiment.m at LL.5,7, instrument and sample properties are now defined as cell arrays and initialised as {}, hence empty. The property is the cell array of samples, not an individual single sample. Previously it was a mixin.heterogeneous array and hence had the base-class IX_samp type, but with cell array there is not a predefined type. It was still empty in the previous version though.

horace_core/sqw/@Experiment/Experiment.m
cmarooney-stfc: These have now been overridden by empty cell arrays as described in the previous reply.

horace_core/sqw/@Experiment/Experiment.m
```diff
@@ -57,22 +57,33 @@
                 end
             elseif nargin==3
```
oerc0122, line 58: This may be better and more clear as a `switch-case` rather than a chain of `if`s

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
cmarooney-stfc: This is just initializing samples with a set of null samples indicating that the sample is not defined at this point. It will be populated further down this function. IX_samp is now the base class of IX_null_sample and IX_sample and although it is not now abstract to allow serializable to work with it, it should still be treated as abstract in how it is used. Hence IX_null_sample is used here.

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
cmarooney-stfc: We have constructors for the individual instruments, null, DGdisk, DGfermi. Here we have an unknown instrument defined through a struct, and need the function to return the correct instrument class object. Even with proper polymorphic behaviour in Matlab (not present) and handle classes (we're not using them and I don't think they provide anything here) the discovery of the struct type has not be done by something that is not a constructor before the construction can be done.

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
cmarooney-stfc: Samples frequently occur in the existing data from rundata, files etc with an empty values or with a fieldless struct. The input data did not have the sample info and presumed that the sample would be defined later. These lines formalize that by making an IX_null_sample to mark the existence of this variable (sample) as a sample with no data yet applied.

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
```diff
@@ -88,5 +159,82 @@
...
+end % numel(headers)
+
+end % function normalise_old_headers
+
+%-------------------------------------------------------------------------
+function instr = make_instrument_from_struct(instr_info)
+%
+% Temporary location for this instrument factory which should really go in
+% Herbert instrument classes location
+%
```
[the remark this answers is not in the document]
cmarooney-stfc: I said in a comment above why a constructor would not work. Your comment here raises the existence of a Herbert factory, which I cannot find. @abuts  please tell me where this factory is. Noting that loadobj would need to know the instrument type before being able to process the data - do we have a simple way already to tell the instrument type, which is what this method is doing here?

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
cmarooney-stfc: Same as previous comment.

horace_core/sqw/file_io/@faccess_sqw_v3/private/get_instr_or_sample_.m
cmarooney-stfc: As previously noted, IX_samp is now conceptually an abstract class - the (Abstract) has only been removed to allow serializable to work. As we do not know what is in the sample at this point, IX_null_sample is the correct indicator for this.

_test/test_sqw_file/change_header_test.m
```diff
@@ -32,6 +32,7 @@
 if ~no_inst, set_instrument_horace(tmpsqwfile,inst); end
 if ~no_samp, set_sample_horace(tmpsqwfile,samp); end
 tmpfromfile=read_sqw(tmpsqwfile);
+newtmpfromfile = sqw(tmpsqwfile);
```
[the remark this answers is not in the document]
cmarooney-stfc: well spotted. But this raises the question, we have the constructor of an sqw from a file, and the read_sqw method, which goes a different route (create struct, construct sqw from struct) to do the same thing. Which is our recommended usage and why? But yes, deleting the extra line next commit.

horace_core/sqw/@Experiment/Experiment.m
```diff
@@ -2,9 +2,9 @@
     %EXPERIMENT Container object for all data describing the Experiment
     
     properties(Access=private)
-        instruments_ = IX_inst();
+        instruments_ = IX_inst.empty;
```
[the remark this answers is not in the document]
cmarooney-stfc: This has been updated to make instruments_ an empty cell array on default construction. But the same issue applied. When Experiment is default constructed, there are no instruments. Whether the old definition of IX_inst as a mixin.heterogneous is used (so initializing as .empty) or the new one as a cell array (so initializing as {}), we want no instruments. The old initialization with a "empty" IX_inst causes definition problems.

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
```diff
@@ -1,37 +1,62 @@
...
-% in this case the header (which is what varargin{1} is in
-% this case) is a cell of runs. Consequently we run over
-% the runs in the cell doing just what we did to one run
-% header in the if block above
+%
+% Input
+% ------
+%   >> ~ : the object (an Experiment), not used - a completely new
+%          Experiment is created and output at the end
+%   >> headers == varargin{1}: cell array of old-style header structs
```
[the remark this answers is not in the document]
cmarooney-stfc: Good point - this was originally created by Alex, but should certainly be possible to simplify it with a headers argument

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
```diff
@@ -1,37 +1,62 @@
...
 for i=1:numel(headers)
+    % Name  i'th header locally
     hdr = headers{i};
+    
+    % Unpack lattice parameters for putting into sample
     alatt = hdr.alatt;
     angdeg = hdr.angdeg;
+    
+    % Local name the instrument (not liking plural instruments - 
+    % should be the single instrument from the old-syle header structure
```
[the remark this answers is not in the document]
cmarooney-stfc: As the old style header comes into this function it is a cellarray of individual run headers each with an instrument and sample field (both singular). Alex modifies this to instruments and samples in normalize_old_headers below, as part of some checking. Reluctant to modify a colleague's deliberate usage - @abuts may explain the nomenclature. instruments is here a single instrument in the single-run old style header, as distinct from the instruments field in class Experiment, which is the whole set of instruments from all the runs.

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
cmarooney-stfc: nice - fix next commit

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
cmarooney-stfc: thanks

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
cmarooney-stfc: thanks

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
cmarooney-stfc: This is an implementation gone off half-cocked. Thanks, will fix

horace_core/sqw/@Experiment/Experiment.m
cmarooney-stfc: unclear if you mean the nargin if or the subordinate ifs below the 2 allowed values. For the former, only 2 ifs plus else. For the latter, it's type checking, not sure if switch supports. Nice thought but let's leave it for now.

oerc0122: Referring to `nargin` where you have 0, 1, 3, other, but it's true it's non-essential.

PR #765, opened 2022-01-19
Fixing Horace GUI

abuts:
Merging to master the bugfixes identified on rel_3_6. All bugfixes are in Horace_gui, so testing is difficult but the user, seems happy

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Horace
+
 Horace is a suite of programs for the visualisation and analysis of large datasets from time-of-flight neutron inelastic scattering spectrometers.
 
 The details on what Horace is and how to use its latest version can be found on its [GitHub Documents pages](https://pace-neutrons.github.io/Horace/unstable/). The documentation for the latest released version is available [here](https://pace-neutrons.github.io/Horace).


horace_core/GUI/horace.m
```diff
@@ -49,6 +49,24 @@
     strcmp(element_class,'d3d') || strcmp(element_class,'d4d') ||...
     strcmp(element_class,'sqw');
 
+function [vec_val,ok] = read_vector(str_val)
+% read vector value from a string, obtained from GUI
```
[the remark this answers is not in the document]
abuts: no, it reads and returns the vector, as provided in description. textscan returns the vector wrapped in cellarray and the code unwraps the cellarray -- this is what you probably misread as returning one value

horace_core/GUI/horace.m
```diff
@@ -49,6 +49,24 @@
     strcmp(element_class,'d3d') || strcmp(element_class,'d4d') ||...
     strcmp(element_class,'sqw');
 
+function [vec_val,ok] = read_vector(str_val)
```
oerc0122, line 52: Why is a new function returning `ok` rather than using error handling?

horace_core/GUI/horace.m
```diff
@@ -49,6 +49,24 @@
...
+% read vector value from a string, obtained from GUI
+ok = true;
+s1=strfind(str_val,'['); s2=strfind(str_val,']');
+if isempty(s1) && isempty(s2)
+    sv_new=textscan(str_val,'%f','delimiter',',');
+elseif ~isempty(s1) && ~isempty(s2)
+    str_val=str_val(s1+1:s2-1);
```
oerc0122, line 59: This function just ignores junk which could be bad. A valid input is
`ghosts[1,2,3]inthecode`
I would recommend you use `startsWith` and `endsWith`  and `(2:end-1)` for more strict input validation.

horace_core/GUI/horace.m
```diff
@@ -1436,52 +1455,32 @@ function Cut_do_cut_pushbutton_Callback(hObject, eventdata, handles)
...
     try
         %must strip out square brackets, if user has inserted them:
-        s1=strfind(a1,'['); s2=strfind(a1,']');
-        if isempty(s1) && isempty(s2)
-            a1new=textscan(a1,'%f','delimiter',',');
-        elseif ~isempty(s1) && ~isempty(s2)
-            a1=a1(s1+1:s2-1);
-            a1new=textscan(a1,'%f','delimiter',',');
-        else
+        [a1new,ok] = read_vector(a1);
```
oerc0122, line 1458: Using proper error handling (since all the error messages are the same, this could be reduced to:
```matlab
try
    a1new = read_vector(a1);
    a2new = read_vector(a2);
    a3new = read_vector(a3);
    a4new = read_vector(a4);
catch ME
    switch ME.identifier
      case HORACE:GUI:vec_read_error
          mess1='   Ensure binning values are entered if the form of lo,step,hi / step / lo,hi    ';
          mess2='NB: enter 0 if you wish to use intrinsic binning and entire data range along axis';
          set(handles.message_info_text,'String',char({mess_initialise,mess1,mess2}));
          guidata(gcbo,handles);
          return;
      otherwise
          rethrow(ME)
    end
end
```

horace_core/GUI/horace.m
```diff
@@ -1968,7 +1967,7 @@ function Rebin_rebin_pushbutton_Callback(hObject, eventdata, handles)
         %must strip out square brackets, if user has inserted them:
         s1=strfind(lostephi,'['); s2=strfind(lostephi,']');
```
oerc0122, line 1968: Isn't this chunk handled by `read_vector`? If not, why not?

oerc0122 (changes requested):
Please reformat to use new error-handling

horace_core/GUI/horace.m
```diff
@@ -1436,52 +1455,32 @@ function Cut_do_cut_pushbutton_Callback(hObject, eventdata, handles)
...
-        s1=strfind(a1,'['); s2=strfind(a1,']');
-        if isempty(s1) && isempty(s2)
-            a1new=textscan(a1,'%f','delimiter',',');
-        elseif ~isempty(s1) && ~isempty(s2)
-            a1=a1(s1+1:s2-1);
-            a1new=textscan(a1,'%f','delimiter',',');
-        else
+        [a1new,ok] = read_vector(a1);
+        if ~ok
             mess1='   Ensure binning values are entered if the form of lo,step,hi / step / lo,hi    ';
```
oerc0122, line 1460: Is this not the `report_error` function below's job?

horace_core/GUI/horace.m
abuts: I did minimal changes to what was there and this is easiest way of achieving this. ~ok will lead to different error messages depending on what vector is read.  Error handler is warped around each call to this operation and lead to different error messages to GUI -- changing this is too much for me to handle at this stage.

horace_core/GUI/horace.m
abuts: despite these operations are introduced in 2016b, I am still a bit reluctant to use operations, not compatible with the previous versions. May be should start changing this.

horace_core/GUI/horace.m
abuts: See above about the exceptions, but logic of this change is  wrong -- the checks are for incorrect value provided by user. In some cases empty is correct value, in some they are not. In other cases you read bunch or [] separated values. 

Not saying that the code in current `horace` is not entirely correct.  It should inform user about what exactly box is actually wrong. 

There are bunch of important changes asked to be done to GUI. Its not the time to start them now -- it would not stop easy.

horace_core/GUI/horace.m
oerc0122: Given we support 2018b+ I don't think we should be avoiding them. At the very least check `1` and `end` only.

horace_core/GUI/horace.m
oerc0122: Try-catches can be easily nested and `rethrow` means if not handled pass up to the next handler, it would be no different to the current if structure, but allows `read_vector` to be potentially portable, rather than limited in scope.

horace_core/GUI/horace.m
oerc0122: Empty would not return an error according to your definition of `read_vector`, I don't see your point there. 

Doing minimal changes now while we're there that make later cleanup easier just make sense to me.

horace_core/GUI/horace.m
abuts: this code may process bunch of brackets -- it splits the string into multiple strings separated by brackets. And read_vector deals with each peace. May be better solution is possible, but, again, not so keen on doing changes which currently can not be unit tested.  Job for a future GUI improvements

horace_core/GUI/horace.m
abuts: report error picks up the message and report it to user over GUI and command window. Here we form the message specific fro the inputs we are processing.

horace_core/GUI/horace.m
abuts: Had some thinking about the changes. `startWith` and `endWith` look much clearer, but substantially change the code behavior.  Will go well together with exceptions handling in the `read_vector` but I do not have desire and time to do this improvement now.

PR #766, opened 2022-01-20
Refactor some of Horace GUI

oerc0122:
Implement suggested changes


abuts (commented):
As I've said, the changes look reasonable, but I can not say they do not have trivial bugs just looking at the code. 
And to test the code at its current state means running GUI and looking through every change made, if it is correct. 
If you did that, I am glad to accept the changes.

cmarooney-stfc:
@oerc0122 Hi jacob, as per Alex last comment, can we merge?

oerc0122:
We don't want to merge this as it's not tested as we don't test the gui in the CI.

PR #767, opened 2022-01-24
Restructure tobyfit tests

oerc0122:
Move Tobyfit tests to be new-style class structure and unify several under two files.

Issues with resfun's test-data filesize.


_test/test_tobyfit/test_tobyfit.m
```diff
@@ -0,0 +1,560 @@
+classdef test_tobyfit < TestCaseWithSave
```
cmarooney-stfc, line 1: Couple of general comments. 

1. It's good that this has been fitted into the TestCase framework, but it is a very thorough refactor. Are we certain that the tests are unchanged in function? How could we verify?
2. There have been changes in the .mat files, which are binary so opaque. It would be good if the changes (which should effectively be null changes) are documented in the  PR description. I would expect that repackaging the saved objects to different names would not be a problem, but data changes might need further scrutiny.

oerc0122: I cannot prove the tests are unchanged beyond the fact that in creating the new object, they were copied and pasted with the visual/legacy checks stripped out, however, I do believe that the only important factor is that the new test-suite is deemed to thoroughly test the aspects of the Tobyfit stuff.

There have been no changes to the data (though I would like to at some point to reduce the 87MB Rb test data size), only repackaged to combine the 4 existing test files into two simpler modules.

_test/test_tobyfit/test_tobyfit.m
```diff
@@ -0,0 +1,560 @@
...
+
+            sample_fe2=IX_sample(true,[1,0,0],[0,1,0],'cuboid',[0.03,0.03,0.04]);
+            sample_fe2.alatt = [2.8504 2.8504 2.8504];
+            sample_fe2.angdeg = [90 90 90];
+
+            this.fe_3=ssi(fe_1,sample_fe2,600);
+            this.fe_4=ssi(fe_2,sample_fe2,600);
+            this.fe_arr2 = arrayfun(@(x) ssi(x, sample_fe2, 600), fe_arr);
+
+            % Mosaic tests - not finished
```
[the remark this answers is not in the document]
oerc0122: Those tests were unfinished by @tgperring in the original, they are reproduced here in their incomplete form for their subsequent completion in a future update. The first executable lines of the original file are:

```matlab

% ***************************************
%    Need to complete this test
% ***************************************
% Temporary dummy test
return
% ***************************************
% ***************************************
```

Beyond that, I do not know what the mosaic tests are to test, so I cannot complete them.

cmarooney-stfc:
@oerc0122 Hi Jacob, this has been around for a while. Review was OK but build is failing. Do we want to put it back in In Progress or is there a simple fix?

oerc0122:
Test is failing due to depending on another PR, which I believe has been merged. Need to kickstart the tests.

oerc0122:
Converted back to draft due to necessary effort to prove correctness.

oerc0122:
Superseded by #939

PR #784, opened 2022-03-03
move detpar to detector arrays

cmarooney-stfc:
It is planned to move the detector info from the separate detpar field in sqw with an array of IX_detector_arrays in the experiment_info field. This PR does this for Horace, initially with one IX_detector_array only. The parallel Herbert PR completes the process.

### DIFF
--- a/_test/test_sqw_class/test_sqw_copy.m
+++ b/_test/test_sqw_class/test_sqw_copy.m
@@ -28,19 +28,36 @@ classdef test_sqw_copy < TestCase
             sqw_copy = copy(sqw_obj);
             
             sqw_copy.main_header.title = 'test_copy';
-            sqw_copy = sqw_copy.change_header(struct([]));
+            assertFalse(equal_to_tol(sqw_copy.main_header, sqw_obj.main_header));
+            
+            % want to test expinfo/instruments and expinfo/detpar
+            % separately, so make 2 copies of expinfo
+            expinf1 = sqw_copy.experiment_info;
+            expinf2 = sqw_copy.experiment_info;
+
+            % all instruments changed with new name to make expinf different
+            for i=1:numel(expinf1.instruments), expinf1.instruments{i}.name = 'copy'; end
+            sqw_copy = sqw_copy.change_header(expinf1);
+            assertFalse(equal_to_tol(sqw_copy.experiment_info, sqw_obj.experiment_info));
+
+            % all detpar detectors changed with new azim value to make expinf different
+            for i=1:numel(expinf2.detector_arrays), expinf2.detector_arrays(i).azim = 0; end
+            sqw_copy = sqw_copy.change_header(expinf2);
+            assertFalse(equal_to_tol(sqw_copy.my_detpar(), sqw_obj.my_detpar()));
+            %{
+            % old version of detpar change with separate detpar left for
+            % reference to see what was intended
             dtp = sqw_copy.my_detpar();
             dtp.azim(1:10) = 0;
             sqw_copy = sqw_copy.change_detpar(dtp);
-            sqw_copy.data.dax = [2, 1];
-            sqw_copy.data.pix.signal = 1;
+            %}
             
             % changed data is not mirrored in initial
-            assertFalse(equal_to_tol(sqw_copy.main_header, sqw_obj.main_header));
-            assertFalse(equal_to_tol(sqw_copy.experiment_info, sqw_obj.experiment_info));
-            assertFalse(equal_to_tol(sqw_copy.my_detpar(), sqw_obj.my_detpar()));
+            sqw_copy.data.dax = [2, 1];
+            sqw_copy.data.pix.signal = 1;
             assertFalse(equal_to_tol(sqw_copy.data, sqw_obj.data));
             assertFalse(equal_to_tol(sqw_copy.data.pix, sqw_obj.data.pix));
+            
         end
         
         function test_copy_excluding_pix_returns_empty_pix_data(obj)
--- a/horace_core/sqw/@Experiment/Experiment.m
+++ b/horace_core/sqw/@Experiment/Experiment.m
@@ -3,7 +3,7 @@ classdef Experiment < serializable
     
     properties(Access=private)
         instruments_ = {}; %IX_inst.empty;
-        detector_arrays_ = []
+        detector_arrays_ = IX_detector_array.empty;
         samples_ = {}; % IX_samp.empty;
         expdata_ = IX_experiment();
     end
@@ -19,6 +19,7 @@ classdef Experiment < serializable
     properties(Dependent,Hidden)
         % property providing compartibility with old header interface
         header
+        detpar
     end
     properties(Constant,Access=private)
         fields_to_save_ = {'instruments','detector_arrays','samples','expdata'};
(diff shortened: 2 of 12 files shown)


_test/test_sqw_class/test_sqw_copy.m
```diff
@@ -28,19 +28,36 @@ function test_copy_returns_distinct_object(obj)
...
-            sqw_copy = sqw_copy.change_header(struct([]));
+            assertFalse(equal_to_tol(sqw_copy.main_header, sqw_obj.main_header));
+            
+            % want to test expinfo/instruments and expinfo/detpar
+            % separately, so make 2 copies of expinfo
+            expinf1 = sqw_copy.experiment_info;
+            expinf2 = sqw_copy.experiment_info;
+
+            % all instruments changed with new name to make expinf different
+            for i=1:numel(expinf1.instruments), expinf1.instruments{i}.name = 'copy'; end
```
oerc0122, line 39: Personally, I really dislike inline `do`s/`if`s like this. Definitely a personal matter.

oerc0122:
```matlab
expinf1.instruments = cellfun(@(x) setfield(x, 'name', 'copy'), expinf1.instruments, 'UniformOutput', false)
``` 
might be cleaner? Though admittedly MATLAB's verbosity in making the output a cell-array this sort of defeats that.

horace_core/sqw/@Experiment/Experiment.m
```diff
@@ -295,6 +300,15 @@
             head = [head{:}];
             head = rmfield(head,{'instrument','sample'});
         end
+        
+        function dp = get.detpar(obj)
+            if numel(obj.detector_arrays)>0
```
oerc0122, line 305: Is this not equivalent to `~isempty` which denotes the meaning more clearly

horace_core/sqw/@rundatah/private/calc_sqw_.m
```diff
@@ -134,7 +134,8 @@
 % ----------------------------------------------------------------------
 d.main_header=main_header;
 d.experiment_info=header;
-d.detpar=det0;
+d.experiment_info.detector_arrays(end+1) = IX_detector_array(det0);
+d.detpar=struct([]);
```
oerc0122, line 138: In keeping with other changes would `struct.empty` be better?

horace_core/sqw/@sqw/sqw.m
```diff
@@ -88,7 +88,11 @@
         wout = noisify(w,varargin);
         
         function dtp = my_detpar(obj)
-            dtp = obj.detpar_x;
+            if ~isempty(obj.detpar_x)
```
oerc0122, line 91: If both are empty this will error. Expected?

horace_core/sqw/@sqw/sqw.m
```diff
@@ -226,8 +230,19 @@
...
             % saveable class
+            % the generic code is preceded by the conversion of old-style
+            % detpar info into the new detector arrays. This allows loading
+            % of .mat files without converting them on disk..
+            if isfield(S,'experiment_info') && isfield(S,'detpar')
+                if ~isempty(S.detpar) && isempty(S.experiment_info.detector_arrays)
+                    dd = IX_detector_array(S.detpar);
+                    S.experiment_info.detector_arrays = IX_detector_array.empty;
+                    S.experiment_info.detector_arrays(end+1) = dd;
+                    S.detpar = struct([]);
```
oerc0122, line 243: `struct.empty`

horace_core/sqw/@sqw/sqw.m
```diff
@@ -226,8 +230,19 @@
     
     methods(Static)
         function obj = loadobj(S)
-            % boilerplate loadobj method, calling generic method of
+            % modified boilerplate loadobj method, calling generic method of
             % saveable class
+            % the generic code is preceded by the conversion of old-style
+            % detpar info into the new detector arrays. This allows loading
+            % of .mat files without converting them on disk..
+            if isfield(S,'experiment_info') && isfield(S,'detpar')
```
oerc0122, line 238: Just tested it and you can do
```matlab
if isfield(S, {'experimental_info', 'detpar'})
```

horace_core/sqw/file_io/@dnd_binfile_common/dnd_binfile_common.m
```diff
@@ -253,7 +253,8 @@ function check_obj_initated_properly(obj)
         function check_error_report_fail_(obj,pos_mess)
             % check if error occured during io operation and throw if it does happened
             [mess,res] = ferror(obj.file_id_);
-            if res ~= 0; error('SQW_FILE_IO:io_error',...
+            if res ~= 0
+                error('SQW_FILE_IO:io_error',...
```
oerc0122, line 257: `Horace:SQW_FILE_IO:io_error`

horace_core/sqw/file_io/@sqw_binfile_common/get_sqw.m
```diff
@@ -72,7 +72,14 @@
 % Get detector parameters
 % -----------------------
 if ~(opts.head||opts.his)
-    sqw_struc.detpar = obj.get_detpar();
+    dp = obj.get_detpar();
+    if isa(headers, 'Experiment')
+        dp = obj.get_detpar();
+        headers.detector_arrays(end+1) = IX_detector_array(dp);
+        sqw_struc.detpar = struct([]);
```
oerc0122, line 79: `struct.empty`

oerc0122 (changes requested):

horace_core/sqw/@Experiment/private/build_from_old_headers_.m
```diff
@@ -106,7 +106,7 @@
 
 % Construct the new header Experiment object
 % update with expdata, which maybe should go in the Experiment constructor
-obj = Experiment([], instruments, samples);
+obj = Experiment(IX_detector_array.empty, instruments, samples);
 obj.expdata = expdata;
```
abuts, line 110: I would allow old interface (user would not be so keen to learn about `IX_detector_array.empty` as `[]` looks much easier ) and make the internal assignment to `IX_detector_array.empty` as this happen

abuts (commented):
essential remark is about `loadobj`. 
Everything else looks like work in progress. 
Think about exposing important properties through public interface. All detector properties (`id, x2 psi, phi,` ... etc should be top level properties, returning appropriate arrays as function of run number. I think, it will greatly simplify the usage of a compressed arrays of instruments.

horace_core/sqw/@sqw/sqw.m
cmarooney-stfc: obj in this case is a sqw (or array thereof). Guess it is conceivable that obj could be an empty array with the right type. If it errors maybe we rewrite in python?

_test/test_sqw_class/test_sqw_copy.m
cmarooney-stfc: Happy to change the formatting. Arguable if the cellfun version is cleaner - it is very opaque and in the interests of making clear what a test does if there is the problem here I would retain the for loop.

PR #789, opened 2022-03-17
oriented_lattice dependent

abuts:
These are the changes, necessary due to Herbert changes

### DIFF
--- a/horace_core/sqw/@rundatah/rundatah.m
+++ b/horace_core/sqw/@rundatah/rundatah.m
@@ -107,7 +107,7 @@ classdef rundatah < rundata
             if nargin == 1 && isa(varargin{1},'sqw')
                 obj = rundata_from_sqw_(varargin{1});
             else
-                obj = obj.initialize(varargin{:});
+                obj = obj.init(varargin{:});
             end
         end
         %------------------------------------------------------------------
--- a/horace_core/sqw/@rundatah/rundatah.m
+++ b/horace_core/sqw/@rundatah/rundatah.m
@@ -15,7 +15,7 @@ classdef rundatah < rundata
         % positions
         qpsecs_cache = []
     end
-    
+
     properties(Access=private)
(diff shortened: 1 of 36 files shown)


horace_core/sqw/@rundatah/rundatah.m
```diff
@@ -157,10 +164,16 @@ function clear_det_cache()
             else
                 proj_mode = 2;
             end
-            % Calculate projections            
+            % Calculate projections
             [u_to_rlu,pix_range,pix] = obj.calc_projections_(obj.detdcn_cache,[],proj_mode);
         end
+        function flds = indepFields(obj)
+            flds = indepFields@rundata(obj);
+            flds = [flds(:)',{'transform_sqw'}];
```
oerc0122, line 172: Does the orientation of the fields actually matter? why not just
```matlab
flds = [flds, 'transform_sqw'];
```

abuts: Interesting. Apparently it looks like it works here. Probably it did not work in a previous Matlab versions

horace_core/sqw/@rundatah/rundatah.m
```diff
@@ -24,6 +24,7 @@ function clear_det_cache()
             % clear cached detectors information and detectors directions
             calc_or_restore_detdcn_([]);
         end
+        %
```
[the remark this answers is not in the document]
abuts: I like % as it separate methods when the class contents is collapsed

horace_core/sqw/@rundatah/rundatah.m
```diff
@@ -15,15 +15,17 @@
         % positions
         qpsecs_cache = []
     end
-    
+
     properties(Access=private)
         transform_sqw_f_=[];
     end
+
```
cmarooney-stfc, line 22: comment for whole file, not specifically this line.

Missing the classVersion method.

PR #790, opened 2022-03-18
Design for container for compressible instruments

cmarooney-stfc:
Design for container for compressible instruments

### DIFF
--- a/documentation/add/09_container_design_for_instruments.md
+++ b/documentation/add/09_container_design_for_instruments.md
@@ -1,0 +1,38 @@
+# Design for a container for compressed (unique instances of) instruments, samples or detectors within the experiment_info object in an SQW object
+
+### Background
+
+SQW objects store instrument, sample and detector instances for use with Tobyfit. Previously the instruments and samples were stored one per run in the header structure, while detector information was stored as a single struct in detpar. 
+
+For large numbers of multiple runs, if the instrument or sample was the same for all runs, the information would be duplicated may times within the SQW. When reduced-size copies of the SQW are made, e.g. as cuts, this duplication would increase.
+
+By contrast, the detpar information is a singleton within the SQW. It may be duplicated by making cuts, which would be undesirable - Toby has identified detector as the largest of our three categories and the one most needing removal of duplicates. At the same time, there are rare occasions when at least two detectors are used, one with some runs, one with the rest; this possibility is not catered for in the previous SQW structure.
+
+A start has been made on this by restructuring the header and detpar information into a single experiment_info object of class Experiment. The instruments are collected in one container, currently a cell array, as are samples, in a separate cell array. Both of these are currently one item per run, i.e. uncompressed. Detector information is stored in the container detector arrays, which is an array of IX_detector_array objects.
+
+This document will confine itself to instruments from this point, but the issue is general across all three categories. 
+
+To enable the instrument information to be compressed easily, we have decided to replace the instrument cell array with a container class designed by us in which the objects can easily be compressed.
+
+### Container design
+
+The current container for instruments is a cell array. Alternatives previously used and considered were mixin.Heterogeneous arrays. These allowed arrays to be used with the multiplicity of possible instrument sub-classes such as IX_inst_DGfermi and IX_inst_DGdisk. These stopped being used as they were apparently not working with serializable. Neither of these alternatives encapsulates the instrument collection such that the container's own methods can compress and index its contents.
+
+The new container needs to have 2 sets of properties:
+
+1. it must be able to add and remove elements such that addition will not add a duplicate of an instrument already in the container, and to index the unduplicated contents such that the correct instrument will be returned for a given run.
+2. it must be able to store heterogeneous objects such as IX_inst_DGfermi and IX_inst_DGdisk such that scans for duplicates are made only on objects of the same subclass.
+
+To meet the second condition, it is proposed that the container will internally have a struct with one element for each subclass type of instruments. That is, a struct of the form:
+
+````
+contents = struct();
+contents.IX_inst_DGfermi = IX_inst_DGfermi.empty;
+contents.IX_inst_DGdisk = IX_inst_DGdisk.empty
+```
+````
+
+The `classname`method can then be applied to the incoming object to determine which array, and hence which field of `contents`should be used to store it.
+
+Alternatively, a `container.Map` could be used in place of the struct. It is suggested that the serializability of `contents`is the important criterion for this choice, and that serializing struct will be more beneficial for code use than `container.Map` but that is a matter for discussion.
+


oerc0122:
I've made a PR into this branch to amend some of the markdown.


 In terms of comments I think we need to look at some form of hashing function to append to `serializable`, or a new `hashable` mixin, preferably the latter. 

Given `containers.Map` only accept limited types as keys, it thereby means that we want to be able to uniquely map our objects into hashes to have consistent reference in `containers.Map` following the hashing, the map should handle the rest of the tracking and such. 

The reasons for potentially attaching hashing to `serializable` might be:
- verification of a correct sending or data package, but this seems unlikely addition. 
- Using the existing `indep_fields` to collect data for hashing

It is very likely, however, that we don't necessarily want the same information hashable as we do serialized, as data can be reconstructed from potentially less or at the very least different information than a hash. 

The choice of hashing algorithm in completely unimportant at this point, and while MATLAB has no native hashing, it has direct access to Java which does and a custom hash is trivial to write.

mducle:
I'm not sure what the serialization routine currently in use is, but the [undocumented](https://undocumentedmatlab.com/articles/serializing-deserializing-matlab-data) `getByteStreamFromArray` is very useful. With this you can then use a bit of java to calculate an MD5 hash of the object:

```
Engine = java.security.MessageDigest.getInstance('MD5');
Engine.update(getByteStreamFromArray(obj));
hash = typecast(Engine.digest, 'uint8');
```

This could be your unique key to prevent duplication; it would not matter what the type of data is...

E.g.

```matlab
classdef unique_con < handle
    properties(Access=private)
        store_ = {};
        hash_ = [];
    end
    methods
        function [idx, hash] = is_in_container(self, object)
            % Returns the index in the container if the object is in it else empty
            Engine = java.security.MessageDigest.getInstance('MD5');
            Engine.update(getByteStreamFromArray(object));
            hash = typecast(Engine.digest, 'uint8')';
            if isempty(self.hash_)
                idx = [];
            else
                [~, ~, idx] = intersect(hash, self.hash_, 'rows');
            end
        end
        function self = add_object(self, object)
            [idx, hash] = self.is_in_container(object);
            if isempty(idx)
                self.hash_ = cat(1, self.hash_, hash);
                self.store_ = cat(1, self.store_, object);
            end
        end
    end
end
```

I've used a `cell` array here because it takes the least memory and works with the `uint8` hashes. If you want to use a `struct` or `containers.Map` you must convert the hash to a `char` array / string with `reshape(dec2hex(hash), 1, 32)` or similar, since only `char` arrays are valid fieldnames / key names in Matlab (a disadvantage compared to Python ;)

abuts:
After some thinking, would add more comments to Duc's (@mducle)  design which is a good starting point. 
it answers the questions (usage scenarios):
1) store  only unique objects in the container
2) partially answers the question of getting objects as function of run id. To fully answer this question, a kind of map run_id->position in the container should be maintained. 
---------
concerns:
are java hashes the same for the objects with equivalent contents? If they work and fast, it would be ok. 
As a whole, the container is the storage for a comparable objects. Would it be better to separate comparison operators and the container? Then it would be freedom to identify appropriate comparison operator on the particular class, and to use the container as generic storage.
---------
missing points (usage scenario):
How all this works with Tobyfit, is it convenient and is it efficient for obtaining what Tobyfit needs from instrument from this container? This question should be clarified with Toby (@tgperring) and may be Jacob (@oerc0122 )

oerc0122:
In partially answering your concerns: 

> Are java hashes the same for the objects with equivalent contents? 

It depends on how they're specified. It *should*, but we might find that for soem reason the byte stream includes (for example) the pointer to the data, which may be unique and then messes with the hash. It may be necessary to define a `hashable` mixin which provides the data to be hashed to uniquely identify an object. 

> If they work and fast, it would be ok.

In terms of speed, MD5 is widely used to determine whether files match and is sufficiently swift for that. Should it turn out it is not, we can look at alternative hashing schemes, but for now until we've determined it's performance critical (my guess is absolutely not given the cost of `gen_sqw`, I don't think we need to worry.

> As a whole, the container is the storage for a comparable objects. 

The objects don't have to be comparable, only the hashes if I'm understanding correctly.

> Would it be better to separate comparison operators and the container? 

I would say attaching the hashing algorithm to the container makes sense and means that the object represents *a* containerisation method. This means that should we decide to change the hashing method it is changed everywhere and the `container` contains *all* the information to store, recover and reproduce the hashes contained within it.

> Then it would be freedom to identify appropriate comparison operator on the particular class, and to use the container as generic storage.

The comparison happens on the hash.

> missing points (usage scenario):
> How all this works with Tobyfit, is it convenient and is it efficient for obtaining what Tobyfit needs from instrument from this container? This question should be clarified with Toby (@tgperring) and may be Jacob (@oerc0122 )

Given this is a cell array (in this instance) or some other array or container, provided the data can be extracted, these can easily be coerced into the correct form for Tobyfit data for minimal cost compared to that of computing Tobyfit.

abuts:
> > Would it be better to separate comparison operators and the container?
> 
> I would say attaching the hashing algorithm to the container makes sense and means that the object represents _a_ containerisation

I would rather attach comparison operator to class or set of classes for one reason. We may want to ignore some properties of an class at comparison. E.g. very often we compare sqw objects with option `ignore_string` to avoid comparing internal filenames. In other situation we may want to include such option. 

If a comparison should or should not include such a things would depend on the purpose, the classes are stored in the container.

abuts:
> Given this is a cell array (in this instance) or some other array or container, provided the data can be extracted, these can easily be coerced into the correct form for Tobyfit data for minimal cost compared to that of computing Tobyfit.

But cellarray of fully fledged objects occupy excessive memory, if you have them all expanded (e.g. full set of detectors for all runs). You should be able to extract your requested functor, without expanding the objects.

mducle:
Ok... I hacked together a quick class: https://gist.github.com/mducle/bc0e93d403ef9f30f605450335306dac

```matlab
li = let_instrument(5, 240, 80, 20, 1);
uc = unique_con; for ii = 1:100; uc(ii) = lm; end;
mi = merlin_instrument(180, 600, 'g');
uc = [uc unique_con(repmat({mm}, 1, 100))];
ss = struct(uc);
```

gives:

```
>> uc

uc = 

Unique container with 200 elements and 2 unique elements

>> whos
  Name            Size             Bytes  Class                                          Attributes

  ii              1x1                  8  double                                                   
  li              1x1              15230  IX_inst_DGdisk                                           
  mi              1x1               6698  IX_inst_DGfermi                                          
  ss              1x1              24272  struct                                                   
  uc              1x1                  8  unique_con                                               

>> uc(55)

ans = 

  IX_inst_DGdisk with properties:

     mod_shape_mono: [1×1 IX_mod_shape_mono]
          moderator: [1×1 IX_moderator]
    shaping_chopper: [1×1 IX_doubledisk_chopper]
       mono_chopper: [1×1 IX_doubledisk_chopper]
          horiz_div: [1×1 IX_divergence_profile]
           vert_div: [1×1 IX_divergence_profile]
             energy: 5
               name: 'LET'
             source: [1×1 IX_source]

>> uc(155)

ans = 

  IX_inst_DGfermi with properties:

        moderator: [1×1 IX_moderator]
         aperture: [1×1 IX_aperture]
    fermi_chopper: [1×1 IX_fermi_chopper]
           energy: 180
             name: 'MERLIN'
           source: [1×1 IX_source]
```

cmarooney-stfc:
Responding to @abuts comments above 21/3/22 14.30. Alex, I paste your comment here (less the initial markdown stuff) so I can comment inline.

AB: This document is highly incomplete 

CM: Yes indeed. As it has already generated a lot of comment, it would be silly to try and extend the document until replies are in. It appears to have been very succesful in generating thoughts, and as such has done its job. I think the word you want is "draft"

AB: and I would not agree with internal structure as the convenient way of containing objects. What would happen if you want to combine two Fermi chopper instruments e.g. MAPS and MARI runs? 

CM: The present (draft) document assumes that the instruments are considered as monolithic objects, rather than breaking them down into their component parts. I will deal with the alternative in a moment. The SQW object design has always been that each run has a defined instrument. Previously it was part of the struct in header for the relevant run number; now it is one of the instruments in the cell array instruments of the Experiment-class experiment_info. If it is required that runs from MAPS and MARI are combined into one SQW, then (say) runs 1-10 will be on MAPS and will have the MAPS instrument, and (say) runs 11-20 each have the MARI instrument.  At present these are uncompressed; if the instruments are compressed then the number of instruments will be reduced to 2, i.e. instruments{1}==MAPS, instruments{2}==MARI, and a separate instrument_index array will record instrument_index(1:10)==1, instrument_index(11-20)==2. All of this will be under the hood in the container object, and if the container is given the run number, it will return the relevant instrument.

    If it is preferred to break down the instrument by component, then the containers can be used component-wise, so a container will store Fermi-choppers, indexed and compressed in the same manner, and the instrument classes will be converted to reference such containers.

AB: How would you deal with combining runs obtained on the same instrument, which have changed, e.g. old LET and new LET, where the detector coverage have changed? 

CM: I am presuming that the old LET and new LET differ in their detector arrays from this description. In these descriptions I am using the shorthand "instrument" for "primary spectrometer" (as is already the case in the various IX_inst_* types in the code) and reserving "detector" for "secondary spectrometer" (as is already the case in the various IX_det* types in the code.) Currently and previously the SQW object contained only one detpar object and hence would presumably not cater for this scenario. The new Experiment-class experiment_info now contains an array of IX_detector_arrays, arbitrarily indexed and currently defaulting to element 1 for all runs. In this case I presume that the old and new LETs will each define a separate IX_detector_array, with detector_index values for each run pointing to detector_array(1) or detector_array(2) as appropriate. As with the instruments, the new detector array container could deal with this. However, as there is only one class type IX_detector_array and it is implemented as a Matlab array, that is currently not necessary, with a simple wrapper sufficing  to do the indexing.

AB: How would the internal structure of the container help you to achieve this purpose? 

CM: See previous replies.

AB: What would happen if we want to combine inelastic X-ray scattering instrument from Diamond (and have written its description) and one of our instruments. 

CM: This is a very interesting extension. I am as yet unaware that it was proposed to use Horace for X-ray scattering as well as for neutron scattering, and would be grateful if you would point me to the design documents for this. As a particular issue I cannot visualise how the scattering ranges (r-theta-phi) would differ between X-ray and neutron, and it is more likely that there would be overlap between the ranges. Hence I would expect Horace to have to deal with this overlap. I would also want to know the extent to which the modelling codes (Euphonic, SpinW etc) would need to cater for different measurement techniques. Again, looking at a combined method design document would be useful.

AB: The described container is completely useless.

CM: This seems a very blanket comment. In the light of the above replies, you may wish to reconsider.

AB:    Usage scenarios.
    While common purpose of the container (keep instrument) is stated, its unclear why you would keep it?

    1a) I would assume that one of the purposes would be to extract instruments contributed to a run.

CM: Yes, indeed as far as I know the only purpose, beyond encapsulating instrument compression.

AB: So, when you do cut, you return objects contributed to the cut
I.e. you would have method, which return sub-container as the function of the run_id-s contributed to the cut.
How the structure would help to achieve this purpose?

CM: Let's start with the sensible way of implementing this, which is to have instruments as handle classes. Then the duplicated container in the cut would only contain pointers to the instruments - no duplication, perfect compression. But copy-on-write will achieve the same effect; the duplicated container will not actually duplicate the instrument.

      The additional scenario is that when everything is written to file, restoration from file will break the links (either handle or CoW) between cut and original SQW. I had agreed with Toby that the actual container of unique objects would sit outside the SQW objects and be restored from file separately, with only the indices retained in the SQW object; hence the container object's functions of storage and indexing would be split.

AB: 1b) I know that Tobyfit uses some characteristics of the instrument. I am not sure what exactly it uses, but seen something, again, obtained as the function of a run-id. Which methods the container should have to support this and how it would provide for these methods? Would be these methods fast enough and occupy reasonable memory? (e.g. PDF as the function of run id)

CM: The aim is to reference the unique instrument. Once it has been obtained, its use should be independent of the containment. The actual referencing - get run index, get instrument index, operates on integer indexes only and should not cause problems.

AB: I think these questions should be answered or at least thought about before coding begins.

CM: Agreed. Your responses to these points is appreciated. Thanks, Chris

PR #802, opened 2022-05-16
Explicit type conversion in `fwrite`s

mducle:
Recently we saw a large slow down (factor of >~100x) in speed of writing `sqw` files in the IDAaaS system which uses the CEPH object store under a network filesystem abstraction layer. Previously (in April) the system was working as fast as the VM-local disks.

After investigations it was found that at least a factor of 10x slow-down was due to the type conversion in the call to `fwrite`. Specifically, Horace stores the signal, error and number of pixel data as double-precision floats in memory _but_ writes these out as single-precision floats and long integers. This implicit conversion, i.e. `fwrite(fout, pix, 'float32')` was causing a large slow down compared to the explicit - `fwrite(fout, single(pix), 'float32')` versions.

This PR changes all applicable calls in Horace to explicitly convert the data before passing to `fwrite`.

With these changes, running on CEPH is only around ~5-10 slower than using the VM-local storage (which is of limited size).


abuts (approved):
Matlab is a strange beast. 
This change apparently never makes things worse but often make them better. 
even on other operating systems.
This means -- good change to have

PR #815, opened 2022-06-16
448 fix mpi threads

oerc0122:
Fix spawning too many threads on MPI program start by adding new `par_threads` attribute to `parallel_config` which is read upon creating workers. 

Also moves sub-process spawning to `start_job` to unify poor-man and mpi startup.

Moves threading options from `hor_config` to `parallel_config`.


horace_core/horace_memory.m
```diff
@@ -16,22 +16,23 @@
 %       >> set(hor_config,'mem_chunk_size',chunk_size,'threads',threads);
 %       >> [mem.chunk_size,mem.threads]=get(hor_config,'mem_chunk_size','threads');
 
+elseif nargin ~= 0 && nargin ~= 2
```
cmarooney-stfc, line 19: This appears to be missing an if.

cmarooney-stfc (changes requested):
In addition to the individual comments below, it would be good to say in the PR comment what the fix is. It's more than moving data between configs

Issue #821, opened 2022-07-07
Relative (root) paths can be sanitised


oerc0122:
There are places in the code which are looking for e.g. the root path (the folder containing `(herbert|horace)_core`) or the `(herbert|horace)_core`s respectively. Currently, this is done (where necessary) through `fileparts(fileparts(which('herbert_init')))` this is not only confusing code, but also unnecessarily complex. 

To solve this we can add global variables `herbert_path`, `horace_path` and `root_path` to be initialised once in `herbert_init` and `horace_init` respectively (and likewise unset in `horace_off`) and reference these as needed.

oerc0122:
1. This is a bad habit.
2. Does this not clear other horace/herbert vars?
3. Is there a way to rectify this? I suppose I could attach it to the `horace|herbert_config`

mducle:
Yes, it's a bad habit but people _will_ use it. It clears all the variables but the Horace paths are still in memory so things like `cut_sqw` will still work. It's just that `ClusterMPI` (e.g. `gensqw` or configuring the parallel framework) and  `horace_off` will now fail...

> Is there a way to rectify this with extra persistence?

Not really. `clear all` really does clear everything.

The workarounds I can think of are:

* Write the path variables to a temporary file.
* Wrap the path variables in a small helper class which can still use the globals but if it detects the globals have been cleared will call `which('horace_init')` / `which('herbert_init')` to reset them (the path is not cleared)

PR #822, opened 2022-07-07
Sanitise relative paths

oerc0122:
See #821 for detailed info

Adds 3 global variables `herbert_path`, `horace_path` and `root_path` which point to relevant parts of the Horace directory structure for internal reference and consistency. 

These are accessible via the `horace_paths` class (in `herbert_core`) there are also methods to get the folder of a function and clearing `paths` as well as the base paths (admin, tests, lowlevelcode) and two commonly used paths (test/common_data, test/common_functions)

These are set in `herbert_init` and `horace_init` or through the `paths` class and cleared in `horace_off`.

Also fixes typo on startup banner.

### DIFF
--- a/horace_core/horace_off.m
+++ b/horace_core/horace_off.m
@@ -8,7 +8,7 @@ function horace_off
 
 % root directory is assumed to be that in which this function resides
 
-pths = paths;
+pths = horace_paths;
 on_path = pths.get_folder('horace_on');
 
 warn_state=warning('off','all');    % turn of warnings (so don't get errors if remove non-existent paths)
--- a/horace_core/horace_off.m
+++ b/horace_core/horace_off.m
@@ -19,9 +19,14 @@ try
     end
     rmpath(paths);
     warning(warn_state);    % return warnings to initial state
+    clear global horace_path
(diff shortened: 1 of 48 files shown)


mducle (approved):
Thanks for changing it to a class!

horace_core/horace_off.m
```diff
@@ -7,22 +7,26 @@
 % T.G.Perring
 
 % root directory is assumed to be that in which this function resides
-rootpath = fileparts(fileparts(which('horace_init')));
+global horace_path
```
mducle, line 10: Oh actually... should this not use the `paths` class?

oerc0122: Done.

mducle (approved):

abuts (approved):
well, do not mind this, though not see big benefit. But let it go -- why not. 
Two notes. 
1) `paths` class sounds very similar to system `path` -- I would rename it to `hor_paths` to avoid confusion or somebody in a future in a toolbox use similar name. 

2) It looks like this does not need a unit test, but I would suggest opposite. Its not difficult to add one

PR #830, opened 2022-07-27
horace on fixed

abuts:
fix the horace_on script allowing to enable herbert, which correspond to another Horace


cmarooney-stfc (approved):

admin/horace_on.m.template
```diff
@@ -30,27 +41,36 @@ if ~isempty(sw_start)
     spinw_on(default_spinw_path);
 end
 
+if nargin > 1
+    independent_herbert = true;
+else
+    independent_herbert = false;
+end
```
oerc0122, line 48: Why not 
```MATLAB
independent_herbert = exist('non_default_her_path', 'var') && ~isempty(non_default_her_path);
```
(or equivalent with if) which both makes intent clear and protects against later modification.

admin/horace_on.m.template
```diff
@@ -1,8 +1,13 @@
...
 %  safely switches Horace on
 %  horace_on()                         -- calls Horace with default settings
 %  horace_on(non_default_horace_path)  -- calls Horace with non-default Horace folder;
-%
+%  horace_on(non_default_horace_path,non_default_her_path)
+%                                      -- calls Horace with non-default Horace folder
+%                                         and non-default Herbert folder,
+%                                         not located within non-default
+%                                         Horace folder (for old versions
+%                                         of Horace/Herbert pairs)
```
oerc0122, line 10: The wording here is confusing
```suggestion
%                                      -- calls Horace with non-default Horace folder
%                                         and non-default Herbert folder,
%                                         for Herbert external to current
%                                         Horace folder (for old versions
%                                         of Horace/Herbert pairs). If empty
%                                         use default.
```

oerc0122: I would also consider changing `calls` to `loads`

admin/horace_on.m.template
```diff
@@ -12,9 +17,15 @@ default_horace_path  = '${Horace_CORE}';
 % with horace_on.m script.
 default_spinw_path = '/usr/local/mprogs/spinW';
 %
-if exist('non_default_path','var') && (strcmpi(non_default_path,'where') || strcmpi(non_default_path,'which'))
-    path = find_default_hor_path(default_horace_path);
-    return;
+if exist('non_default_hor_path','var')
+    if (strcmpi(non_default_hor_path,'where') || strcmpi(non_default_hor_path,'which'))
```
oerc0122, line 21: Would also check for `isempty` here

admin/horace_on.m.template
```diff
@@ -30,27 +41,36 @@ if ~isempty(sw_start)
...
+        init_herbert(herbert_path);
     catch ME
         disp('HORACE_ON:wrong_dependencies - cannot initiate Herbert');
         rethrow(ME);
     end
 else % reinitialize Herbert on where it is now.
-    her_path = fileparts(which('herbert_init.m'));
-    init_herbert(her_path);
+    if independent_herbert
+        init_herbert(non_default_her_path);
```
oerc0122, line 65: Why not move the `if independent_herbert` block outside the `if ~herbert_initiated` block and use `herbert_path` set there?

admin/horace_on.m.template
```diff
@@ -70,7 +90,7 @@ function her_path = build_default_her_path(hor_path)
 if strcmp(hor_folder,'horace_core')
     her_path = fullfile(fp,'herbert_core');
 else
-    her_path = fullfile(hor_path,'herbert_core');    
+    her_path = fullfile(hor_path,'herbert_core');
```
oerc0122, line 93: If this is an external herbert, we don't want `her_path` defined from `hor_path`, it should be taken from where `herbert_init` is.

oerc0122 (changes requested):

admin/horace_on.m.template
abuts: I prefer avoiding checking for input variable existence, as in fact it only checks if the position of the variable is filled. 
I.e. if you have `my_fun(a,b)` and call it `my_fun(b)` the check inside `my_fun`  `exist('b','var')` would return `false`. So the check gives you false feeling that you are checking for variable existence when in fact you are checking for `nargin`. `isempty` can be added -- no harm.

oerc0122: This reasoning doesn't make sense to me. You are using positional arguments, so of course it only checks the position. Within the function the name of the outer variable doesn't mean anything. You are checking exactly what you asked for. If you want to use keyword arguments, then you can use keyword arguments within `inputParser` and subsequently check for presence.

```matlab
p = inputParser();
addRequired(p,'a');
addParameter(p,'b', []);
parse(p,a, b);

if p.Results.usingDefaults.b
   ...
end
```
But there are complaints that this is slow in performance critical functions (while I don't think this is).

Checking nargin adds a level of abstraction in determining which argument is present or not, reduces extensibility and maintainability for other programmers and in fact, with newer (2022a) MATLAB keyword syntax what you are checking is fundamentally wrong.
```matlab
my_fun(b='b')
% nargin -> 1
% b -> exists
```

admin/horace_on.m.template
abuts: because `independent_herbert` is used in both Herbert branches.

admin/horace_on.m.template
abuts: its exactly what the code doing. Its not building Herbert path from Horace if Herbert is independent

cmarooney-stfc:
as this has been merged, moving to Done

PR #831, opened 2022-07-29
test_fake_sqw_bin_pix use skipTest function

oerc0122:
Minor fix to make `test_fake_sqw_bin_pix` use `skipTest`

### DIFF
--- a/_test/test_mex_nomex/test_fake_sqw_bin_pix.m
+++ b/_test/test_mex_nomex/test_fake_sqw_bin_pix.m
@@ -3,7 +3,7 @@ classdef test_fake_sqw_bin_pix < TestCase
 
     properties
         tmp_data_folder;
-        skip_this_test=false;
+        skip_tests;
         current_mex_state;
         current_thread_state;
         sample_dir;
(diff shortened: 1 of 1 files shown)


_test/test_mex_nomex/test_fake_sqw_bin_pix.m
```diff
@@ -18,23 +18,23 @@
...
             this.current_thread_state = hc.threads;
             cleanup_obj=onCleanup(@()set(hor_config,'use_mex',this.current_mex_state));
 
             hc.use_mex = 1;
-            if hc.use_mex ~= 1
-                this.skip_this_test= true;
-            end
+
+            this.skip_tests = ~hc.use_mex;
+
```
cmarooney-stfc, line 29: As the code stands, this sets this.skip_tests to 0. Presuming the idea is that you can change hc.use_mex in L.26 and then this one follows on automatically, so L.26 is a menu. Comment it so that this is explicit.

cmarooney-stfc (changes requested):

_test/test_mex_nomex/test_fake_sqw_bin_pix.m
oerc0122: I believe the idea is that setting `use_mex` runs tests to verify `mex` can be used.
```MATLAB
        function this = set.use_mex(this,val)
            if val>0
                use = true;
            else
                use = false;
            end
            if use
                % Configure mex usage
                % --------------------
                [~, n_errors, can_combine_with_mex] = check_horace_mex();
                if n_errors>0
                    use = false;
                    warning('HOR_CONFIG:set_use_mex',...
                        ' mex files can not be initiated, Use mex set to false');
                end
                if ~can_combine_with_mex
                    config_store.instance().store_config(this,'combine_sqw_using','matlab');
                end
                
            end
            config_store.instance().store_config(this,'use_mex',use);
            
        end
```

PR #834, opened 2022-08-02
Hotfix for paths

oerc0122:
`horace_paths` did not get paths correctly if cleared.


cmarooney-stfc (commented):
Noting I've seen this - no further action as it's a hotfix and merged

PR #855, opened 2022-09-12
Fix issues with header datetime

oerc0122:
Do some cleanup and fix potential issues with tests falling on minute boundary


horace_core/sqw/@main_header_cl/main_header_cl.m
```diff
@@ -130,50 +147,89 @@
...
         function nf = get.nfiles(obj)
             nf = obj.nfiles_;
         end
+
         function obj = set.nfiles(obj,val)
-            if ~isnumeric(val) || val<0
+            if ~isnumeric(val) || val<0 || ~isscalar(val)
                 error('HORACE:main_header:invalid_argument', ...
-                    'Number of files should have non-negative numeric value. It is %s',...
+                    'Number of files should have a single non-negative numeric value. It is %s',...
```
cmarooney-stfc, line 171: This is a bit unclear. Prefer "number of file should BE a single non-negative numeric value. Instead it is BEING SET TO %s...."

horace_core/sqw/@main_header_cl/main_header_cl.m
```diff
@@ -130,50 +147,89 @@
...
         function obj = set.creation_date(obj,val)
             % explicitly set up creation date and make it "known"
-            obj = set_creation_date_(obj,val);
+            if ischar(val)
+                dt = obj.convert_datetime_from_str(val);
+            elseif isa(val,'datetime')
+                dt  = val;
+            else
+                error('HORACE:main_header:invalid_argument', ...
+                      'file creation date may be datetime class or string, transferrable to datetime function according to format %s. Provided %s', ...
```
cmarooney-stfc, line 210: Prefer "creation date MUST be datetime .... INSTEAD %s IS PROVIDED

horace_core/sqw/@main_header_cl/main_header_cl.m
```diff
@@ -197,48 +254,49 @@
...
+
     methods(Static)
         % Service routines:
         function datt = convert_datetime_from_str(in_str)
             % convert main_header_cl's generated time string into
-            % datetime format (standard datetime conversion works very
-            % strangely)
-            val = num2cell(sscanf(in_str,main_header_cl.DT_format_));
-            datt = datetime(val{:});
+            % datetime format
```
cmarooney-stfc, line 278: Remembering your comment on this and appreciating that the original strangely comment is opaque. But we have now lost a warning that something may go wrong if the standard conversion is used. Maybe stating what the standard conversion is and advising it has had previous problems?

cmarooney-stfc (changes requested):

horace_core/sqw/@main_header_cl/main_header_cl.m
oerc0122: This is an `error`, so it is not being set to anything, but ignored.

abuts (approved):
Do not see the changes fixing any bugs but some minor improvements are there so why not

oerc0122:
@abuts The fix is that by using actual `datetime` rather than the custom from format method, it resolves `16:15:60` correctly to `16:16:00`

abuts:
then it should be unit test for that, but ok

PR #877, opened 2022-10-20
Update Horace_GUI.rst

russellewings:
Trivial change of word "which" to "that", ostensibly to test that the editing button works!


mducle:
Looks good to me.

PR #884, opened 2022-10-31
Misc fixes which means slurm commands now work

oerc0122:


abuts (commented):
Difficult to judge -- all I need to see unit tests, though it may be difficult to do for cluster

PR #887, opened 2022-10-31
deprecated projaxes

abuts:
added projaxes wrapper which refers to ortho_proj and generates deprecation warning


mducle (approved):
Looks good to me!

Issue #885, opened 2022-10-31
enable crystal alignment


abuts:
disabled by Tessela and dependent on cut. Should be fixed

Also may be related and be fixing #644 and #580

abuts:
fixed a while ago. Part of #892

PR #890, opened 2022-11-03
put back erroneously removed signal function

abuts:
and modified it according to the recent design


horace_core/sqw/@sqw/sqw.m
```diff
@@ -77,28 +75,26 @@
         w = sigvar_set(win, sigvar_obj);
         sz = sigvar_size(w);
         %------------------------------------------------------------------
+        wout=signal(w,name)  % Set the intensity of an sqw object to the
+        %                    % values for the named argument
```
cmarooney-stfc, line 79: Is it better to call this set_signal? SIgnal sounds like it is returning the signal values.

horace_core/sqw/@sqw/signal.m
```diff
@@ -0,0 +1,46 @@
+function wout=signal(w,name)
+% Set the intensity of an sqw object to the values for the named argument
```
cmarooney-stfc, line 2: as noted in sqw.m, this sounds like a function that returns a signal. Set_signal may be more appropriate

From the argument description below, it looks as if this is a special use to allow the input parameters to be visualised. Saying what this usage is explicitly would be helpful

horace_core/sqw/@sqw/private/coordinates_calc_.m
```diff
@@ -0,0 +1,155 @@
...
+%               {'h'}
+%               {'d1','d2'}
+%               {'d3','E','Q','k'}
+%
+% Output:
+% -------
+%   ok      true if all OK; false if invalid coordinate name given
+%
+%   mess    empty if all OK, error message otherwise
+%
```
cmarooney-stfc, line 28: ok and mess are no longer outputs

horace_core/sqw/@sqw/private/coordinates_calc_.m
```diff
@@ -0,0 +1,155 @@
...
+xpix=cell(1,numel(xlist));
+if ind.h||ind.k||ind.l||ind.Q
+    header_ave=w.header_average();   % get average header
+    this_proj = w.data.proj;
+    hkl_proj = ortho_proj([1,0,0],[0,1,0], ...
+        'alatt',this_proj.alatt,'angdeg',this_proj.angdeg);
+    % TODO: this is compartibility function. It will change when alginment matrix
+    % is attached to pixels. In fact, it redefines b-matrix (and partially U-matix
+    % used for alignment), which is the function of lattice. See ticket #885
+    hkl_proj = hkl_proj.set_ub_inv_compat(header_ave.u_to_rlu(1:3,1:3));
```
cmarooney-stfc, line 97: see other comment in this review

horace_core/sqw/@sqw/private/coordinates_calc_.m
```diff
@@ -0,0 +1,155 @@
...
+
+
+    % Matrix and translation to convert from pixel coords to hkl
+    % Example of the code to use dealing with #825
+    % uhkl=header_ave.u_to_rlu(1:3,1:3)*w.pix.q_coordinates+repmat(header_ave.uoffset(1:3),[1,npixtot]);
+    uhkl = hkl_proj.transform_pix_to_img(w.pix.q_coordinates);    
+    if ind.Q
+        % Get |Q| -- We would use pix coordinates directly, but the pix
+        % coordinates may be invalid due to alignment. See #885 to resolve
```
cmarooney-stfc, line 106: need more info in #885

cmarooney-stfc (changes requested):

horace_core/sqw/@SQWDnDBase/SQWDnDBase.m
```diff
@@ -31,6 +31,8 @@
         w                 = sigvar_set(win, sigvar_obj);
         sz                = sigvar_size(w);
         %------------------------------------------------------------------
+        wout=signal(w,name)  % Set the intensity of an sqw object to the
+        %                    % values for the named argument
```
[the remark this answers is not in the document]
abuts: discussed above. Thinking about this a bit more, I do not see reason why it would not work on `DnD`. It just have not been implemented

horace_core/sqw/@sqw/sqw.m
mducle: The method name should remain `signal` for backwards compatibility. Also, it _does_ return the signal, but as a new `sqw` object rather than as a vector. The docstring here should probably say `% Returns a new sqw object with its intensity set to the values of the named argument` - e.g. we are not setting the intensity of `w`, but rather the intensity of a new object `wout`.

horace_core/sqw/@sqw/cut.m
```diff
@@ -117,7 +118,7 @@
 targ_proj.angdeg = header_av.angdeg;
 % TODO: this is compartibility function. It will change when alginment matrix
 % is attached to pixels. In fact, it redefines b-matrix (and partially U-matix
-% used for alignment), which is the function of lattice
+% used for alignment), which is the function of lattice. See ticket #885
```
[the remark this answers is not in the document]
abuts: The same peace of code is in signal too, so I've modified it here too. Wording may be done better.

horace_core/sqw/@sqw/private/average_bin_data.m
```diff
@@ -10,7 +10,7 @@
 %
 %   vals    array, or cell array of arrays, each with npixtot elements, to be
 %          averaged over the bins of w. That is, the first w.data.npix(1) values
-%          are averaged to give one element, the next w.data.npix(2) averaged for
+%          are averaged to give one element, the next`w.data.npix(2) averaged for
```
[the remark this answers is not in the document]
abuts: yes, modified above

horace_core/sqw/@sqw/private/coordinates_calc_.m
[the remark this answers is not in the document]
abuts: comment left from previous version of the code

abuts: done

horace_core/sqw/@sqw/signal.m
mducle: Yes, it can be used to visualise the input parameters, but actually, its main use is for corrections. For example, the intensity of magnetic excitations are often approximately proportional to the energy transfer, so that high energy excitations can appear much fainter. So, sometimes you might want to analyse S(Q,E)/E rather than S - at the moment (because this function was removed from Horace 3.6 but was in 3.5) you have to do something `wcorr = w / sqw_eval(w, @(h,k,l,e) e)` - whereas before you could do `wcorr = w / w.signal('E')` - or to be more correct you would use the Boltzmann thermal occupation factor `wcorr = w / exp(-w.signal('E')/temperature)`. For phonon signals, you might want to correct for the `Q^2` dependence with e.g. `wcorr = w / w.signal('Q').^2`

abuts: As we are recovering old function, mentioned elsewhere and used in other peoples code I can not easy change its name. 
A separate ticket deprecating `signal` and introducing `set_signal` would be easy to write and implement if the stakeholders approve it.

horace_core/sqw/@sqw/sqw.m
abuts: see answer above

abuts:
> 

not really many independent changes. There some typos fixed and I see them and some code blocks rearranged for better visibility. Is not worth see a separate ticket for this.

mducle (approved):

PR #895, opened 2022-11-08
Add code to use unique_objects_containers to store instruments and samples without duplicates.

cmarooney-stfc:


_test/test_algorithms/test_cut.m
```diff
@@ -58,25 +59,6 @@
...
-            old_conf = conf.get_data_to_store();
-            conf.mem_chunk_size = 8000;
-            cleanup = onCleanup(@() set(hor_config, old_conf));
-
-            sqw_cut = cut(obj.sqw_file, obj.ref_params{:});
-
-            ref_sqw = read_sqw(obj.ref_cut_file);
-            assertEqualToTol(sqw_cut, ref_sqw, 1e-5, ...
-                'ignore_str', true,'-ignore_date');
-        end
```
abuts, line 78: why this have disappeared? It is important test for essential functionality

_test/test_experiment/test_experiment_methods.m
```diff
@@ -90,7 +90,7 @@ function test_to_from_old_structure_single_head(obj)
             assertEqual(rec_exp.runid_map.values,{1});
 
             % these properties are only partially recovered from headers
-            assertEqual(rec_exp.samples{1},IX_samp('',[1.1,2.2,3.2],[90,91,92]));
+            assertEqual(rec_exp.samples{1},IX_null_sample('',[1.1,2.2,3.2],[90,91,92]));
             % instrument is not recovered from headers
```
abuts, line 94: `IX_samp` contains lattice. `IX_null_sample` contains no useful information. Have this convention changed and what does this test change mean?

_test/test_sqw_file/test_instrument_methods.m
```diff
@@ -109,8 +109,9 @@ function test_set_instrument(obj)
             ldr1.delete(); % clear existing loader not to hold test file in case of further modifications
             
             assertEqual(numel(inst),1) % all instruments for this file are the same
+            inst = inst{1}; % unpack the single instrument in the cellarray
             hdr = wtmp.experiment_info;
-            assertEqual(hdr.instruments(186),inst);
+            assertEqual(hdr.instruments{186},inst); % compare the last instrument in the header with the single instrument
             
             %---------------------------------------------------------------------
             % NOT IMPLEMENTED or implemented wrongly. Does not accept array
```
abuts, line 117: Is this still the case? I believe the purpose of the work was to fix this.

_test/test_algorithms/test_cut.m
cmarooney-stfc: Sorry, it got shoved down to the end of the file while fixing a conflict. Now back in original position. Thanks

_test/test_experiment/test_experiment_methods.m
abuts: It is some misunderstanding here. `IX_samp` and `IX_inst` is not abstract and I remember instantiating `IX_samp` when reading binary sqw files v2 (always) and v3 -- without proper sample defined. It is used in transformations and alignment. I am opened its code and it is not abstract. Not sure about `IX_inst`. It is not used without instrument defined (so full fledged v3 `sqw`), so it may be abstract if no instrument is available. Not sure what is actually in the code. 

your definition defines the idea of `Null_sample` If it is null, you can not use it.

abuts (approved):
ok, It does what is says, do I do not mind having it. I have voiced my objection about other things found here, but I do not think it worth spending time on addressing them.  

One thing I would mention is: 
Does it really necessary to keep `n_duplicates_` array? it easy recovered from `idx_` if necessary but I would not see much need for it to store it independently.

PR #909, opened 2022-11-22
Remove printing in parallel calls

oerc0122:
Remove (comment in first instance) printing which can overflow the output buffer and cause Java executors to freeze waiting for output buffer to be cleared.


cmarooney-stfc (approved):

abuts (commented):
I do not know. These rows were always there and have never caused any problem. 
If removing them sorts the issue -- I do not mind, if not -- its useless and complicates debugging operations

PR #915, opened 2022-11-25
Add loading instrument info to loader_nxspe

mducle:
This PR adds the ability for Horace (the `loader_nxspe` function) to load instrument information (from the "primary spectrometer" up to but not including the sample and **not detector information**) from `nxspe` files which has certain fields.

Example usage:

```matlab
gen_sqw({fullfile(horace_path().test_common, 'inst_maps_ei400_600hz.nxspe'}, '', 'zzz.sqw', ...
        400, 1, [1 1 1], [1 1 1], [1 0 0], [0 1 0], 0, 0, 0, 0, 0);
w1 = read_sqw('zzz.sqw'); 
maps_inst = w1.experiment_info.instruments{1}
```

The output should be the same as that produced by:

```matlab
maps_inst = maps_instrument(400, 600, 's', '-version', 2, '-moderator', 'base2016')
```

You can generate the required `nxspe` files using eniius (`pip install eniius`) in Python:

```python
import eniius
eniius.Eniius(eniius.horace.maps_instrument(ei=400, freq=600, chopper='s')).to_nxspe('maps.nxspe')
```

Eniius always generates "version 2" instruments with "base2016" moderator terms for MAPS and MERLIN and empirical moderators for LET (this behaviour may change in future).


herbert_core/classes/data_loaders/@loader_nxspe/loader_nxspe.m
```diff
@@ -19,7 +19,8 @@
         % the structure, containing the folder structure of the nxspe file
         % as defined in hdf5 file
         nexus_dataset_info_ = [];
-        %
+        % a structure for instrument information
+        nexus_instrument_ = [];
```
cmarooney-stfc, line 23: See overall review comment  - this line comment is a placeholder

cmarooney-stfc (changes requested):
Checking if this change also covers samples and detectors - and if other PRs will cover these if not

thanks

herbert_core/utilities/hdf_nexus/read_nexus_groups_recursive.m
```diff
@@ -0,0 +1,45 @@
...
+    %   hdf_info_struct - the result of h5info on the required group
+    %   filename (string) - the NeXus file name
+    %   nexus_path (string) - the NeXus path to the required group.
+    %   group_struct - the group structure of part of a h5info output
+    %
+    if nargin == 1 && isstruct(filename) && isfield(filename, 'Filename')
+        dinfo = filename;
+        filename = dinfo.Filename;
+    elseif isstring(filename) || ischar(filename)
+        if isstring(nexus_path) || ischar(nexus_path)
```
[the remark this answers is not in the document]
mducle: No sure I understand... `nexus_path` must be a string/char if we use it with `h5info` (which produces a struct). Otherwise I'm assuming it's already a struct. If it's neither then there will be an error later on (in `read_nexus_datasets`).

herbert_core/utilities/hdf_nexus/read_nexus_groups_recursive.m
```diff
@@ -0,0 +1,45 @@
+function datastruct = read_nexus_groups_recursive(filename, nexus_path)
+    % Recursively reads a NeXus file group and returns a structure of the data
+    %
+    % datastruct = read_nexus_group_recursive(hdf_info_struct)
+    % datastruct = read_nexus_group_recursive(filename, nexus_path)
+    % datastruct = read_nexus_group_recursive(filename, group_struct)
+    %
+    % Inputs:
+    %   hdf_info_struct - the result of h5info on the required group
+    %   filename (string) - the NeXus file name
```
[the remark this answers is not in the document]
mducle: Well... the first argument can be struct `hdf_info_struct` or a string `filename`. The help text is meant to reference the possible syntax, in lines 4-6 so in this case "filename" must be a string.

herbert_core/classes/data_loaders/@loader_nxspe/loader_nxspe.m
mducle: I don't understand your comment "this line comment is a placeholder"?

This change does not cover detectors or samples. Just the "primary spectrometer" (all instrument components before the sample).

herbert_core/utilities/hdf_nexus/read_nexus_groups_recursive.m
```diff
@@ -0,0 +1,45 @@
...
+    for ii = 1:numel(dinfo.Groups)
+        pathfields = split(dinfo.Groups(ii).Name, '/');
+        datastruct.(pathfields{end}) = read_nexus_groups_recursive(filename, dinfo.Groups(ii));
+    end
+end
+
+function datasets = read_nexus_datasets(filename, nexus_struct)
+    datasets = struct();
+    for ii = 1:numel(nexus_struct.Datasets)
+        nexus_path = [nexus_struct.Name '/' nexus_struct.Datasets(ii).Name];
```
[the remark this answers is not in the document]
mducle: This is a NeXus path not a filesystem path. It must always be separated by a `/`.

mducle:
@oerc0122 @cmarooney-stfc  Could you review this when you have time, please?

herbert_core/classes/data_loaders/@loader_nxspe/loader_nxspe.m
cmarooney-stfc: The words "this comment is a placeholder" means that I want to draw attention to something for the whole PR, but it is easy to miss the whole-review comment and so I put a comment here to send you to the whole-PR comment. It is not a comment on line 23. 

Your comment that this change covers only the primary spectrometer is good but it could usefully go in the PR description as well.

herbert_core/classes/data_loaders/@loader_nxspe/loader_nxspe.m
```diff
@@ -207,11 +218,20 @@
...
+            end
+        end
+        function rv = has_loaded_instrument(obj)
+            rv = ~isempty(obj.nexus_instrument_);
+        end
+        function instrument = get_instrument(obj)
+            if ~isempty(obj.nexus_instrument_)
+                instrument = obj.nexus_instrument_;
+            else
+                instrument = self.read_instrument_info_();
```
cmarooney-stfc, line 232:
```suggestion
                instrument = obj.read_instrument_info_();
```

herbert_core/classes/data_loaders/@loader_nxspe/loader_nxspe.m
```diff
@@ -245,6 +265,90 @@
...
+            moderator = obj.read_inst_moderator_(dataset);
+            % NXSPE files *must* have a "fermi" component, so we distinguish
+            % instrument types based on presence of 'shaping_chopper' and 'mono_chopper'
+            % as we only have two types of instruments supported at present
+            errmsg = ['nxspe file has incomplete instrument data. Please manually set ' ...
+                       'instrument data if you want to perform resolution convolution.'];
+            if all(isfield(dataset, {'shaping_chopper', 'mono_chopper'}))
+                % The struct must have also have 'horiz_div', and 'vert_div' fields
+                if ~any(isfield(dataset, {'horiz_div', 'vert_div'}))
+                    error('HERBERT:loader_nxspe:missing_instrument_fields', errmsg);
```
cmarooney-stfc, line 296: the use of errmsg here means we cannot distinguish the various errors

herbert_core/utilities/hdf_nexus/read_nexus_groups_recursive.m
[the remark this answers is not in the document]
cmarooney-stfc: may be clearer to rename the first arg as e.g. arg1 and then the switch sorts out its meanings

cmarooney-stfc (changes requested):

herbert_core/classes/data_loaders/@loader_nxspe/loader_nxspe.m
mducle: Ummm I guess you're objecting to their being two error messages with the same identifier and text? (In lines 296 and 302)... what about the suggestion below?

(you comment was a bit unclear...)

herbert_core/classes/data_loaders/@loader_nxspe/loader_nxspe.m
```diff
@@ -245,6 +265,90 @@
...
+                end
+                instrument = obj.read_disk_inst_(dataset, source, moderator);
+            else
+                % The struct must have also an 'aperture' field
+                if ~isfield(dataset, 'aperture')
+                    error('HERBERT:loader_nxspe:missing_instrument_fields', errmsg);
+                end
```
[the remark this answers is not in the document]
mducle: Yes, I think the previous logic was a bit convoluted (I have not idea what I was thinking then). The requirements for an `nxspe` file to have valid instrument information as far as Horace is concerned is that:

* All files must have a `moderator` and `source`.
* Disk chopper instruments must have `shaping_chopper`, `mono_chopper`, `horiz_div` and `vert_div` (this will create an `IX_inst_DGdisk`
* Fermi chopper instruments must have `aperture` (this will create an `IX_inst_DGfermi`.

Previously I was looking at the presence of the two choppers to check if the user wanted to define a disk or Fermi chopper instrument and then erroring if they did not specify the divergences but really I should have just checked for the presence of all necessary fields. So now it only fails if it does not find either of the conditions needed for either a disk or Fermi instrument.

PR #926, opened 2023-01-05
Fixture of the issue with the new cut with undefined integration range

abuts:
New approach to cuts requests image range to be valid in any case, even if the cut does not contain these dimensions. I've fixed the issue for the particular case provided, though the proper approach to the question is a bit unclear.


horace_core/sqw/@axes_block/build_from_input_binning.m
```diff
@@ -87,6 +87,20 @@
             range = [range(1),range(3)];
         end
     end
+    if numel(range) == 2
+        if range(1)==range(2)  % some old file formats do not store proper img_range
+            % and store 0,0 for integration ranges. Here we try to mitigate
+            % this
+            av_pt = 0.5*(range(1)+range(2));
```
cmarooney-stfc, line 94: Assuming your comment is correct and the case you are trying to catch is range==[0,0], av_pt==0. Even if not, av_pt==range(1). May be worth tidying this up on this basis.

horace_core/sqw/@axes_block/build_from_input_binning.m
```diff
@@ -87,6 +87,20 @@
...
+            if abs(av_pt)<abs(SQWDnDBase.border_size)
+                range(1) = -abs(SQWDnDBase.border_size);
+                range(2) =  abs(SQWDnDBase.border_size);
+            else
+                range(1) = av_pt*(1-abs(SQWDnDBase.border_size));
+                range(2) = av_pt*(1+abs(SQWDnDBase.border_size));
+            end
```
cmarooney-stfc, line 101: Now you have 2 cases: av_pt on the border - let it straddle a bit, and, av_pt off the border, push apart by a border width. May be worth explaining why only the 0 border is treated - are there any others?

cmarooney-stfc (changes requested):

horace_core/sqw/@axes_block/build_from_input_binning.m
abuts: 0-border is treated differently because multiplying by 0 gives 0, not a range. If av_pt is large, adding/subtracting ~`eps` to it do not make any difference. In this case, its better to treat border as relative

abuts: as for the first notice -- that's probably correct. `av_pt == range(1)` if `range(1)==range(2)`

joshuadkitenge (changes requested):
I've tested this with my example and there are now values in the mnf2_cut after cut has been ran.

However the example still doesn't run as it did on Hoarce 3.6.2 . The chisq values on this branch is many orders of magintude bigger than the results yielded using Horace 3.6.2 for the same exact problem.

I'm guessing this may be a issue with multifit ?  

I've attached the command line output at the bottom. 

**Example output : Horace 3.6.2** (only the top and bottom of the command line output)

_**Top of the output**_

Screenshot 2023-01-05 144233

**_bottom of the output_**

image

**Example output : Horace 4.0.0** (The whole output)

Screenshot 2023-01-05 143201

PR #927, opened 2023-01-11
Add ADD updating single-filebacked

oerc0122:
ADD detailing decisions from meeting 11/1/23

### DIFF
--- a/documentation/add/10_single_filebacked_operations.md
+++ b/documentation/add/10_single_filebacked_operations.md
@@ -3,35 +3,36 @@ Date: 2023-01-11
 
 ## Objectives and problems to solve
 
-Currently the implementation of single-filebacked objects is based upon a temporary file beinhg created as infrequently
+Currently the implementation of single-filebacked objects is based upon a temporary file being created as infrequently
 as possible and modifications to the sqw data taking place within that file. The issue with this mechanism is that, while
 it may result in less data being written if operations only modify components of the data (e.g. signal/variance for many
 unary/binary ops), nevertheless if a job fails or is cancelled partway through, the user is left without a viable (or worse yet a
-viable but incorrect/invalid object).
+viable but incorrect/invalid) object.
 
 Thus, it was decided that the operation should only replace the temporary file once the full operation has been
 completed and otherwise leaves behind a correct file, and any intermediaries are cleared upon finalisation of an
 operation.
 
-Upon instantiating an `sqw` object from a file, several things take place and an `sqw` object is constructed with the
+Upon instantiating an `sqw` object from a file, several pieces of data are loaded and an `sqw` object is constructed with the
 following structure:
 
 - sqw
-  - main header
-  - pix == `PixelData` object
-  - experiment info
-  - headers
+  - `main_header`
+  - `pix` => `PixelData` object
+  - `data` => `dnd` object
+  - `experiment_info`
+  - `detpar`
 
 The `PixelData` object here is the object of interest.
 
 ## Prior Implementation
 
-Prior to the refactor, there existed one object, `PixelData` which handled both in-memory and file-backed operations,
-though was incomplete.
+Prior to the refactor, there existed one class, `PixelData` which handled both in-memory and file-backed operations,
+though this was incomplete in its handling of both.
 
-It also relied on creating multiple files of "pages" of partial pixel data. This would allow functions to be applied to
-part of an `sqw` object; however, Horace does not support application to parts of `sqw`s, except through explicitly
-creating a slice of an `sqw` object through `cut`, which creates a separate instance of `PixelData` anyway.
+It relied on creating multiple files of "pages" of partial pixel data. This would, in principle, allow functions to be
+applied to part of an `sqw` object; however, Horace does not support application to parts of `sqw`s, except through
+explicitly creating a slice of an `sqw` object through `cut`, which creates a new instance of `PixelData`.
 
 ## Current Implementation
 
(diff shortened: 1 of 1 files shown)


documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
+# Intermediaries of Single Filebacked Operations
+Date: 2023-01-11
+
+## Objectives and problems to solve
+
+Currently the implementation of single-filebacked objects is based upon a temporary file beinhg created as infrequently
+as possible and modifcations to the sqw data taking place within that file. The issue with this mechanism is that while
```
cmarooney-stfc, line 7:
```suggestion
as possible and modifications to the sqw data taking place within that file. The issue with this mechanism is that, while
```

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+it may result in less data being written if operations only modify components of the data (e.g. signal/variance for many
+unary/binary ops) if a job fails or is cancelled partway through the user is left without a viable (or worse yet a
```
cmarooney-stfc, line 9:
```suggestion
unary/binary ops), nevertheless if a job fails or is cancelled partway through, the user is left without a viable (or worse yet a
```

`Nevertheless` is padding designed to ensure that the following if clause is clearly separated from the previous if.

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+viable but incorrect/invalid object).
+
+Thus, it was decided that the operation should only replace the temporary file once the full operation has been
+completed and otherwise leaves behind a correct file and any intermediaries are cleared upon finalisation of an
```
cmarooney-stfc, line 13:
```suggestion
completed and otherwise leaves behind a correct file, and any intermediaries are cleared upon finalisation of an
```

Avoids ambiguity that `and any` goes with `correct file`.

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+operation.
+
+Upon instantiating an `sqw` object from a file, several things take place and an `sqw` object is constructed with the
+following structure:
+
+- sqw
+  - main header
+  - pix == `PixelData` object
+  - experiment info
+  - headers
```
cmarooney-stfc, line 23: Just checking this - before we renamed the experiment info data, it was the headers. What's in the headers now?

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+
+The `PixelData` object here is the object of interest.
+
+## Prior Implementation
+
+Prior to the refactor, there existed one object, `PixelData` which handled both in-memory and file-backed operations,
+though was incomplete.
+
+It also relied on creating multiple files of "pages" of partial pixel data. This would allow functions to be applied to
+part of an `sqw` object, however, Horace does not support application to parts of `sqw`s, except through explicitly
```
cmarooney-stfc, line 33:
```suggestion
part of an `sqw` object; however, Horace does not support application to parts of `sqw`s, except through explicitly
```

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+creating a slice of an `sqw` object through `cut`, which creates a separate instance of `PixelData` anyway.
+
+## Current Implementation
+
+Following the refactor the `PixelData` object was changed to the following hierarchy:
```
cmarooney-stfc, line 38:
```suggestion
Following the refactor the `PixelData` class was changed to the following hierarchy:
```

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+
+- `PixelDataBase` - Abstract base class
+  - `PixelDataMemory`
+  - `PixelDataFileBacked`
+- `PixelData` -- Exists only as a wrapper to allow tests to load old `PixelData` objects as new `PixelDataBase` format,
```
cmarooney-stfc, line 43:
```suggestion
- `PixelData` -- Exists only as a wrapper to allow tests to load old `PixelData` objects as the new `PixelDataBase` format
```

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+  cannot be manually instantiated.
+
+`PixelDataMemory` is a simple problem to solve as all operations happen in-place at once in memory and are subsequently
```
cmarooney-stfc, line 46: Not clear that it is a problem! Maybe rephrase to make it clear that it is not a problem.

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+saved to disc only upon a saveobj operation being called on the parent `sqw` object.
+
+`PixelDataFileBacked`, however, is a trickier proposition as intermediary objects are created out of necessity (if the
```
cmarooney-stfc, line 49:
```suggestion
`PixelDataFileBacked`, however, is a trickier proposition, as intermediary objects are created out of necessity (if the
```

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+object is to be file-backed it is implied to be too large to fit in memory all at once or at the very least it would be
+inconvenient to do so); as such, the resulting product must be constructed piecemeal.
+
+Upon the first operation being applied to the `PixelDataFileBacked` object, data is loaded from the sqw-file through
+`faccess` routines in chunks (still called "pages"), the operation is applied to each chunk and then appended to a
```
cmarooney-stfc, line 54:
```suggestion
`faccess` routines in chunks (still called "pages"), the operation is applied to each chunk, which is then appended to a
```

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+
+Currently, the temporary file does not contain any headers or other information required for constructing an `sqw`
+object and instead just contains raw pixel numbers.
+
+## Proposed Implementation
+
+While the object hierarchy will remain unchanged from the current implementation, concerns about half-completed
+operations irreparably (or invisibly) corrupting pixels will require attention. To this end, it is proposed that all
+operations on `PixelDataFileBacked` will result in a new temporary file being created and only upon completion of the
+operation will the file be moved to overwrite the old temporary.
```
cmarooney-stfc, line 70: Not sure this is completely clear.

I *think* that for any operation (say the third), the previous second operation will have created a new temporary file; at the succesful end of the third operation the new temporary file will be moved to overwrite the old temporary file. Thus there are 2 temporary files during the operation, and the new one overwrites the old one when the operation concludes. The operations are all the same operation e.g. add and differ only in applying to different pages of the file.

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+
+Access to properties can still be done through a `memmapfile` for loading data segments, however, since assignment will
```
cmarooney-stfc, line 72:
```suggestion
Access to properties can still be done through a `memmapfile` for loading data segments; however, since assignment will
```
`However` is annoying in being usable at the end of a clause as well as the beginning hence the need for firmer punctuation to disambiguate

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+no longer be done through `memmapfile`, it may be advantageous to consider strided loading.
```
cmarooney-stfc, line 73: This is a statement of an option, not a firm proposal. Will you do this, or not?

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+
+The new implementation will also include writing the header-data into the temporary file, to allow it to be directly
+loaded through `faccess` as an `sqw`. `saveobj` on a `PixelDataFileBacked` will simply change the name of the temporary
+sqw to a more permanent name.
```
cmarooney-stfc, line 77: That is, it is only a file rename, not a copy, yes?

cmarooney-stfc (changes requested):

documentation/add/10_single_filebacked_operations.md
oerc0122: Yes, eg. 
```MATLAB
save(my_favourite_pixeldata, 'data.sqw')
```
would rename from `blah_1295012059.tmp_sqw` to `data.sqw`

documentation/add/10_single_filebacked_operations.md
oerc0122: I'm not sure, I was going to double check what the actual structure was and forgot.

documentation/add/10_single_filebacked_operations.md
oerc0122: I am undecided and exploring what is needed in terms of functionality in the FileBacked stuff.

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
+# Intermediaries of Single Filebacked Operations
+Date: 2023-01-11
+
+## Objectives and problems to solve
+
+Currently the implementation of single-filebacked objects is based upon a temporary file beinhg created as infrequently
+as possible and modifications to the sqw data taking place within that file. The issue with this mechanism is that, while
```
abuts, line 7: This is not what is currently happens. 
Two currently  working ways of creating new sqw files is 
1) Calculate pix in memory and store them in a file
2) `gen_sqw` -- combining pixels according to bins taken from different smaller sqw files
Everything else is subset of these two operations
Third way -- combining final file from modified pages implemented by Tessella is broken.

oerc0122: "Current" is the `single-fb` branch

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+
+- `PixelDataBase` - Abstract base class
+  - `PixelDataMemory`
+  - `PixelDataFileBacked`
+- `PixelData` -- Exists only as a wrapper to allow tests to load old `PixelData` objects as the new `PixelDataBase` format
+  cannot be manually instantiated.
+
+`PixelDataMemory` is a simple problem to solve as all operations happen in-place at once in memory and are subsequently
+saved to disc only upon a saveobj operation being called on the parent `sqw` object.
+
```
abuts, line 48: `saveobj` operation  is absolutely independent way of saving pixels into `.mat` files. These pixels are stored on disk and loaded back as the whole block using Matlab mechanism

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+`PixelDataFileBacked`, however, is a trickier proposition, as intermediary objects are created out of necessity (if the
+object is to be file-backed it is implied to be too large to fit in memory all at once or at the very least it would be
+inconvenient to do so); as such, the resulting product must be constructed piecemeal.
+
+Upon the first operation being applied to the `PixelDataFileBacked` object, data is loaded from the sqw-file through
```
abuts, line 53: Chunks should die. They do not fit usage patterns. The pixels block should be loaded from source file using either `memmapfile` or normal read operations -- whatever is simple and stored into a **new binary file** by binary write operations (as `memmapfile` does not support this. The new binary file should become final `sqw` file by rename operation, probably invoked by `save` operation applied to filebacked resulted sqw object

documentation/add/10_single_filebacked_operations.md
```diff
@@ -0,0 +1,77 @@
...
+`faccess` routines in chunks (still called "pages"), the operation is applied to each chunk, which is then appended to a
+temporary file (e.g. `sqw000015251.tmp_sqw`).
```
abuts, line 55: the pages loaded in memory, processed and put back into the target file. The target file is temporary until algorithm which modifies **all** source blocks completes.  After the algorithm is completed, we are obtaining new sqw object. 
The object may be filebacked, i.e. contains only main part of its data in memory and refer to the just produced modified pixel data in file. 

We may access these data trough subsequent binary read operations (may be preferred for unit operations) or through `memmapfile` operations (preferable for cuts). What is better I do not know. 
May be an filebacked operation should complete by closing final file and opening it for `memmapfile` access if filebacked object remains in memory

abuts: both approaches have their benifits and drawbacks

PR #938, opened 2023-02-03
Persistent cluster

oerc0122:
Potentially breaking changes, experiment to see how tests are affected.

Design might need further work


abuts (commented):
all looks fine except JobDispatcher should be a handle and idle job should be written launched between other jobs. 
Iddle job would keep cluster running and regularly asking if other job is submitted. 
Cluster dies after iddle job runs specified time interval

As first step here would be making JD a handle and everything else could be done as the next step

abuts:
I am closing this as job is half-done but its easier to start it from the beginning than to understand what is left and what is working

PR #941, opened 2023-02-08
fix parallel call

oerc0122:
Series of fixes to `parallel_call` to handle passing of `nWorkers`, handle Tobyfit functions in `simulate`, add `dummy` MPI test to `test_parallel_call` to check core features working, move `split_data` to static class method on `mfclass` for external use and rename. 

TODO: Add background function support to `parallel_call` for MF jobs


herbert_core/classes/MPIFramework/parallel_call/parallel_call.m
```diff
@@ -1,23 +1,21 @@
...
-        varargout{1} = parallel_sqw_eval(w, args);
+        varargout{1} = parallel_sqw_eval(func, nWorkers, args);
 
       case 'multifit_func_eval'
 
-        w = args{1};
-        args = args(2:end);
+        % mf_func_eval signature:
+        %multifit_func_eval (wmask, xye, fun_wrap, bfun_wrap, pin_wrap, bpin_wrap,...
+        %    f_pass_caller, bf_pass_caller, pfin, p_info, output_type)
```
cmarooney-stfc, line 16: do we want to keep this?

cmarooney-stfc (changes requested):

herbert_core/classes/MPIFramework/parallel_call/parallel_call.m
oerc0122: I personally think it's useful for determining why it's structured the way it is. I would probably prefer putting the other function signature at top of the other case.

Issue #942, opened 2023-02-09
Solve conflicts between config files pre- and post- merge of Horace and Herbert


tgperring:
Running multiple instances of Horace on one device has often caused problems due to sharing the config files and temporary folder. At the very least it usually results in copious and opaque warning messages being produced when typing horace_on , which is disconcerting to those who are not of a sanguine temperament.  The merge of Herbert and Horace and their corresponding configurations, running the release versions (3.6.2) and master (4.0.0) appears to cause particularly severe clashes in my experience.

Being able to independently run the latest release (3.6.2) and master (4.0.0) in separate matlab instances on the same machine is desirable for two reasons
(1) Release 3.6.2 is the benchmark against which the huge changes that have since taken place in master and regular cross-referencing of output in the development phase is something developers will want to do all the time (I certainly do)
(2) When the MVP release is finally made, it is desirable that users can fall back on 3.6.2 if problems are highlighted in using 4.x.x

To achieve this, the configuration files and temporary file folders will need to be version number distinct.

PR #946, opened 2023-02-12
Change config_store and tmp_dir() so that different versions of Horac…

tgperring:
[continued from title] …e use different locations for the config files and also for temporary files, by appending the version to the mprogs_config folder, and creating a subfolder in the temporary area found/created by tmp_dir.m() location. See ticjket #942 . This enables different versions of Horace to be run without their configurations interfering, and avoid conflicts with temporary file creation (e.g. during gen_sqw). For example, the current version 4 development version uses C:\Users\tgp98\Documents\matlab\mprogs_config_4.0.0.dev on Toby's laptop (c.f. C:\Users\tgp98\Documents\matlab\mprogs_config), and C:\Users\tgp98\AppData\Local\Temp\Horace_4.0.0.dev (cf C:\Users\tgp98\AppData\Local\Temp). Most particularly, it means that it is possible to run the current release and development (Master) versions on the same machine.


cmarooney-stfc (commented):
The overall aim of comparing the current branch (master or other) with a release seems to work. I've added one minor comment on existing logic. However this still does not permit comparison between 2 branches (master, my branch) which is very useful to find out why what was working now doesn't, by comparing behaviours between them.

tgperring:
> 

Ere

tgperring:
Re Jacob's comment: different git versions still use the same config. At the moment we always have the version as 4.0.0.dev. This is at the root of Chris' comment; we only have the one location in addition to the one that is used by the release version. That is, we only have two configs, one for the present release and one for the development. To strike a balance we probably want to to have a config file version number, and only have a new location when the config version is updated. However, I am reluctant at this point to edit further, as the primary goal of having the release version and development version has been solved, and which will also mean that when we release the MVP we can always fall back on the release while we troubleshoot (the inevitable) problems with the MVP when it is exposed to users.

PR #948, opened 2023-02-13
port new changes from rel 3.6.3 to horace 4.0

abuts:
These changes should reflect changes, done on IDAAAS


herbert_core/admin/is_idaaas.m
```diff
@@ -36,18 +36,14 @@
         size_suffix = '';
         return;
     end
-    
+
     cpu_pos = strfind(mess,'CPU(s)');
     mess = strsplit(mess(cpu_pos(1):end));
     n_cpu = str2double(mess{2});
     if n_cpu <10
```
oerc0122, line 43: Why is large if `n_cpu < 10`? not `n_cpu > 10`?

herbert_core/configuration/@config_store/config_store.m
```diff
@@ -446,27 +451,38 @@ function set_config_path(obj,new_path,varargin)
...
+                if ver >= 400
+                    ver_string = '_v4';
+                else
+                    ver_string = '';
+                end
                 [is_virtual, type] = is_idaaas();
+
                 if is_virtual
-                    obj.config_folder_name_ = ['mprogs_config_', ver_string, '_', type];
+                    obj.config_folder_name_ = ['mprogs_config','_', type,ver_string];
```
oerc0122, line 483:
```suggestion
                    obj.config_folder_name_ = ['mprogs_config_', type, ver_string];
```

herbert_core/configuration/@config_store/config_store.m
```diff
@@ -446,27 +451,38 @@ function set_config_path(obj,new_path,varargin)
...
                 else
-                    obj.config_folder_name_ = ['mprogs_config_', ver_string];
+                    obj.config_folder_name_ = ['mprogs_config', ver_string];
```
oerc0122, line 485:
```suggestion
                    obj.config_folder_name_ = ['mprogs_config_', ver_string];
```

herbert_core/admin/@opt_config_manager/private/find_comp_type_.m
```diff
@@ -50,14 +44,16 @@
...
     % if lscpu returns more then one numa node strigs, first string defines
     % the number of numa nodes and all subsequent strings describe each
     % node. So, if there are more then 2 string, its more then one numa
     % node and we consider this computer to be an hpc system.
-    if numel(rez)>2; hpc_computer = true;
-    else;          hpc_computer = false;
+    if numel(rez)>2 || nproc>4
+        hpc_computer = true;
+    else
+        hpc_computer = false;
```
oerc0122, line 56:
```suggestion
    hpc_computer = numel(rez) > 2 || nproc > 4;
```

herbert_core/configuration/@config_store/config_store.m
```diff
@@ -446,27 +451,38 @@ function set_config_path(obj,new_path,varargin)
...
             else
-                ver_string = herbert_version();
+                % analyze version number and if it exceeds the number
+                % requested, specify other version folder.
+                % To be extended in a future when verison number changes
+                % and config folder either need changes or not depending on
+                % various Horace version(s)
+                ver = herbert_version('-num');
+                if ver >= 400
+                    ver_string = '_v4';
```
oerc0122, line 476:
```suggestion
                    ver_string = num2str(floor(ver/100))
                    ver_string = ['_v', ver_string];
```
Or perhaps better simply:
```suggestion
                    ver_string = num2str(floor(ver))
                    ver_string = ['_v', ver_string];
```

_test/test_transformation/test_spher_proj.m
```diff
@@ -0,0 +1,32 @@
+classdef test_spher_proj<TestCase
```
oerc0122, line 1: Is this meant to be included? Unrelated.

herbert_core/utilities/misc/horace_memory.m
```diff
@@ -0,0 +1,77 @@
...
+        [nok,mem_string] = system('free | grep Mem');
+        if ~nok
+            [phys_memory,free_memory] = parse_mem_string(mem_string);
+        end
+    end
+    if nok
+        phys_memory = [];
+        free_memory = [];
+    end
+end
```
oerc0122, line 31:
```suggestion
if ispc
    [~,sys] = memory();
    ok = true;
    phys_memory = sys.PhysicalMemory.Total;
    free_memory = sys.PhysicalMemory.Available;
elseif ismac
        [err,mem_string] = system('top -l 1 -n 0');
        ok = err == 0;
        if ok
            [phys_memory,free_memory] = parse_mac_mem_string(mem_string);
        end
elseif isunix
    [err,mem_string] = system('free | grep Mem');
    ok = err == 0;
    if ok
        [phys_memory,free_memory] = parse_mem_string(mem_string);
    end
end

if ~ok
    phys_memory = [];
    free_memory = [];
end
```

herbert_core/utilities/misc/horace_memory.m
```diff
@@ -0,0 +1,77 @@
...
+% Analyze result of free function on linux
+try
+    cont = regexp(mem_string,'\s+','split');
+    phys_memory =  1024*sscanf(cont{2},'%d');
+    free_memory =  1024*sscanf(cont{4},'%d');
+catch
+    phys_memory = [];
+    free_memory = [];
+
+end
```
oerc0122, line 43: Would prefer `try-catch` outside of functions, and probably throw if `nok` (or `err ~= 0`  w/ refactor), defaulting to empty arrays

oerc0122 (changes requested):

_test/test_transformation/test_spher_proj.m
abuts: yes, its erroneous test came from other ticket. Commented for now

oerc0122: `git rm --cached` is what you want, you could clean it up more nicely, but that will remove it from the commit.

herbert_core/configuration/@config_store/config_store.m
abuts: this is not what we want. I do not want configuration folder changes with every version. Only when it is necessary and can not be avoided it should change. 
Otherwise, when we go to version 4 and patch release 4.01 comes users will get it not configured again. Extremely annoying.

herbert_core/admin/is_idaaas.m
oerc0122: That is not what this statement says. This says `if n_cpu < 10; size_suffix = idaas_large`

herbert_core/configuration/@config_store/config_store.m
oerc0122: Hence why the former only updates when the major version changes, but we don't have to manually do so.

herbert_core/admin/@opt_config_manager/private/find_comp_type_.m
abuts: very "matter of taste" suggestion. I would not fining this more readable then the previous form and of course unlike C especially  in a loop -- no difference for performance

herbert_core/configuration/@config_store/config_store.m
abuts: BTW, this is question of large discussion with Toby, so I am leaving this for further consideration.

herbert_core/utilities/misc/horace_memory.m
abuts: another "matter of taste" suggestion. I would not fining this more readable then the previous form. May accept it for simplicity, but do not like spending time on such minuscules issues.

herbert_core/utilities/misc/horace_memory.m
abuts: ibid. Leave as it is.

herbert_core/configuration/@config_store/config_store.m
oerc0122: It seems foolish to introduce a change we're going to have to manually update once a year when it's trivial to automate, either that or we have to put it in the release SMG.

abuts: Its absolutely non-trivial to introduce reasonable automation because we do not know the purpose.  I believe that we do not generally want configuration changing with version and if it changes, autoconfiguration should work. This is specific case of the fundamental changes in the code when **users** may want to run two versions simultaneously. Normally such behavior is depreciated. 

SMG is useful in any case. Unfortunately there are so many such cases. 

I am asking Toby to contribute

tgperring:
I think we shouldn't get too distracted at the moment. The primary short-term goals are to be able to run the release version (3.6.x) and the master or a branch in different Matlab sessions on the same machine - this will enable(i) developers to perform simultaneous comparison between the benchmark (3.6.x) and development version (4.0.*), and (ii) when we release the MVP v4 that users can fall back on 3.6.* when the inevitable problems are found in 4.*, without the configs or tmp_dir interfering in any way.

Any solution that works for development and MVP is good enough for the moment. Ideally, we'd have a way of allowing different branches of 4.* to have separate config and tmp_dir - this is something Chris would find very convenient I know, to help find out out why a development branch he is working on is going wrong. But this should be implemented only if quick to implement and with pragmatism is the guiding principle. [E.g. could be have horace_on take an option, with an obscure name that a user will never think of, which gives the suffix that defines the config and tmp folders
>> horace_on (..., '-DEV_config', '928_Toby_dev_stuff', '-DEV_tmp', 'toby_junk')
which will use (create if necessary) on my laptop:
config_folder         'C:\Users\tgp98\Documents\matlab\mprogs_config_928_Toby_dev_stuff'
tmp_dir               'C:\Users\tgp98\AppData\Local\Temp\Horace_toby_junk\

With my wisdom of Solomon, I propose that we keep what Alex has done, which essentially separates the release from later versions, and *ONLY IF QUICK AND EASY TO IMPLEMENT* something like my suggestion above to give developers complete control. After MVP we can alter the code when we have a better plan for how we want configs to operate for users. It will conicide with Jacob's suggestions until we reach Horace version 5! We have so many things that are critical to work on, that the eye should be on making life easy for developers at the moment.

abuts:
Toby's suggestion is possible but not an hour to implement properly. It is 1-2 days work (with appropriate tests)
As this is housekeeping ticket intended to be quick -- I am closing it. @cmarooney-stfc  may use config folder setting if he wants weird configurations.  One more change to do -- config folder settings through property should not copy existing configuration -- this is probably what @cmarooney-stfc  wants.

PR #950, opened 2023-02-15
Add istext function

oerc0122:


cmarooney-stfc (approved):

abuts (approved):
Mantid does 50% of this kind of PR.  We should do it more often

PR #957, opened 2023-02-24
Complete FileBacked ops

oerc0122:
Filebacked operations support through creating temporary file with `PixelTmpFile` object which handles temporary file (naming, temp. naming, deletion on object deletion).

Rename series of tests for more brief, informative names. 

Supporting `mask`, `binary_op`, `unary_op`, `noisify`, `sqw_eval`, `shift_pixels`.

Do not support setting data in-place for most operations.


_test/test_sqw_pixels/test_mask.m
```diff
@@ -292,10 +270,10 @@ function test_mask_pixels_removes_pixels_given_in_mask_array(obj)
             new_sqw = mask_pixels(sqw_obj, mask_array);
 
             assertEqual(new_sqw.pix.num_pixels, sum(mask_array));
-            assertFalse(equal_to_tol(new_sqw.data.s, sqw_obj.data.s,[0, 1e-4]));
-            % masking have not changed binning, so img_range remains the
+%             assertEqualToTol(new_sqw.data.s, sqw_obj.data.s, [0, 1e-4]);
```
oerc0122, line 273: Should masking preserve `s`?

cmarooney-stfc (changes requested):

horace_core/sqw/PixelData/@PixelTmpFile/PixelTmpFile.m
```diff
@@ -0,0 +1,48 @@
...
+
+            for i = 1:5
+                obj.file_name = fullfile(tmp_dir(), ...
+                                             [name, '.pix_', str_random()]);
+                if ~is_file(obj.file_name) && ~is_file(obj.tmp_name)
+                    break
+                end
+            end
+            % Unlikely to happen, but best to check fail to generate
+            if i == 5 && (is_file(obj.file_name) || is_file(obj.tmp_name))
```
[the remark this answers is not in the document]
oerc0122: True, but that is the reason for the `(is_file(obj.file_name) || is_file(obj.tmp_name))` matlab's loops end ON the final point (which makes sense given they're iterating the elements of an array) not past them.

oerc0122:
Current issues:
- temporary files not cleared (potential issue with memmapfile not releasing locks)
- Assigning to new object (e.g. `c = binary_op(@plus, a, b)` will lead to subsequent operations on `c` affecting `a` due to `PixelTmpFile` carrying over to `c` so filenames will be same, leading to over-write of `a`'s memmapfile.

oerc0122:
Temporary files now clear

horace_core/sqw/PixelData/@PixelTmpFile/PixelTmpFile.m
```diff
@@ -1,5 +1,21 @@
 classdef PixelTmpFile < handle
-% Class to handle deletion of TmpFile on closure
+% Class temporary PixelDataFileBacked files
```
cmarooney-stfc, line 2:
```suggestion
% Class of or for temporary PixelDataFileBacked files
```
choose of or for

horace_core/sqw/PixelData/@PixelTmpFile/PixelTmpFile.m
```diff
@@ -1,5 +1,21 @@
...
+%
+% Created by PixelData file-backed on starting modifying operation
```
cmarooney-stfc, line 4:
```suggestion
% Created by PixelData file-backed on starting modifying operation.
```
Based on the assumption that the next line starts a new sentence. If PixelDataFileBacked.get_new_handle is a qualifier for operation then that is not the case, and there should be a which before stores in L5. Basic point is that this needs separating out somehow to be understandable

cmarooney-stfc (changes requested):

oerc0122:
- [ ] `sqw_eval_file_backed` needs implementing (not tested apparently)

oerc0122:
- [ ] `sqw_eval` still not fully file-backed through all paths

abuts (changes requested):
I do not like this and do not understand what is done here.  This is resurrection of  Tessella approach which does separate pixels and separate sqw, just without splitting pix into bunch of tmp files. 

I had a look at mask and unary operation, and see that masked pixel array is saved into separate pix file. 
I was discussing this possibility with instrument scientists and everybody was against separating sqw data and pixel data, so this is unacceptable. 

Next operation should move this file into sqw pixels and this operation is not here. In addition, filebacked operation would be long operation so it should not be there at all. 

I suspect other operations are performed in a similar fashion. Filebacked operations are not implemented. 

I may accept it as intermediate step before implementing full operations, which create total sqw file, as initial pattern was suggesting, but then I do not understand what is the meaning of all this work.

horace_core/sqw/@sqw/private/do_sqw_eval_filebacked_.m
```diff
@@ -22,57 +22,63 @@
...
+
     qw = calculate_qw_pixels(wout);
     sig_chunk = sqwfunc(qw{:}, pars{:});
 
-    ldr = write_pix_chunk_with_signal_(ldr, pix, sig_chunk);
-
-    img_signal = increment_signal_sums_( ...
-        img_signal, sig_chunk, npix_chunks{chunk_num}, idxs(:, chunk_num) ...
-    );
+    data(s_ind, :) = sig_chunk;
```
abuts, line 57: If written like this, averages are not recalculated

oerc0122: I'm not sure what you mean by this. What averages? How should they be calculated?

abuts (approved):
I think it may go as the first step and then the implementation of the discussed changes should follow

PR #958, opened 2023-02-27
Add parallel multifit tests

oerc0122:


cmarooney-stfc (approved):

abuts (approved):
Despite parallel multifit does not work well, tests are always useful for the future

PR #977, opened 2023-03-07
Fix horace_install with herbert_off gone

oerc0122:
Fix to removing herbert_off for horace_install and other misc. Should be hotfix.

Still not sure where 
> Function herbert_off.m not found in /home/jacob/PACE/HorBack/herbert_core
> Clearing rootpath and subdirectories from Matlab path in any case

Is coming from, but should be simple fix

### DIFF
--- a/herbert_core/herbert_init.m
+++ b/herbert_core/herbert_init.m
@@ -1,9 +1,8 @@
 function herbert_init
 % Adds the paths needed by Herbert.
 %
-% In your startup.m, add the Herbert root path and call herbert_init, e.g.
-%       addpath('c:\mprogs\herbert')
-%       herbert_init
+% In your startup.m, add the Horace root path and call horace_on, do not
+% call this directly
 %
 % Is PC and Unix compatible.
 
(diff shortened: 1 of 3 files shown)


herbert_core/herbert_init.m
```diff
@@ -76,11 +76,9 @@ function addgenpath_message (varargin)
     try
         addpath (genpath_special(string),'-frozen');
     catch ME
-        herbert_off
```
cmarooney-stfc, line 79: as there is now no functionality in the catch, you could remove the try/catch/rethrow

herbert_core/herbert_init.m
```diff
@@ -25,7 +25,7 @@
 % end
 % Add paths
 % ---------
-addpath(herbert_path);  % MUST have herbert_path so that herbert_init, herbert_off included
+addpath(herbert_path);  % MUST have herbert_path so that herbert_init, included
```
cmarooney-stfc, line 28: The source of 

Function herbert_off.m not found in /home/jacob/PACE/HorBack/herbert_core
Clearing rootpath and subdirectories from Matlab path in any case

is at L.106 in this file below.

cmarooney-stfc (changes requested):
The herbert_off not found message can now be dealt with

PR #979, opened 2023-03-08
rename a projection

abuts:
Trivial, though meaningful change -- testing work in new enviroment


oerc0122:
My main question is, why `aProjectionBase` not `ProjectionBase`?

horace_core/sqw/@sqw/private/cut_sqw_sym_main_single.m
```diff
@@ -81,7 +81,7 @@
         header_ave=header_average(header);
         
         
-        % Because Alex's aProjection class (of which proj is an instance)
+        % Because Alex's aProjectionBase class (of which proj is an instance)
         % is a rather odd hybrid object with projection and cut information
```
oerc0122, line 85: Is this comment still true?

horace_core/sqw/coord_transform/@aProjectionBase/aProjectionBase.m
```diff
@@ -151,22 +151,22 @@
...
+            % obj  -- Initialized instance of aProjectionBase class
             % remains
             %      -- if input arguments contains key-value pairs which do
-            %         not describe aProjection class, the output contains
+            %         not describe aProjectionBase class, the output contains
             %         cellarray of such parameters. Empty, if all inputs
             %         define the projection parameters.
             %
-            opt_par = aProjection.init_params;
+            opt_par = aProjectionBase.init_params;
```
oerc0122, line 169: If this is a function, would prefer parens
```suggestion
            opt_par = aProjectionBase.init_params();
```

horace_core/sqw/coord_transform/@aProjectionBase/calc_pbins.m
```diff
@@ -76,20 +76,20 @@
...
 n=length(pbin);
 if n<3 || n>4
-    error('aPROJECTION:invalid_arguments',...
+    error('HORACE:aProjectionBase:invalid_arguments',...
         'Must provide binning descriptor for all three momentum axes and (optionally) the energy axis');
 end
 
 if ~(isempty(pbin{1})||(isa_size(pbin{1},'row','double') && length(pbin{1})>=1 && length(pbin{1})<=4)) || ...
         ~(isempty(pbin{2})||(isa_size(pbin{2},'row','double') && length(pbin{2})>=1 && length(pbin{2})<=4)) || ...
         ~(isempty(pbin{3})||(isa_size(pbin{3},'row','double') && length(pbin{3})>=1 && length(pbin{3})<=4))
```
oerc0122, line 85: Would prefer this check to be a sensibly named function, then checked here through `all(cellfun(...))` particularly because it's also used repeatedly below. But not for this PR

oerc0122 (approved):
Beyond the question of why `aProjectionBase`, PR is fine.

horace_core/sqw/@sqw/private/cut_sqw_sym_main_single.m
abuts: no, this is about very old projection

horace_core/sqw/coord_transform/@aProjectionBase/calc_pbins.m
abuts: This is very old code which probably should go. I wanted to deleted it but this is used by cut_sym only so the time for changes have not come.

cmarooney-stfc (approved):

PR #980, opened 2023-03-09
971 refactor get new handle

oerc0122:
Opening as draft to get feedback on approach. 

Missing:
- [x] SQW `shift_pixels`
- [ ] `put_dnd`
- [x] others not checked


_test/test_sqw_class/test_eval.m
```diff
@@ -36,20 +34,18 @@ function test_func_eval_sqw(obj)
...
-            ds = sqw_eval(obj.sqw_obj, ...
-                          @test_eval.sqw_eval_tester, [], '-average');
+            ds = sqw_eval(obj.sqw_obj, @test_eval.sqw_eval_tester, [], '-average');
 
             sig = ds.data.s;
             pix = ds.pix;
 
             assertEqual(sig(1), pix.signal(1));
-            assertEqual(pix.signal(1), 210);
+            assertEqual(pix.signal(1), 36);
```
oerc0122, line 43: Side effect of using HKLE as function elements will fix

_test/test_sqw/test_sqw_dnd_eval.m
```diff
@@ -202,8 +186,7 @@ function test_sqw_eval_dnd(obj)
             elseif sz(2) ~=1
                 error('FUNC_EVAL:runtime_error','incorrect shape of input arrays');
             end
-            dis = ones(sz);
-            dis(1) = numel(x);
+            dis = 2.*ones(sz);
         end
```
abuts, line 190: such a bed practice for chaning test data. The previous implementation was verifying if it picks up all data, not a partial one. 
if you leave it as it was before, I would see it works as before. Now I do not know without running the code

oerc0122: The function relying on the size of the data doesn't work for file-backed vs. memory-backed because the data are not all loaded at once

horace_core/sqw/PixelData/@PixelDataFileBacked/PixelDataFileBacked.m
```diff
@@ -386,6 +417,49 @@ function format_dump_data(obj, pix)
...
+            start_idx = 1;
+            for i = 1:numel(varargin)
+                curr_pix = varargin{i};
+                for page = 1:curr_pix.num_pages
+                    [curr_pix,data] = curr_pix.load_page(page);
+                    obj.format_dump_data(data, start_idx);
+                    start_idx = start_idx + size(data,2);
+                end
+            end
+
```
abuts, line 459: without playing with the code, I do not see where pix_data_averages are recalculated. If they are, then fine. Would do. If not -- I am surprized unit test have not found it

oerc0122: I still don't know what you mean by `pix_data_averages` do you have an example?

horace_core/sqw/PixelData/@PixelDataFileBacked/do_unary_op.m
```diff
@@ -9,7 +9,9 @@
 
 pix_out = obj;
 
-pix_out = pix_out.get_new_handle();
+if isempty(pix_out.file_handle_)
+    pix_out = pix_out.get_new_handle();
+end
```
abuts, line 14: I suspect this is a bug or unacceptably inefficient code. 
page averages have to be calculated per page and then combined together. 
In addition, signal and error of the DnD image have to be calculated in the same loop as pixels (they are also "kind of" pixel averages. 

If everything here formally works (I may assume that you loop over pixels here) then loop over pixels calculateing pixel averages, and then loop over pixels on SQW level recalculating DnD.s, DnD.e, and DnD.npix (probably untached here, unless you are masking) this unacceptable.

Looping over avearage file on iDaaaS now takes about 5-10 minutes at best. Doing it three times would not satisfy anybody.

horace_core/sqw/@sqw/noisify.m
```diff
@@ -26,14 +26,14 @@
...
 for i=1:numel(w)
     if has_pixels(w(i))   % determine if sqw or dnd type
-        
+
         % Delegate to PixelData to noisify that object on a page-by-page
         % basis using the Herbert noisify.
+        wout(i) = wout(i).get_new_handle();
         wout(i).pix = w(i).pix.noisify(varargin{:});
-        wout(i)=recompute_bin_data(wout(i));
+        wout(i) = recompute_bin_data(wout(i));
```
abuts, line 34: again, double loop over pixels and then over data. Very inefficient

horace_core/sqw/PixelData/@PixelDataFileBacked/do_unary_op.m
oerc0122: As noted, performance is not a concern at this stage. Functionality is, we can optimise it later. As it is, this is an infinite speedup over whatever exists for cases where file backing is necessary.

abuts (commented):
If pix_averages calculated properly in all filebacked operations, looks fine as work in progress. 

Need optimization to consider it as final result

cmarooney-stfc (approved):
nothing obvious to comment on

Issue #986, opened 2023-03-17
Herbert MPI fails intermittently


oerc0122:
Herbert MPI sometimes fails with 
```
The name Java process sucsessfulluy completed is not a registered message name
```

cmarooney-stfc:
PR-987 was merged

PR #987, opened 2023-03-17
Fix herbert MPI issue #986

oerc0122:


cmarooney-stfc (approved):
err_mess prevents mess from interfering with subsequent code.
successfully typo fixed as caught by Java if not

PR #992, opened 2023-03-20
Stop `bmatrix` returning mess and use try-catch

oerc0122:


horace_core/lattice_functions/rlu_corr_to_lattice.m
```diff
@@ -17,14 +17,17 @@
...
-%                  in the two frames are related by 
+%                  in the two frames are related by
 %                       v(i)= rotmat(i,j)*v0(j)
 %   ok              =true if all ok; =false otherwise
 %   mess            Error message if ~ok; ='' if ok
 
-[b0,arlu,angrlu,mess] = bmatrix(alatt0,angdeg0);
-if ~isempty(mess)
-    ok=false; return
+try
```
cmarooney-stfc, line 25: this eliminates ok,mess in the call but restores it for what this function returns. Should we maintain the try,catch here or just let the error propagate?

cmarooney-stfc (changes requested):

horace_core/lattice_functions/rlu_corr_to_lattice.m
oerc0122: As other functions may rely on `rlu_corr_to_lattice`'s `ok/mess`, it is safer to leave it for now, a further cleanup can be done later which includes `rlu_corr_to_lattice`

Issue #846, opened 2022-09-02
The projection, recovered from ub matrix does not correspond to initial projection, producing initial ub matrix in non-orthogonal case


abuts:
The tests, demonstrating this are:
`_test\test_transformation\test_instr_proj_construction:test_get_projection_from_other_aligned_data` 
+ tests for alignment:
`_test\test_change_crystal\test_change_crystal_file_eq_memory`

This issue should be fixed in one or another form.

abuts:
addressed by #892

Issue #827, opened 2022-07-19
reconsile Toby's projection and get_proj_from old data


abuts:
To support old Horace versions, new projection is recovered from old projection data (UB matrix)
The code, used to do that (`ortho_proj->private->build_from_old_struct`) is different from the one, Toby uses for symmetrized projections (`sqw->get_proj_and_pbin` -- now removed within the ticket #810)  Both code versions should be reconciled to assure they provide correct results in any case

abuts:
addressed by #892

PR #996, opened 2023-03-24
nonorthogonal projection recovery

abuts:
This change adds piece of code, which recovers non-orthogonal projection, stored in old file format with u_to_rlu matrix only. 
This closes tickets #846 and #827


horace_core/sqw/coord_transform/@ortho_proj/ortho_proj.m
cmarooney-stfc, line 1: This appears to have been extended long since from orthogonal to orthogonal and non-orthogonal. It may be worth renaming to something general rather than ortho

horace_core/sqw/coord_transform/@ortho_proj/ortho_proj.m
```diff
@@ -334,7 +347,49 @@
             % property
             [~, mat] = obj.uv_to_rot();
             mat = [mat,[0;0;0];[0,0,0,1]];
-        end        
+        end
+
+        function  [rlu_to_ustep, u_rot, ulen] = uv_to_rot(proj,ustep)
+            % Determine the matrices partially used for conversion
```
cmarooney-stfc, line 348: what does partially mean here? Please describe the partialness and point to it in the code

Also, why have you moved this function from after the next one? It means I can't see change in the code.

horace_core/sqw/coord_transform/@ortho_proj/ortho_proj.m
abuts: probablty it should be named `triclinc_proj`  or `tricl_proj` and after discussion (connotations?) it should be renamed accordingly. 
Should be separate ticket though, as a lot of renaming and changes to tests over saved values.

horace_core/sqw/coord_transform/@ortho_proj/ortho_proj.m
abuts: I do not remember what "partial" meant in this case (may be 3x3 or only some of them used) so removed this word.

PR #1012, opened 2023-04-20
Add generate SQW cube

oerc0122:
Add ability to generate a cube of SQW data for simple testing of symmetrise/cut. For even counts these are
`[-X/2 + 0.5 : 1 : X/2 - 0.5]` i.e. for 2 -> `[-0.5 0.5]`
for odd counts these are
`[-floor(X/2):floor(X/2)]` i.e. for 3 -> `[-1 0 1]`
These are then `cut` such that bins contain 1 pixel per bin of evenly spaced bins in a grid


horace_core/sqw/@sqw/generate_cube_sqw.m
```diff
@@ -0,0 +1,65 @@
+function out = generate_cube_sqw(shape)
+    % Generate a cube of SQW data FOR TESTING CUT/SYMMETRISE
+    % The data will be invalid in most circumstances
```
cmarooney-stfc, line 3: what does invalid mean in this context?

horace_core/sqw/@sqw/generate_cube_sqw.m
```diff
@@ -0,0 +1,65 @@
...
+
+    if isscalar(shape) && isnumeric(shape)
+        ndim = 4;
+        npix = shape^ndim;
+        coords = zeros(9, npix);
+
+        if mod(shape, 2) == 0
+            col = [-shape/2 + 0.5:shape/2 - 0.5];
+        else
+            col = [-floor(shape/2):floor(shape/2)];
```
cmarooney-stfc, line 14: this is a bit opaque. If shape is restricted to integers, as in the PR main comment, then floor(shape/2) is equivalent to (shape+1)/2 which would be easier to think about for maintenance purposes. It would be useful to comment this at the top of the function and test for it.

if in fact shape is any real value, then again comment at top of function. But is that what you want?

horace_core/sqw/@sqw/generate_cube_sqw.m
```diff
@@ -0,0 +1,65 @@
...
+        end
+
+        for i = 1:ndim
+            coords(i, :) = repmat(repelem(col, shape^(i-1)), 1, shape^(ndim-i));
```
cmarooney-stfc, line 18: Be really good to have a description in words of what this does, may be with an example.

cmarooney-stfc (changes requested):

horace_core/sqw/@sqw/generate_cube_sqw.m
oerc0122: It means that you can't actually use the data for much beyond cuy/symmetrise and will generally result in crashes or nonsense if used outside of those.

_test/test_sqw/test_dummy_sqw.m
```diff
@@ -97,5 +100,33 @@ function test_build_dummy_sqw(obj)
...
             assertElementsAlmostEqual(pix,pix1,'absolute',1.e-7);
         end
+
+        function test_gen_cube_2x2x2x2(obj)
+            tsqw = sqw.generate_cube_sqw(2);
+
+            assertTrue(isa(tsqw,'sqw'));
+            assertEqual(tsqw.pix.num_pixels, 2^4);
+
+            tval = tsqw.pix.coordinates';
```
cmarooney-stfc, line 110: add comment saying test is to ensure no duplicates created

PR #1014, opened 2023-04-25
Fix issue with close binning ranges converted to integration

oerc0122:
Make it so that non-overlapping close binning ranges are not converted into integration ranges


cmarooney-stfc (changes requested):
as advised verbally

abuts (approved):
do not see any problem with these changes. May be the issue with default range being integration range and binning step is provided should be fixed as we are here anyway.

PR #1029, opened 2023-06-05
ortho proj comparison

abuts:
This is helper changes, defining ortho_proj comparison operator. 
The projections are considered equal if their transformation are equal.

### DIFF
--- a/horace_core/sqw/coord_transform/@ortho_proj/ortho_proj.m
+++ b/horace_core/sqw/coord_transform/@ortho_proj/ortho_proj.m
@@ -536,29 +536,32 @@ classdef ortho_proj<aProjectionBase
         end
         function [is,mess] = eq(obj,other_obj,varargin)
             % Overloaded equality operator comparing the projection
-            % transformation rather then 
-            % 
+            % transformation rather then all projection properties
             %
             % Inputs:
             % other_obj -- the object or array of objects to compare with
             % current object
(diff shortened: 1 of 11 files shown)


horace_core/sqw/coord_transform/@ortho_proj/ortho_proj.m
```diff
@@ -534,6 +534,38 @@
             flds = saveableFields@aProjectionBase(obj);
             flds = [flds(:);obj.fields_to_save_(:)];
         end
+        function [is,mess] = eq(obj,other_obj,varargin)
```
oerc0122, line 537: Do not overload the `eq` (`==`) operator like this, the interface should be the same as the native/in-built function (`ok = eq(a,b)`), we already have `equal_to_tol` which defines this style of interface and should be used instead.

abuts: well, we need this kind of overload in eq_to_toll. It is actively used there. The paren (in serializable) also have such interface.

And native overload also works, so I do not see reason not to use this kind of overloading unless you want two different eq functions with minor changes.

oerc0122: Yes, I would much rather have two different equality functions rather than abusing MATLABs permissive overloading.

PR #1038, opened 2023-06-14
Parallel cut

oerc0122:


cmarooney-stfc (approved):

abuts (commented):
I do not know. It is very different from anything I would imagine. 

A parallel cut would well satisfy split/combine pattern where you split your sqw object onto smaller chunks, calculate contributing pixels and their averages on each node, and send results (at least from the average in one of the operation modes) to a headnode to combine partial images into the whole image. 

The infrastructure, which is created for that in `parallel_worker` and `JobExecutor` is completely unused so the questions occur, where the progress logging of each parallel job goes, how partial combine happen and where the synchronization for combining chunks into the whole cut occurs as combine should wait until all chunks calculate their averages to process the final result. 

Probably it is reimplemented within the code but I do not see it. This looks like a part of a bigger job based on `distribute` method so its a part of a bigger job. As `distribute` with its unit tests is not in the master, I can not evaluate what it is doing. 

In addition, I do not see any point in splitting `cut_dnd`

PR #1044, opened 2023-06-16
1021 tgp serializable class help

tgperring:
Refactor serializable class
- split the classdef file into individual methods
- add an introduction to the classdef documentation to give an overview and explain what a developer needs to do to use serializable
- add copious documentation to all the methods
- add new method 'convert_old_struct' to make writing converter from old class definitions to new much clearer
- change variable names to be clearer (e.g. 'inputs' , which is an input structure, to S; logical return flag from eq method renamed from 'is' to 'iseq' to match the convention of intrinsic Matlab eq functions)
-


herbert_core/utilities/classes/@serializable/serializable.m
```diff
@@ -1,249 +1,353 @@
...
-            struc = to_struct(obj);
-            ser_data = serialise(struc);
-        end
+    methods (Static)
+        % Restore object or object array from a structure created by to_struct.
+        % The optional input object over-rides the object class type held in the
+        % structure, and provides any missing fields if recovering from an
+        % incomplete structure.
+        obj = from_struct (S, obj_template)
+    end
```
abuts, line 261: I would write varargin here. The form makes you think that `obj_template` is a mandatory property while it is rarely used optimization property

abuts (approved):
looks fine. I've added minor comments, which hope can be useful

PR #1049, opened 2023-06-19
1047 fix var names

### LINKED ISSUE #1047: Consistent variable names on configs
Standardise naming schemes

oerc0122:
Standardise names between different configurations:
- `par_threads` -> `parallel_threads`
- `known_clust_configs` -> `known_cluster_configs`
- Removed `accumulating_process_num`
- Moved `wkdir_is_default` and `is_auto_parallel_threads` to hidden blocks
- Made internal references to other configs be the actual name of the config.

Fixes #1047

### DIFF
--- a/_test/test_mpi_wrappers/test_parallel_config_and_clusters_factory.m
+++ b/_test/test_mpi_wrappers/test_parallel_config_and_clusters_factory.m
@@ -182,7 +182,7 @@ classdef test_parallel_config_and_clusters_factory < TestCase
 
             pc.parallel_cluster='her';
             assertEqual(pc.parallel_cluster,'herbert');
-            all_clcfg = pc.known_clust_configs;
+            all_clcfg = pc.known_cluster_configs;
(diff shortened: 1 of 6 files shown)


abuts (changes requested):
I do not see one thing here -- `opt_config_manager` and default computer configurations. They contan these or some of these properties names. 

If I am missing it or there are no such properties, then fine. If not, need change the names of properties there too. 
Otherwise -- fine

tgperring (approved):
The names look good to me.

oerc0122:
> I do not see one thing here -- `opt_config_manager` and default computer configurations. They contan these or some of these properties names.
> 
> If I am missing it or there are no such properties, then fine. If not, need change the names of properties there too. Otherwise -- fine

@abuts After double checking, none of the parameters changed in this PR are in `OptimalConfigInfo.xml`, `accumulating_process_num` does not appear in the source of any `*.*` file

PR #1050, opened 2023-06-19
Fix overriden `delete` operation

### LINKED ISSUE #1045: `delete` overridden in `test_equal_to_tol`

oerc0122:
Fix override of delete in `test_sqw_class/test_equal_to_tol` by calling superclass method

Fixes #1045

### DIFF
--- a/_test/test_sqw_class/common_sqw_class_state_holder.m
+++ b/_test/test_sqw_class/common_sqw_class_state_holder.m
@@ -3,7 +3,7 @@ classdef common_sqw_class_state_holder < common_state_holder
     % completed
     properties
     end
-    
+
     methods
         function obj = common_sqw_class_state_holder()
             %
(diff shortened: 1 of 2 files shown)


abuts (approved):
As far as I remember, unlike C++ `delete` does not need overriding, and each level of inheritane chain calls its delete if it is defined there (child delete does not suppresses parent delete). 

Seen it somewhere. Wagualy remember. Anyway, overriding does not do any harm so looks ok. 

Yep, verified it
"If you create a hierarchy of classes, each class can define its own delete method. When destroying an object, MATLAB calls the delete method of each class in the hierarchy. Defining a delete method in a handle subclass does not override the handle class delete method. Subclass delete methods augment the superclass delete methods."
so this is unnecessary. Superclass delete will be called twice. 
Will do no harm.

Issue #1022, opened 2023-05-31
`Out of Memory` in `file_access_3_3`


oerc0122:
In `sqw_binfile_common:410`, the line
```
            bytes = fread(obj.file_id_,fn_size,'char');
```
Results in an `Out of Memory` error on IDAAaS with a 200GiB file, while having over 100GiB of free memory. 

```
>> obj.file_id_
   ans = 4
>> fn_size
   ans = 30
```

Unsure why, but was fixed locally by replacing the line with:
```
            bytes = fread(obj.file_id_,fn_size,'integer*1');
```

PR #1075, opened 2023-07-03
faccess fixture

abuts:
simple solution of #1022 using fopen operation which defines char as 8-bits symbols


cmarooney-stfc (changes requested):

horace_core/algorithms/@cut_data_from_file_job/private/put_sqw_data_npix_and_pix_to_file_.m
```diff
@@ -41,7 +41,7 @@
     if exist(outfile,'file') == 2
         delete(outfile);
     end
-    fid=fopen(outfile,'A');    % no automatic flushing: can be faster
+    fid=fopen(outfile,'A','l','Windows-1252');    % no automatic flushing: can be faster
```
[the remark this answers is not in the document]
oerc0122: As mentioned before, I agree with @cmarooney-stfc . If we want to read N bytes, we should use the appropriate length types to read them rather than setting an encoding. Any of the following (https://uk.mathworks.com/help/matlab/ref/fread.html#btp1twt-1-precision):
```
uint8
int8
integer*1
uchar
unsigned char
ubit8
schar
signed char
bit8
char*1
```
Would work. It should match the type you're actually trying to read to make it clear, I would recommend `integer*1` since this is the appropriate MATLAB type.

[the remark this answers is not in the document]
abuts: it is fully portable. The suggestion applies to read, and I applied the formatting to open, making `char == char*1`. There are subtle differences in parsing read data for different types of input and I do not want suddently hit these subtletly

horace_core/sqw/file_io/@horace_binfile_interface/private/read_header_.m
```diff
@@ -43,7 +43,7 @@
     fid=file_name;  % copy fid
     do_fseek(fid,0,'bof');  % set the file position indicator to the start of the file
 else
-    fid=fopen(file_name,permission_req);
+    fid=fopen(file_name,permission_req,'l');
```
oerc0122, line 46: Why no encoding?

horace_core/sqw/PixelData/@PixelDataFileBacked/PixelDataFileBacked.m
```diff
@@ -174,79 +174,11 @@
             % 8: 3xNpix array
```
oerc0122, line 174: Shouldn't be included in this PR

abuts: That's the conflict resolution (and subtle bug in init)

oerc0122: It still shouldn't be included in this PR

abuts: there are no other way of resolving conflicts

oerc0122:
```
git checkout master horace_core/sqw/PixelData/@PixelDataFileBacked/PixelDataFileBacked.m
git commit --amend
git push -f
```

abuts: So, I would be pushing to master invalid (outdated) code? I do not want this -- I will certainly forget to change it to recent version

oerc0122: Where are you pushing outdated code to master? What are you talking about? This pulls the original file from master and puts it in this branch. If you need to make the change of `sqw_fopen` to the file, then you can do so before you commit. If you want to keep the other changes then you can check this filee out to another branch and open that as a PR.

abuts: I have no idea what the problem is, but I've seen the conflict in PixelDataFileBacked and resolved it with the latest changes to code.

oerc0122: Because it's a significant change not relevant to this PR

abuts: If I see the issue, I am fixing it whatever possible and convenient. The relevance is very relative thing.

PR #1087, opened 2023-07-07
Fix join

### LINKED ISSUE #531: SQW join method removes all pixel data
The `sqw.join( )` method builds a new SQW object from a number of pieces. 

The PixelData is aggregated using the `PixelData.cat(...)` method which creates a correctly sized, but incorrectly sorted PixelData object. The final operation of the join method calls `cut` to resort this data. 

Adding debugging to the code demonstrates that after this cut is performed the PixelData object is empty.

The use of a cut to perform the resort if commented:
```
% It seems that sorting of the pix field is necessary, however I'm not
% aware of an easy way of accomplishing this.
% Re-performing a cut with
%   wout=cut_sqw(wout,proj,[],[],[],[]);
% does the trick, but that is projection-specific and it might not be right
% to build up a proj struct from the information present in a sqw.
```

As tests for these methods are disabled, the methods look broken on 04/08/22

oerc0122:
Modernises join and makes it work with the new sqw structures.

Fixes #531

### DIFF
--- a/horace_core/sqw/@sqw/join.m
+++ b/horace_core/sqw/@sqw/join.m
@@ -52,12 +52,12 @@ clear main_header;
 clear wi_main_header;
 
 % Start pulling in data
-header = cell(size(w));
+experiment = cell(size(w));
 detpar = cell(size(w));
 data = cell(size(w));
 pix = cell(size(w));
 for i = 1:nfiles
-    header{i} = w(i).header; % Will be used as is
+    experiment{i} = w(i).experiment_info; % Will be used as is
     detpar{i} = w(i).detpar; % Needs to be reduced to only a single struct
     data{i} = w(i).data;     % Needs to be combined
     pix{i} = w(i).pix;
@@ -86,13 +86,15 @@ main_header.nfiles = sum(run_contributes); % For the output structure
 
 data = data(run_contributes);
 pix = pix(run_contributes);
+experiment = experiment(run_contributes);
 
 for i = 1:numel(pix)
     pix{i}.run_idx = i; % repopulate individual run numbers
 end
 
 wout.main_header = main_header;
-wout.experiment_info = Experiment(header); % This should be a cell array of the individual headers
+ % This should be a cell array of the individual headers
+wout.experiment_info = Experiment.combine_experiments(experiment, false, true);
 wout.detpar = detpar0;
 
 wout.data = data{1};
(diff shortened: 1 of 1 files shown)


horace_core/sqw/@sqw/join.m
```diff
@@ -17,162 +17,134 @@
...
-    detpar{i}=w(i).detpar; % Needs to be reduced to only a single struct
-    data{i}=w(i).data;     % Needs to be combined
-    pix{i} =w(i).pix;
+header = cell(size(w));
+detpar = cell(size(w));
+data = cell(size(w));
+pix = cell(size(w));
+for i = 1:nfiles
+    header{i} = w(i).header; % Will be used as is
+    detpar{i} = w(i).detpar; % Needs to be reduced to only a single struct
```
abuts, line 64: `w(i).header` is legacy interface. Should use experiment_info instead

horace_core/sqw/@sqw/join.m
```diff
@@ -17,162 +17,134 @@
...
-for i=1:length(rc_idx)
-    pix{rc_idx(i)}.run_idx=i; % repopulate individual run numbers
+for i = 1:length(rc_idx)
+    pix{rc_idx(i)}.run_idx = i; % repopulate individual run numbers
 end
 
-
 % Now I'm not entirely sure how to proceed. So I'll stab blindly and hope
 % that recombining the data.s, data.e, data.npix, and data.pix arrays and
 % then using recompute_bin_data will do the trick.
```
abuts, line 97: that's irrelevant -- the fixture looks correct, so this comment is outdated

horace_core/sqw/@sqw/join.m
```diff
@@ -17,162 +17,134 @@
...
+
+% build a new PixelData object from the contributing files data
+% pix_ex = cellfun(@(x) x, pix(run_contributes), 'UniformOutput', false);
+
+ind = zeros(sz);
+num_pix = sum(wout.data.npix, 'all');
+wout.pix = PixelDataMemory(num_pix);
+
+for i = 1:nfiles
+    if run_contributes(i)
```
abuts, line 130: very inefficient implementation. This one is slow up to uslesness due to the way Matlab treats loops and array indexing. Much more reasonable would be swapping order of loops, i.e. collecting all contributed pixels from different files and bunching them all together to put them into target file. 

Would assume this is a quick fix to solve the problem, but more efficient implementations are already in Horace -- `write_nsqw_to_sqw` works (Maltab and C++ implementations) do the same for files and expanding them to memory based implementation is a trivial excersize. 

(May also be beneficial to look on Matlab `write_nsqw_to_sqw` from point of view of using generic interface to PixelData -- then the combine code will be applicable in both places)

Inabilty to use existing highly optimized code in favour of custom half-baked solutions is depressing

abuts (commented):
Very poor code practice, using poorly optimized custom solutions instead of proven higly optimized code existing in other places. 

From other point of view, looks like does the job of splitting/combining. Due to time constrain probably should be left as it is.

PR #1095, opened 2023-07-10
1066 Add filebacking to section

### LINKED ISSUE #1066: Section needs filebacked implementation
`sqw.section` currently loads selected pixels into memory.

oerc0122:
Add filebacked calculation capabilities to `sqw.section` to filter pixels, along with corresponding unit test.

Fixes #1066

### DIFF
--- a/horace_core/sqw/@sqw/section.m
+++ b/horace_core/sqw/@sqw/section.m
@@ -77,6 +77,7 @@ for n = 1:numel(win)
 
             pix_ind = 1;
 
+            new_pix_range = [];
             for iter = 1:num_chunks
                 chunk = block_chunks{iter};
                 pix_start = chunk{1};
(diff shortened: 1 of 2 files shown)


cmarooney-stfc (approved):

horace_core/sqw/@sqw/section.m
```diff
@@ -67,7 +67,34 @@
...
+            pix_ind = 1;
+
+            for iter = 1:num_chunks
+                chunk = block_chunks{iter};
+                pix_start = chunk{1};
+                block_sizes = chunk{2};
+
+                data = win(n).pix.get_pix_in_ranges(pix_start, block_sizes, false, false);
+
+                wout(n).pix.format_dump_data(data.data, pix_ind);
```
abuts, line 87: All operations which take partial pixels must recalculate pixel data averages.  For memory backed it happens automatically when your assing new data to `PixedDataMemory` but here it has to be present. 

I hope pix API does everything else correctly

abuts (changes requested):
needs `pix.data_range`  recalclated during paging operations unless dumping API does this. (Have not seen it before and dumping does not look like good candidate to do this_

horace_core/sqw/@sqw/section.m
oerc0122: Section doesn't change the bins, it merely extracts them, so this is taking whole bins and the bin averages don't need recalculating. The `data_range` is updated in the underlying `DnDBase.section`

abuts (approved):
This should work if handles work correctly

PR #1098, opened 2023-07-10
Make symmetrise operate in projection basis

### LINKED ISSUE #1039: symmetrise relative to projection
Currently symmetrise reduces the dataset and considers the irreducible wedge according to `[1, 0, 0]` in the absolute cartesian frame rather than the plane of projection. Users expect the projection to be considered in symmetrise as it is in cut.

oerc0122:
Transform symmetry irreducible region according to current projection to symmetrise in the basis of the current projection rather than the cartesian lattice basis. 

Only currently considering lattice-based projections (`ortho_proj`) may have trouble with others. Will consult and add warning as necessary.

Needs unit tests to prove

Fixes #1039

### DIFF
--- a/horace_core/sqw/@sqw/symmetrise_sqw.m
+++ b/horace_core/sqw/@sqw/symmetrise_sqw.m
@@ -98,9 +98,10 @@ if isa(sym, 'SymopReflection')
 
     for i = 1:fold
         % transform existing range into transformed range
-        idx = ~sym(i).in_irreducible(cc_exist_range);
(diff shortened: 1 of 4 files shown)


abuts (commented):
It looks ok for pixels, except if pixels are filebacked changes should happen for image too and it is not here. 
May be it is done somewhere else and not the topic of this ticket, but this is very inefficient as forces you to scan pixels twice. 

This is general comment for filebacked operations. Image changes should work alongside to pix changes

PR #1109, opened 2023-07-12
Add `PixelDataFileBacked.ready_dump`

oerc0122:
Add `ready_dump` public function for use where external functions need to check internal `file_handle_`


horace_core/sqw/PixelData/@PixelDataFileBacked/PixelDataFileBacked.m
```diff
@@ -283,11 +287,22 @@ function saveobj(~)
...
+
+        function has = get.has_open_file_handle(obj)
+            has = ~isempty(obj.file_handle_);
+        end
     end
 
     %======================================================================
     % File handling/migration
     methods
+        function obj = ready_dump(obj)
```
cmarooney-stfc, line 299: not at all clear why it's this name. If you can't find a better one, maybe comments to explain it

cmarooney-stfc (changes requested):

abuts (commented):
Theoretically no objections against this, except it fails 

More importantly, I do not believe handle can be obtained on pixels only except for testing purposes. 

It is probably

PR #1110, opened 2023-07-12
Add file-backed scale factor to hor_config

### LINKED ISSUE #1108: Move `scale_fac` into config
Currently the factor determining whether to memory-back or file-back is stored as a constant in `PixelDataBase.do_filebacked` this should be moved instead into one of the configs (`hor` or `hpc`; likely `hor` due to `mem_chunk_size` also being in `hor_config`)

oerc0122:
Move the file-backed scale factor to `hor_config`

Fixes #1108

### DIFF
--- a/horace_core/configuration/@hor_config/hor_config.m
+++ b/horace_core/configuration/@hor_config/hor_config.m
@@ -47,6 +47,14 @@ classdef hor_config < config_base
         % with ~4Gb it has to be reduced to 10^6
         mem_chunk_size;
 
+        % Factor for determining whether to do certain operations
+        % as file-backed or memory-backed. The value is roughly equivalent
+        % to the number of "pages" which would need be produced before considering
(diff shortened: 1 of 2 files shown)


cmarooney-stfc (approved):

horace_core/configuration/@hor_config/hor_config.m
```diff
@@ -122,6 +130,7 @@
     properties(Access=protected, Hidden=true)
         % private properties behind public interface
         mem_chunk_size_ = 1e7;
+        fb_scale_factor_ = 3;
```
abuts, line 134: To maintain previous behaviour as much as possible, I would make it 10, as 10 chunks was comfortable value to make memory based cuts

abuts (commented):
looks good. Need to have 10 as default for not to change previous Horace behaviour too much.

PR #1134, opened 2023-07-18
Reformat `calculate_qw_pixels2` and make file-safe

### LINKED ISSUE #1125: calculate_qw_pixels2.m

oerc0122:
- Use `get_fields` to extract whole arrays
- General non-function changing cleanup to whitespace/error handling/etc.

Fixes #1125

### DIFF
--- a/horace_core/sqw/@sqw/calculate_qw_pixels2.m
+++ b/horace_core/sqw/@sqw/calculate_qw_pixels2.m
@@ -1,7 +1,7 @@
 function qw=calculate_qw_pixels2(win)
 % Calculate qh,qk,ql,en for the pixels in an sqw dataset from the headers
 %
-%   >> qw=calculate_qw_pixels(win)
+%   >> qw = calculate_qw_pixels2(win)
 %
 % This method differs from calculate_qw_pixels because it recomputes the values
 % of momentum and energy from efix, emode and the detector information. This is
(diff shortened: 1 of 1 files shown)


horace_core/sqw/@sqw/calculate_qw_pixels2.m
```diff
@@ -140,14 +155,12 @@
...
 qtmp(1,:) = ki' + qtmp(1,:);            % qspec proper now
 qtmp = mtimesx_horace (spec_to_rlu,reshape(qtmp,[3,1,numel(ki)]));
 qtmp = squeeze(qtmp);
 
 % Package output
-q = cell(1,3);
-q{1} = qtmp(1,:)';
-q{2} = qtmp(2,:)';
-q{3} = qtmp(3,:)';
+num2cell(qtmp', 1);
```
cmarooney-stfc, line 164: apparent q=missing
```suggestion
q=num2cell(qtmp', 1);
```

cmarooney-stfc (changes requested):

abuts (approved):
This is old code which should be moved to `instr_porj` As such, a lot of work should be done to implement this 
Substantial bug to fix would be usage `Experiment.runid_map` to build correspondence between runid-s and and number of header to extract run parameters.  In current form it will fail on 50% of modern data.

Other changes, i.e. use `Experiment` instead of header are less important and may be the part of technical dept of moving from function on sqw obhect ti `instr_proj` (there is also substantial code overlap with `gen_sqw` or rather `rundatah.build_sqw`)

abuts (commented):
This is old code which should be moved to `instr_porj` As such, a lot of work should be done to implement this 
Substantial bug to fix would be usage `Experiment.runid_map` to build correspondence between runid-s and and number of header to extract run parameters.  In current form it will fail on 50% of modern data.

Other changes, i.e. use `Experiment` instead of header are less important and may be the part of technical dept of moving from function on sqw obhect ti `instr_proj` (there is also substantial code overlap with `gen_sqw` or rather `rundatah.build_sqw`)

PR #1144, opened 2023-07-20
Refactor & rename `signal` to `coordinates_calc`; file-back it

oerc0122:
File-back `signal`, rename to `coordinates_calc`, fix long-standing bug in `average_bin_data`

### DIFF
--- a/_test/test_sqw_class/test_sqw_signal.m
+++ b/_test/test_sqw_class/test_sqw_signal.m
@@ -20,44 +20,60 @@ classdef test_sqw_signal < TestCaseWithSave
             obj.sqw_2d = read_sqw(test_sqw_2d_fullpath);
             obj.save();
         end
+
+        function test_signal_fb(obj)
+            clObj = set_temporary_config_options(hor_config, 'mem_chunk_size', 1000);
+
+            w1modE = coordinates_calc(obj.sqw_2d, 'E');
+
+            fb_sqw = obj.sqw_2d.copy();
+            fb_sqw.pix = PixelDataFileBacked(fb_sqw.pix);
+
+            assertTrue(fb_sqw.pix.is_filebacked);
+
+            w1modE_fb = coordinates_calc(obj.sqw_2d, 'E');
+            assertEqualToTol(w1modE, w1modE_fb, '-ignore_date');
+
+        end
+
         function test_w2E_option(obj)
-            w1modE = signal(obj.sqw_2d,'E');
+            w1modE = coordinates_calc(obj.sqw_2d,'E');
 
             assertEqualToTolWithSave(obj,w1modE,'ignore_str',true, ...
                 'tol',[1.e-9,1.e-9],'-ignore_date');
-            
+
         end
 
         function test_w2Q_option(obj)
-            w1modQ = signal(obj.sqw_2d,'Q');
+            w1modQ = coordinates_calc(obj.sqw_2d,'Q');
 
             assertEqualToTolWithSave(obj,w1modQ,'ignore_str',true, ...
                 'tol',[3.e-7,3.e-7],'-ignore_date');
-            
+
         end
 
         function test_w2l_option(obj)
-            w2modL = signal(obj.sqw_2d,'l');
+            w2modL = coordinates_calc(obj.sqw_2d,'l');
 
             assertEqualToTolWithSave(obj,w2modL,'ignore_str',true, ...
                 'tol',[1.e-9,1.e-9],'-ignore_date');
 
         end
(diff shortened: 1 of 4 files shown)


_test/test_sqw_class/test_sqw_signal.m
```diff
@@ -20,44 +20,60 @@
             obj.sqw_2d = read_sqw(test_sqw_2d_fullpath);
             obj.save();
         end
+
+        function test_signal_fb(obj)
+            clObj = set_temporary_config_options(hor_config, 'mem_chunk_size', 1000);
+
+            w1modE = coordinates_calc(obj.sqw_2d, 'E');
+
+            fb_sqw = obj.sqw_2d.copy();
```
cmarooney-stfc, line 29: this is slightly confusing as initially after the copy fb_sqw is not filebacked; that happens on the next line. given that the idiom is reasonable matlab, maybe just a comment that initially fb_sqw is not filebacked

_test/test_sqw_class/test_sqw_signal.m
```diff
@@ -20,44 +20,60 @@
...
+            fb_sqw.pix = PixelDataFileBacked(fb_sqw.pix);
+
+            assertTrue(fb_sqw.pix.is_filebacked);
+
+            w1modE_fb = coordinates_calc(obj.sqw_2d, 'E');
```
cmarooney-stfc, line 34: I don't think this is filebacked, it is the same as L.27. Do you need fb_sqw as the first arg?

horace_core/sqw/@sqw/coordinates_calc.m
```diff
@@ -0,0 +1,189 @@
+function w = coordinates_calc(w, name)
+% Replace the sqw's signal and variance data with coordinate values (see below)
```
cmarooney-stfc, line 2: should be worth commenting that this is the function formerly known as signal, for traceability.

Also at this point "the sqw" is not defined; you mean w. Can we also be a DND? (see comment below)

horace_core/sqw/@sqw/coordinates_calc.m
```diff
@@ -0,0 +1,189 @@
...
+%               'E'                 energy transfer
+%               'Q'                 |Q|
+%
+% Output:
+% -------
+%   w    Output sqw object with the signal and variance of the pixels set to the
+%          value of the input parameter name (also updates DND)
+%
+% EXAMPLE
+%   >> wout=coordinates_calc(w,'h')   % set the intensity to the component along a*
```
cmarooney-stfc, line 23:
```suggestion
%   >> wout=coordinates_calc(w,'h')   % set the intensity to the Q-component along a*.
```

horace_core/sqw/@sqw/coordinates_calc.m
```diff
@@ -0,0 +1,189 @@
...
+%               'd1','d2',...       Display axes (for as many dimensions as
+%                                  the sqw object has)
+%               'h', 'k', 'l'       r.l.u.
+%               'E'                 energy transfer
+%               'Q'                 |Q|
+%
+% Output:
+% -------
+%   w    Output sqw object with the signal and variance of the pixels set to the
+%          value of the input parameter name (also updates DND)
```
cmarooney-stfc, line 20: really unclear what (also updates DND) means. You could mean that for DNDs, the image intensity is altered in this way; you could mean that for SQWs the DND component has its intensity altered; or both. Worth being detailed and explicit.

horace_core/sqw/@sqw/coordinates_calc.m
```diff
@@ -0,0 +1,189 @@
+function w = coordinates_calc(w, name)
+% Replace the sqw's signal and variance data with coordinate values (see below)
+%
+%   >> wout=coordinates_calc(w, name)
+%
+% Input:
+% -----
+%   w       Input sqw object
```
cmarooney-stfc, line 8:
```suggestion
%   w       Input sqw object  or array of objects
```

cmarooney-stfc (changes requested):

_test/test_sqw_class/test_sqw_signal.m
oerc0122: You are correct, fixed.

horace_core/sqw/@sqw/coordinates_calc.m
oerc0122: This is a method on `sqw`. "the sqw" is the `sqw` you called this from.

Will add reference to signal.

PR #1151, opened 2023-07-26
Add function to apply general operations to pixels

### LINKED ISSUE #1150: Make method to apply a function to pixels
Create a function which applies an arbitrary function to a pixeldata/sqw object and updates the relevant info.

oerc0122:
Applies a function/cell-array of functions to an object with pixels and recomputes the DnD if requested (defaults to `true` on SQWs as safer, but slower option)

Fixes #1150

### DIFF
--- a/horace_core/sqw/PixelData/@PixelDataFileBacked/apply.m
+++ b/horace_core/sqw/PixelData/@PixelDataFileBacked/apply.m
@@ -1,0 +1,108 @@
+function [obj, data] = apply(obj, func_handle, args, data)
+% Apply a function (`func_handle`) to pixels (`obj`) with extra arguments `args`
+%  and recomputes the DnD if provided in `data`
+%
+% Inputs:
+%
+%   obj
+%        PixelDataFileBacked object
+%
+%   func_handle
+%        Function handle or cell array of function handles to apply
+%        `func_handle` must have a signature corresponding to:
+%
+%        pix_obj = func_handle(pix_obj, args{1}, ..., args{N})
+%
+%   args
+%        cell-array of extra args to pass to `func_handle`
+%        args can either be a cell array of args for all func_handles
+%        or a cell-array of cell-arrays matching the length of func_handles
+%        to provide args to each function
+%
+%   data
+%        DnD object whose data are to be recomputed according to the
+%        result of `func_handle(obj,...)`
+
+
+    if ~exist('args', 'var') || isempty(args)
+        args = {{}};
+    end
+
+    if ~iscell(args)
+        args = {{args}};
+    end
+
+    if isa(func_handle, 'function_handle')
+        func_handle = {func_handle};
+    end
+
+    if numel(args) == 1
+        args = repmat(args, numel(func_handle), 1);
+    elseif numel(args) ~= numel(func_handle)
+        error('HORACE:apply:invalid_argument', ...
+              'Number of arguments does not match function handle')
+    end
+
+    if exist('data', 'var')
+        [obj, data] = apply_op_dnd(obj, func_handle, args, data);
+    else
+        obj = apply_op_no_dnd(obj, func_handle, args);
+    end
+end
+
+function [obj, data] = apply_op_dnd(obj, func_handle, args, data)
+
+    obj = obj.prepare_dump();
+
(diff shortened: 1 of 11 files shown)


abuts (approved):
Looks reasonable except function handle interface is very limited
And unit tests show that as you have to introduce a class, which then called through function handle and applied. 

Why not to introudce `PixOpBase` class, which by default calculate pix data averages, but should be overloaded to satisfy all various requests and change apply to work with this class. 

`apply` insufficient for binary operations but extension is obvious

horace_core/sqw/PixelData/@PixelDataMemory/apply.m
```diff
@@ -0,0 +1,53 @@
...
+    if ~exist('args', 'var') || isempty(args)
+        args = {{}};
+    end
+
+    if ~iscell(args)
+        args = {{args}};
+    end
+    if isa(func_handle, 'function_handle')
+        func_handle = {func_handle};
+    end
```
cmarooney-stfc, line 34: excuse my paranoia but if func_handle is an array of function handles, do you need to expand it before making a cell of it

horace_core/sqw/PixelData/@PixelDataMemory/apply.m
```diff
@@ -0,0 +1,53 @@
...
+
+    if numel(args) == 1
+        args = repmat(args, numel(func_handle), 1);
+    elseif numel(args) ~= numel(func_handle)
+        error('HORACE:apply:invalid_argument', ...
+              'Number of arguments does not match function handle')
```
cmarooney-stfc, line 40:
```suggestion
              'Number of arguments does not match number of function handles')
```

horace_core/sqw/PixelData/@PixelDataFileBacked/apply.m
```diff
@@ -0,0 +1,108 @@
...
+        args = {{}};
+    end
+
+    if ~iscell(args)
+        args = {{args}};
+    end
+
+    if isa(func_handle, 'function_handle')
+        func_handle = {func_handle};
+    end
```
cmarooney-stfc, line 37: see memory version

horace_core/sqw/PixelData/@PixelDataFileBacked/apply.m
```diff
@@ -0,0 +1,108 @@
...
+
+    if numel(args) == 1
+        args = repmat(args, numel(func_handle), 1);
+    elseif numel(args) ~= numel(func_handle)
+        error('HORACE:apply:invalid_argument', ...
+              'Number of arguments does not match function handle')
```
cmarooney-stfc, line 43: see memory version

cmarooney-stfc (changes requested):

horace_core/sqw/PixelData/@PixelDataMemory/apply.m
oerc0122: If it's a cell array it will not get here. You cannot have an array of function handles. 
https://uk.mathworks.com/help/matlab/matlab_prog/fundamental-matlab-classes.html

horace_core/sqw/PixelData/@PixelDataFileBacked/apply.m
oerc0122:
```suggestion
              'Number of arguments does not match number of function handles')
```

oerc0122:
> Looks reasonable except function handle interface is very limited And unit tests show that as you have to introduce a class, which then called through function handle and applied.
> 
> Why not to introudce `PixOpBase` class, which by default calculate pix data averages, but should be overloaded to satisfy all various requests and change apply to work with this class.
> 
> `apply` insufficient for binary operations but extension is obvious

@abuts What class is being introduced? The `sigvar` call in the tests is simply to exactly reproduce what `do_unary_op` does and is entirely unnecessary in practice. The code simply applies a function which takes/returns a pix object. 

I really don't see what benefits a `PixOpBase` class would provide over a simple method besides specifying options in a `struct` like way which I could implement trivially anyway.

I think at the moment, special casing operations between two `sqw` objects is fine as we have to loop through both. We can simply replace `do_unary_op`, but I don't see a need at the moment.

cmarooney-stfc (approved):

PR #1169, opened 2023-08-06
1146 make object lookup serializable

tgperring:
Made object_lookup serializable


herbert_core/utilities/classes/@object_lookup/private/parse_sz_repmat.m
```diff
@@ -0,0 +1,57 @@
...
+if ~iscell(sz_repmat)
+    sz_repmat = {sz_repmat};    
+end
+
+% Check input is not empty
+nsz_repmat = numel(sz_repmat);
+if nsz_repmat==0
+    error('HERBERT:object_lookup:invalid_argument', ...
+        'If it has been given, the set of repeat sizes cannot be empty')
+end
```
oerc0122, line 17: This doesn't do what you think. If `sz_repmat = []`, this will still respond positively because of `{sz_repmat}`

herbert_core/utilities/classes/@object_lookup/private/parse_sz_repmat.m
```diff
@@ -0,0 +1,57 @@
...
+
+% Check each input is a valid single expansion argument to repmat
+is_size_vector = @(x)(isnumeric(x) && isrow(x) && numel(x)>=1 && ...
+    all(x>=0) && all(rem(x,1)==0));
+if ~all (cellfun(@(x)is_size_vector(x), sz_repmat), 'all')
+    error('HERBERT:object_lookup:invalid_argument', ...
+        'Repeat sizes must all be an integer>=0 or a valid Matlab array size')
+end
```
oerc0122, line 25: See `isindex` function `herbert_core/utilities/general/isindex.m`

herbert_core/utilities/classes/@object_lookup/private/parse_sz_repmat.m
```diff
@@ -0,0 +1,57 @@
...
+
+
+%-------------------------------------------------------------------------------
+function sz_out = make_size_vector (sz_in)
+% Make the size vector for a square matrix if a scalar input
+if numel(sz_in)==1
+    sz_out = [sz_in, sz_in];
+else
+    sz_out = sz_in;
+end
```
oerc0122, line 57: `repmat` does this implicitly, why do this here?
```
>> q = repmat(1, 3)

q =

     1     1     1
     1     1     1
     1     1     1
```

Also, you could just have:
```suggestion
function sz = make_size_vector (sz)
% Make the size vector for a square matrix if a scalar input
if isscalar(sz)
    sz = [sz, sz];
end
```

herbert_core/utilities/classes/@object_lookup/private/parse_sz_repmat.m
```diff
@@ -0,0 +1,57 @@
+function sz_repmat_out = parse_sz_repmat (sz_repmat, n_objArr, ntot_objArr)
+% Check that the repmat option for implicitly expanding the stored arrays in an
+% object_lookup object is valid. Throw an error if not.
+%
+%   >> sz_repmat_out = parse_sz_repmat (sz_repmat, n_objArr, ntot_objArr)
```
oerc0122, line 5: I do not see what benefit this function adds beyond what repmat does.

herbert_core/utilities/classes/@object_lookup/object_repmat.m
```diff
@@ -0,0 +1,177 @@
...
+if ~obj.filled
+    error('HERBERT:object_lookup:invalid_argument', ...
+        'The object_lookup is not initialised')
+end
+
+% Parse input
+% -----------
+% Determine number and type OK
+narg = numel(varargin);
+if ~(narg==2 || narg==3) || ~ischar(varargin{end-1})
```
oerc0122, line 81: This check for `char` is obscuring what you actually want to check. Check explicitly for `-repeat` here.

herbert_core/utilities/classes/@object_lookup/object_repmat.m
```diff
@@ -0,0 +1,177 @@
+function obj_out = object_repmat (obj, varargin)
+% Return a given object array from the original set of arrays
+%
+% Apply to a specific array or set of arrays in in the object_lookup:
+%   >> obj_out = object_repmat (obj, iarray, '-repeat', sz_repmat)
+%
+% Apply to the whole object_lookup:
+%   >> obj_out = object_repmat (obj, '-repeat', sz_repmat)
```
oerc0122, line 8: `-arg` is reserved for `true`/`false` arguments. 
https://github.com/pace-neutrons/Horace/blob/master/documentation/adr/0022-use-keyword-args.md

herbert_core/utilities/classes/@object_lookup/object_repmat.m
```diff
@@ -0,0 +1,177 @@
...
+% Expand arrays
+% -------------
+% Get indexing arrays and sizes of stored arrays
+sz = obj.sz_;
+indx = obj.indx_;
+
+% Reshape the index arrays to the sizes of the object arrays to which repmat
+% will be applied
+sz_sel = sz(iarray);
+indx_sel = cellfun (@(x,y)reshape(x,y), indx(iarray), sz_sel, 'uniformoutput', false);
```
oerc0122, line 115: Don't need anonymous function for calling with same args
```suggestion
indx_sel = cellfun (@reshape, indx(iarray), sz_sel, 'uniformoutput', false);
```

herbert_core/utilities/classes/@object_lookup/object_lookup.m
```diff
@@ -31,61 +45,51 @@
...
-    %   rand_ind        - generate random points for indexed occurences in object_lookup
-    %
     % See also pdf_table_lookup
     
     properties (Access=private)
-        % Class version number
-        class_version_ = 1;
+        % Object array of unique instance of objects in the input array or cell array
+        % (Column vector)
+        object_store_ = zeros(0,1)
```
oerc0122, line 53: Don't need to initialise to 0d object generally. MATLAB defaults to `[]`

herbert_core/utilities/classes/@object_lookup/object_lookup.m
```diff
@@ -149,484 +182,374 @@
...
-                
-            elseif nargin>0
-                if nargin==1 || (nargin==3 && ischar(varargin{1}) &&  numel(varargin{1})>=2 ...
-                        && strncmpi(varargin{1},'-repeat',numel(varargin{1})))
-                    % Input can only be one of the forms:
+            if nargin==0
+                % Default constructor - do nothing
+
+            else
+                if (nargin==2 || nargin==3) && all(cellfun (@isnumeric, varargin(2:end)))
```
oerc0122, line 202:
```suggestion
            else if (nargin==2 || nargin==3) && all(cellfun (@isnumeric, varargin(2:end)))
```

herbert_core/utilities/classes/@object_lookup/object_lookup.m
```diff
@@ -149,484 +182,374 @@
...
+            % arguments) to pass.
+            if ~iscell(val)
+                val = {val};    % turn into a cell array with one element
+            end
+            positive_int_array = @(x)(all(x>=1,'all') && all(rem(x,1)==0, 'all'));
+            if ~all (cellfun (@isnumeric, val), 'all') ||...
+                    ~all (cellfun (@(x)positive_int_array(x), val), 'all')
+                error('HERBERT:object_lookup:invalid_argument', ...
+                    'Property ''indx'' must be a cell array of integer arrays')
+            end
```
oerc0122, line 388:
```suggestion
            positive_int_array = @(x)(isnumeric(x) && all(x>=1,'all') && all(rem(x,1)==0, 'all'));
            if ~all (cellfun (@(x)positive_int_array(x), val), 'all')
                error('HERBERT:object_lookup:invalid_argument', ...
                    'Property ''indx'' must be a cell array of integer arrays')
            end
```

herbert_core/utilities/classes/@object_lookup/object_lookup.m
```diff
@@ -149,484 +182,374 @@
...
+            
+            % Stored object arrays can have zero size
+            
+            if ~iscell(val)
+                val = {val};    % turn into a cell array with one element
+            end
+            
+            size_array = @(x)(isrow(x) && numel(x)>=2 && all(x>=0) && all(rem(x,1)==0));
+            if ~all (cellfun (@isnumeric, val), 'all') ||...
+                    ~all (cellfun (@(x)size_array(x), val), 'all')
```
oerc0122, line 422: See `isindex`

herbert_core/utilities/classes/@object_lookup/object_lookup.m
```diff
@@ -149,484 +182,374 @@
...
                 error('HERBERT:object_lookup:invalid_argument', ...
-                    'Replacement for property ''object_store'' must be scalar or have the same number of objects')
+                    'Property ''sz'' must be a cell array of valid Matlab array sizes')
+            end
+            
+            % Trim excess trailing singleton dimensions in the size vectors, and
+            % make the set of indices a column cell array
+            sz_trim = @(sz)(sz(1:max([2, find(sz~=1,1,'last')])));
+            obj.sz_ = cellfun (@(x)(sz_trim(x)), val(:), 'uniformOutput', false);
```
oerc0122, line 430: Isn't this `squeeze`?

herbert_core/utilities/classes/@object_lookup/object_lookup.m
```diff
@@ -149,484 +182,374 @@
...
-            %
-            % See also structIndep, structPublic, structArrPublic
             
-            if numel(obj)>1
-                S = arrayfun(@fill_it, obj);
-            else
-                S = structIndep(obj);
+            % Check the sizes of the stored object arrays are consistent with the
+            % length of the index arrays
+            nel = cellfun(@prod, obj.sz_);
```
oerc0122, line 506:
```suggestion
            nel = obj.nelmts;
```

herbert_core/utilities/classes/@object_lookup/object_lookup.m
```diff
@@ -149,484 +182,374 @@
...
+            n_indx = cellfun(@numel, obj.indx_);
```
oerc0122, line 507: Use public accessors unless you want to explicitly check internals.
```suggestion
            n_indx = cellfun(@numel, obj.indx);
```

oerc0122 (changes requested):

cmarooney-stfc (approved):

herbert_core/utilities/classes/@object_lookup/private/parse_sz_repmat.m
tgperring: Not quite the same as what I want. isindex will pass if the input is a column, or empty 1x0 or 0x1, or logical, whereas I want a numeric array length at least one of +ve integers

herbert_core/utilities/classes/@object_lookup/private/parse_sz_repmat.m
tgperring: The code does an implicit repmat: a stored object array in object_lookup is not fed into repmat because the unique elements are already stored. Instead, what is being done is the stored size array is altered (hence the need to get the size of the result if repmat *were* to be applied), and the index arrays into the stored object list are also adjusted.

herbert_core/utilities/classes/@object_lookup/private/parse_sz_repmat.m
tgperring: See above. The point about object_lookup is to only keep the minimum storage.

herbert_core/utilities/classes/@object_lookup/object_repmat.m
```diff
@@ -0,0 +1,177 @@
+function obj_out = object_repmat (obj, varargin)
```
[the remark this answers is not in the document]
tgperring: Some changes now made. The operation of object_repmat does more than a repmat on a single object - it can perform an *implicit* repmat on a subset of a collection of objects. However, I agree that the '-repeat' keyword syntax is an unnecessary confusion and the keyword itself can be removed to leave the syntax closer to that of of a vanilla repmat. Object_repmat does need the indices of the subset of stored object arrays as the second argument (after the object_lookup object itself), but '-repmat' is truly unnecessary.  I left it from an earlier version so as not to require changing a couple of Tobyfit routines.

herbert_core/utilities/classes/@object_lookup/object_repmat.m
tgperring: Comment out of date given a change made to the syntax of object_lookup.

herbert_core/utilities/classes/@object_lookup/object_repmat.m
tgperring: Comment out of date given a change made to the syntax of object_lookup.

herbert_core/utilities/classes/@object_lookup/object_repmat.m
```diff
@@ -0,0 +1,177 @@
...
+% -----------
+% Determine number and type OK
+narg = numel(varargin);
+if ~(narg==2 || narg==3) || ~ischar(varargin{end-1})
+    error('HERBERT:object_lookup:invalid_argument', ...
+        'Invalid number or type of input argument(s)')
+end
+
+% Get iarray
+if narg==3
```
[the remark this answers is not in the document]
tgperring: The value of iarray has to be assigned, and the assignment depends on narg.

herbert_core/utilities/classes/@object_lookup/object_lookup.m
tgperring: I explicitly want the size [0,1] rather than [0,0]

herbert_core/utilities/classes/@object_lookup/object_repmat.m
tgperring: I didn't know that.  It makes code tidier. Now adopted.

herbert_core/utilities/classes/@object_lookup/object_repmat.m
```diff
@@ -0,0 +1,177 @@
...
+        % numel(sz_repmat). Therefore simply reassign sz and indx
+        sz = cellfun(@(x)size_repmat(sz{1},x), sz_repmat,...
+            'uniformOutput', false);
+        indx = cellfun(@(x)reshape(repmat(indx_sel{1},x),[],1), sz_repmat,...
+            'uniformOutput', false);
+    else
+        sz{iarray} = size_repmat(sz{1},sz_repmat{1});
+        indx{iarray} = reshape(repmat(indx_sel{1},sz_repmat{1}),[],1);
+    end
+end
```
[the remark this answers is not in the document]
tgperring: Hmm.... it looks like I've got no choice. The Matlab documentation explicitly states (R2018a)

"A = cellfun(func,C1,...,Cn) applies func to the contents of the cells of C1,...,Cn, so that A(i) = func(C1{i},...,Cn{i}). The function func must take n input arguments and return a scalar. The cell arrays C1,...,Cn all must have the same size."

Just in case Matlab has introduced cellarray expansion I tested for:
         c = cellfun (@horzcat, a, b, 'uniformOutput', false)
With 
      a = {[2,3], [101,102,103]}          % cell array, two elements
      b = {[11,12,13]}                       % cell array one element

I get the error message:
Error using cellfun
All of the input arguments must be of the same size and shape.
Previous inputs had size 2 in dimension 2. Input #3 has size 1

I tried with R2023a (the latest version) too; it gives the same error message.

herbert_core/utilities/classes/@object_lookup/object_lookup.m
tgperring: The suggested code does not have the desired functionality. The original code first checks that all elements of the cellarray are numeric; only then does it test that they are integer. In the suggested code  the comparison all(x>=1,'all')  would be attempted regardless of the classes of the elements of the cellarray, which depending on the class could throw an error (e.g. an element itself being a cell array) or unexpected behaviour (e.g. character array).

herbert_core/utilities/classes/@object_lookup/object_lookup.m
tgperring: Earlier comment re isindex not doing precisely what I want also applies here

herbert_core/utilities/classes/@object_lookup/object_lookup.m
tgperring: Squeeze also gets rid of singletons in the middle of an array, whereas I just want to remove trailing singletons beyond the first two
x = rand(3,1,5)
I still want to keep the size as [3,1,5], but size(squeeze(x)) is [3,5]

herbert_core/utilities/classes/@object_lookup/object_lookup.m
tgperring: I've generally preferred to use the private properties directly rather than the public getters on efficiency grounds, but this is actually something I want to talk with you about on general design principles and private .v. public properties coming out of thinking about the design of serializable. I'm leaving as is for the time-being until I have a better mental model.

herbert_core/utilities/classes/@object_lookup/object_lookup.m
tgperring: See reply to previous comment.

oerc0122 (approved):

PR #1188, opened 2023-08-16
Fix symmetrise with proj

### LINKED ISSUE #1191: Symmetrise does not work with cellarray of symops
Symmetrise does not work with cellarray of symops and as such not with `SymopRotation.fold`

oerc0122:
Rotational symmetry requires properly defined projection to work, default to hkl basis.

Add tests for rotational reduction (just of functionality, need proper data tests later)

Fixes #1191

### DIFF
--- a/horace_core/sqw/@sqw/symmetrise_sqw.m
+++ b/horace_core/sqw/@sqw/symmetrise_sqw.m
@@ -55,17 +55,19 @@ if ~has_pixels(win)
           'input object must be sqw type with detector pixel information');
 end
 
-if isa(varargin{end}, 'aProjection')
+if isa(varargin{end}, 'aProjectionBase')
     proj = varargin(end);
     varargin = varargin(1:end-1);
 
-    if ~proj{1}.nonorthogonal
+    if proj{1}.nonorthogonal
         error('HORACE:symmetrise_sqw:invalid_argument', ...
               'Cannot symmetrise to non-orthogonal projection');
     end
 
 else
-    proj = {};
+    proj = {line_proj([1 0 0], [0 1 0], ...
+                      'alatt', win.data.proj.alatt, ...
+                      'angdeg', win.data.proj.angdeg)};
 end
 
 
@@ -89,7 +91,7 @@ wout = copy(win);
 
 [sym, fold] = validate_sym(sym);
 transforms = arrayfun(@(x) @x.transform_pix, sym, 'UniformOutput', false);
-wout = wout.apply(transforms, {proj{:}}, false);
+wout = wout.apply(transforms, {{{proj{:}}}}, false);
 
 %=========================================================================
 % Transform Ranges:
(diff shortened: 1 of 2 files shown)


horace_core/sqw/@sqw/symmetrise_sqw.m
```diff
@@ -89,7 +92,7 @@
 
 [sym, fold] = validate_sym(sym);
 transforms = arrayfun(@(x) @x.transform_pix, sym, 'UniformOutput', false);
-wout = wout.apply(transforms, {proj{:}}, false);
+wout = wout.apply(transforms, {{{proj{:}}}}, false);
```
cmarooney-stfc, line 95: A comment to explain what this is doing would improve the maintainability

horace_core/sqw/@sqw/symmetrise_sqw.m
```diff
@@ -216,9 +219,10 @@
 
     elseif iscell(sym)
         % If it's come from SymopRotation.fold or manual equivalent
-        if all(cellfun(@(x) isa(x, {'SymopRotation', 'SymopIdentity'}), sym))
-            if ~sym{1} == SymopIdentity() || ...
-                        isa(sym{1}, 'SymopRotation') && sym{1}.theta_deg ~= 0
+        if all(cellfun(@(x) isa(x, 'SymopRotation') || ...
+                            isa(x, 'SymopIdentity'), sym))
+            if sym{1} ~= SymopIdentity() || ...
+                         isa(sym{1}, 'SymopRotation') && sym{1}.theta_deg ~= 0
```
cmarooney-stfc, line 225: I'm generally not following this. Options for sym{1} are identity or non-zero rotation. Assuming identity as a special case, the remaining sym{2:end} should follow the recipe below on lines 244-7. If sym{1} is a rotation, then for those lines, shouldn't the loop go from 2:fold, not 3:fold? Either way, more explanation in comments please.

horace_core/sqw/@sqw/symmetrise_sqw.m
```diff
@@ -237,10 +241,10 @@
                        'Expected: %d, received: %d', fold, numel(sym)])
             end
 
-            for i = 1:fold-1
+            for i = 3:fold
                 % If not regular fold
-                if sym{i+1}.theta_deg ~= sym{2}.theta_deg || ...
-                        sym{i+1}.theta_deg / fold ~= sym{2}.theta_deg
+                if sym{i}.theta_deg ~= sym{2}.theta_deg && ...
+                        sym{i}.theta_deg / (i-1) ~= sym{2}.theta_deg
```
cmarooney-stfc, line 247: It appears there are 2 legal cases - either all the syms have the same theta OR they increase uniformly, x, 2x, 3x, 4x etc. This condition is an AND - which may be ok as neither condition is met, but not so readable for maintenance. 

Also the use of integer division with reals possibly rounded makes me nervous.

Also it is unclear what happens to sym{1} in either of its cases.

At least more explanation please, also check for errors.

cmarooney-stfc (changes requested):

cmarooney-stfc (approved):

PR #1196, opened 2023-08-22
Fix alignment issues in both fb and mb forms

oerc0122:
- Use apply rather than separate function
- Add (logging) printing to apply


horace_core/sqw/@sqw/apply_alignment.m
```diff
@@ -26,40 +28,44 @@
...
+    % unlike majority of other operations, save would not be effective
+    % here, as image have not changed and have already been saved while
+    % getting new handle. Just swap pixels to take the new file as a
+    % source.
+    pix.delete();
+    del_memmapfile_files(old_file)
+    [ok,mess]=movefile(new_file,old_file,'f');
+    if ~ok
+        warning('HORACE:file_locked', ...
+                'Can not replace old file %s with new file %s. Continuing with new file', ...
```
cmarooney-stfc, line 53:
```suggestion
                'Can not replace old file %s with new file %s. Continuing with old file', ...
```

cmarooney-stfc (changes requested):

abuts (commented):
need to familarize myself more with `apply` Otherwise, alignment looks incorrect as data ranges are not initalized properly while assigning to data

horace_core/sqw/PixelData/@PixelDataFileBacked/apply.m
abuts, line 1: I am working on fixing issue with data range, but to make current code reasobavle, the following change should be done to `apply`, rows 89-90 (new file):
```
        obj.data_range = ...
obj.pix_minmax_ranges(curr_pix.data, obj.data_range_)
```
(add underscore to exisitng data range)

oerc0122: Is that including if we nullify the `data_range` at the start of the operation?

E.g. l. 108
```
obj.data_range = PixelDataBase.EMPTY_RANGE;
```

abuts: Initial nulification looks fine. Problems with 89-90 is that currently call to obj.data_range will cause recalculation of the whole data range. 
This will be changed in a nearest future, but meanwhile accessing private member will achieve the desirable goal.

oerc0122: Updated, will that be ok?

abuts (approved):
looks fine for me

cmarooney-stfc (approved):

PR #1206, opened 2023-08-31
Fix loading sqw on cut; always fileback

### LINKED ISSUE #1202: Entire sqw file appears to be loaded when making a cut
Profiling and testing the Horace tutorial for making the iron sqw file from 46 .spe files and taking a small 1D cut appears to show that the whole sqw file (5GB on disk ==> 10GB if fully loaded in memory) is being read before the method that actually performs the cut is called.  There appears to be a related issue in Horace 3.6.3 in that a page as defined by mem_chunk_size is read - not the whole file but still a significant fraction. For the 1D cut that is created in the tutorial script this amounts of an overall time to make the cut increasing by a factor of 5 or more. Further investigation is needed to see if the problem scales to realistic size sqw files.

oerc0122:
Also update upgrade message

Fixes #1202


cmarooney-stfc (approved):

tgperring:
Looks OK to me

PR #1219, opened 2023-09-07
Remove redundant "tail" argument from memmap loading

oerc0122:


abuts (changes requested):
This change is incorrect and harmful. 
As soon as pixels are last dataset in sqw file, all this works. As soon as you write something after pixels when updating some datasets, it fails with very strange error if `tail` is not present.

It is of course very bad that unit tests do not cach this change, though I believe I was writing some tests for that. 
This ticket may be modified to write unit tests which check tail presence necessary

oerc0122:
> This change is incorrect and harmful. As soon as pixels are last dataset in sqw file, all this works. As soon as you write something after pixels when updating some datasets, it fails with very strange error if `tail` is not present.
> 
> It is of course very bad that unit tests do not cach this change, though I believe I was writing some tests for that. This ticket may be modified to write unit tests which check tail presence necessary

That sounds like it's masking some other error and we should **definitely** not have this. Retry since I've rewritten the PixelData dumpers.

oerc0122:
And thus it is clearly masking a deeper problem. This is not a solution.

abuts:
closed as redundant and unnecessary

PR #1246, opened 2023-09-13
version switch

abuts:
The changes seems address errors with new configuration and version switch observed on IDAAAS


admin/horace_on.m.template
```diff
@@ -10,8 +10,26 @@ function path=horace4_on(non_default_hor_path,non_default_her_path)
 %                                         of Horace/Herbert pairs)
 %
 %
+warn_state=warning('off','all');    % turn of warnings (so don't get errors if remove non-existent paths)
```
oerc0122, line 13:
```suggestion
warn_state=warning('off','all');    % turn off warnings (so don't get errors if removing non-existent paths)
```

admin/horace_on.m.template
```diff
@@ -10,8 +10,26 @@ function path=horace4_on(non_default_hor_path,non_default_her_path)
...
+try
+    horace_off();
+    warning(warn_state);    % return warnings to initial state
+    evalin('base','clearvars')
+    clearvars -global;
+    clear classes;
+catch
+    warning(warn_state);    % return warnings to initial state
+end
+warning(warn_state);    % return warnings to initial state
```
oerc0122, line 29:
```suggestion
try
    horace_off();
catch
end

warning(warn_state);    % return warnings to initial state

evalin('base','clearvars')
clearvars -global;
clear classes;
```
The reset in the `catch` is unnecessary as it will move onto the next line anyway.

Keep the scope of `try`s as small as possible and only capture potential failures.

herbert_core/admin/@opt_config_manager/opt_config_manager.m
```diff
@@ -58,17 +61,20 @@
         %
         this_pc_memory_ = 8*1024*1024*1024; %default PC memory is 8Gb
     end
+    properties(Constant,Access=private)
+        % different pc types, we have optimizezed Horace/Herbert for.
```
oerc0122, line 65:
```suggestion
        % different pc types, where we have estimated specs for Horace.         
```

herbert_core/admin/@opt_config_manager/opt_config_manager.m
```diff
@@ -131,25 +137,30 @@
...
             % the type can be set by name (from the list of the names
             % specified in the list definition) or by the number of the pc
             % type in the same list
             if isnumeric(val)
-                n_types = numel(obj.known_pc_types_);
+                n_types = numel(obj.pc_types_);
                 if val>0 && val<=n_types
-                    pc_type = obj.known_pc_types_{val};
+                    val = obj.pc_types_{val};
+                    pc_type = obj.known_pc_types_(val);
```
oerc0122, line 154:
```suggestion
                    pc_type = obj.known_pc_types(val);
```

herbert_core/admin/@opt_config_manager/opt_config_manager.m
```diff
@@ -131,25 +137,30 @@
...
+
         function obj = set.this_pc_type(obj,val)
             % explicitly setting pc type for testing or debugging purposes.
             %
             % the type can be set by name (from the list of the names
             % specified in the list definition) or by the number of the pc
             % type in the same list
             if isnumeric(val)
-                n_types = numel(obj.known_pc_types_);
+                n_types = numel(obj.pc_types_);
```
oerc0122, line 151: Probably want public getter for all of these.

herbert_core/admin/@opt_config_manager/private/find_comp_type_.m
```diff
@@ -17,29 +17,29 @@
...
+        pc_type = types('jenkins_win');  % 'jenkins_win'
     end
 elseif isunix
     if ismac %MAC
-        pc_type = types{3};
+        pc_type = types('a_mac');
         return;
     end
     [nok,mess] = system('lscpu');
     if nok  %still MAC or strange unix without lscpu. Assuming mac.
```
oerc0122, line 41: Issue a `warning` here.

herbert_core/classes/MPIFramework/@parallel_config/parallel_config.m
```diff
@@ -494,9 +494,12 @@
             n_poss_threads = floor(obj.n_cores/n_workers);
 
             if n_threads < 0
-                error('HERBERT:parallel_config:invalid_argument', 'parallel_threads must be positive or 0 (automatic)')
+                error('HERBERT:parallel_config:invalid_argument', ...
+                    'parallel_threads must be positive or 0 (automatic)')
             elseif n_threads > n_poss_threads
-                warning('HERBERT:parallel_config:parallel_threads', 'Number of par threads (%d) might exceed computer capacity (%d)', n_threads, n_poss_threads)
+                warning('HERBERT:parallel_config:parallel_threads', ...
+                    'Total number of threads used by all parallel workers n_treads*n_workers = (%d) might exceed computer capacity (%d)', ...
```
oerc0122, line 501:
```suggestion
                    'Total number of threads used by all parallel workers n_threads*n_workers = (%d) might exceed computer capacity (%d)', ...
```

oerc0122 (changes requested):

admin/horace_on.m.template
abuts: clearvars will remove `warn_state`  variable, so suggested row 25 will fail

oerc0122: I'm not sure what you mean. The `clearvars` happens after line 25. Otherwise the original code would fail.

abuts: clearing all variables when intitializing Horace is a nuclear options, and we do not want to do it unless it is absolutely necessary. Unfortunately, it appears to be that it is necessary if Horace version have changed. Otherwise, we want to keep our workspaces intact.

herbert_core/admin/@opt_config_manager/opt_config_manager.m
```diff
@@ -58,17 +61,20 @@
         %
         this_pc_memory_ = 8*1024*1024*1024; %default PC memory is 8Gb
     end
+    properties(Constant,Access=private)
+        % different pc types, we have optimizezed Horace/Herbert for.         
+        pc_types_ = {'win_small','win_large','a_mac',...
+            'unix_small','unix_large',...
+            'idaaas_small','idaaas_large', 'jenkins_win', 'jenkins_unix'};
+        known_pc_types_ = containers.Map(opt_config_manager.pc_types_, ...
+            opt_config_manager.pc_types_) ;
```
cmarooney-stfc, line 70: this appears to be a map to itself

cmarooney-stfc (changes requested):

herbert_core/admin/@opt_config_manager/opt_config_manager.m
abuts: well, using private property within class method is a good practice. No need to go through public interface which is often slower.

herbert_core/admin/@opt_config_manager/opt_config_manager.m
abuts: yes, Better solution exist but not worth spending time on it.

Issue #848, opened 2022-09-06
unit tests and clarification for replicate_array


abuts:
there are two very different versions or replicate_array function present in Horace and Herbert. Are these versions the same or do they do different things? 
In addition to that, the function is not tested at all. 

Unit tests should be written and the `replicate_array` code verified.

CM: 18/9/23:
Reviewing the 2 functions identified here, being herbert_core/utilities/maths/replicate_array.m and horace_core/sqw/@sqw/private/replicate_array.m, the two functions do do the same thing. Their methodology is slightly different. The herbert one calculates a series of positions in an array large enough to hold the replicate, and then fills that array between these positions; while the horace one creates a set of sub-arrays, one for each replications, and then concatenates them. Both have the same special case that if the replication numbers argument is empty, an empty integers array (size 0,1) is created, which is not documented. Toby is original author of both.

TGP 09/02/24:
It turns out that at in R2015a, well after Horace was originally written, that Matlab introduced a function repelem that performs essentially the same operation. Cursory tests at the command line suggest that the only difference is in the shape of the output array: repelem preserves the shape of the input array, just expanded along dimensions, whereas replicate_array always produces a column vector. That is, 
          vout = replicate_array (v, n)
is equivalent to 
          vout = repelem (v(:), n(:))
Timing tests across a wide range of vector v lengths and average values of elements of n show that 
(1) The sqw/private method is between 2 and 8 times slower than the Herbert version
(2) repelem is between 20% slower and a factor of 25 times faster than the Herbert version

The correct solution will be either to simply wrap repelem to form replicate_array, replicate_iarray and replicate_logarray, possibly even substitute the calls throughout Horace and Herbert with calls to repelem.

tgperring:
This issue was resolved in PR #1543

PR #1284, opened 2023-09-18
Test herbert and horace versions of replicate_array

cmarooney-stfc:
Visual inspection of the 2 replicate_array methods in horace and herbert indiciate that they have identical functionality although they are differently implemented; one uses an array_func to make individual replicas which are concatenated, while the other makes a large array, works out the replica positions and fills them.

This PR provides a test for the herbert version. The horace version is tested similarly by providing a public static interface function in the sqw.m file itself. It may be that this needs to be retired after this testing to preserve the integrity of the sqw class.

Alex claims that the horace version is twice as fast. Additional timings for this test suggest that for this size test the timing can vary either way so that issue is not dealt with here.

Related: #848

### DIFF
--- a/_test/test_algorithms/test_replicate.m
+++ b/_test/test_algorithms/test_replicate.m
@@ -1,0 +1,53 @@
+classdef test_replicate < TestCase
+    
+    properties
+    end
+    
+    methods
+        
+        function obj = test_replicate(varargin)
+            if nargin > 0
+                name = varargin{1};
+            else
+                name= mfilename('class');
+            end
+            obj = obj@TestCase(name);
+        end
+        
+        function test_replicate_herbert(~)
+            % test the herbert replicate_array
+            
+            % array of numbers of replicates to be added
+            np = [1 2 3 4 5 6 7 8 9 10];
+            % array of things to be replicated
+            vv = [111 222 333 444 555 666 777 888 999 1000];
+            % replication of vv by np using the herbert/utilities/maths
+            % version of replicate array
+            ww = replicate_array(vv,np);
+
+            % check size of replicated array
+            assertEqual( numel(ww), 55);
+            % check last members of replications are correct
+            assertEqual([ww(1),ww(3),ww(6),ww(10),ww(15), ...
+                         ww(21),ww(28),ww(36),ww(45),ww(55)], ...
+                         [111 222 333 444 555 666 777 888 999 1000]);
+            % check first members of replications are correct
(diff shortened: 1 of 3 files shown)


abuts (changes requested):
I think we have verified the code and have proved that the version which is in `@sqw/private` is less efficient then one in `math/utilities`. 
The private version should go and everything should go through the public version. 
They are equivalent otherwise -- no point of keeping less efficient version

tgperring:
I think this PR is redundant. While doing a code review of a PR from Jacob recently I saw he had used a matlab intrinsic function repelem (introduced in R2015b) that performs in essence the same function as replicate_array, replicate_iarray and replicate_logarray. I performed fairly comprehensive timing tests and found that depending on input arguments it is anywhere between a factor 1 and 25 times faster than the faster of the two replicate_array versions here (which is in turn anywhere between a factor of 2 and 8 times faster than the slower version). I've updated ticket #848 with more details.

abuts:
The changes are not about `repelem`. Here you combine multiple chunks in a single array, i.e. multiple pixel data arrays into signle pixel array. There are two ways of doing this: create chunks and `concat` them or allocate whole array and place chunks into appropriate positions. We have identified that the second way is faster and here we need to change code to use faster method.

tgperring:
Alex, I don't understand your comment. I know that this PR and the ticket #848 are about the function replicate_array, which exists in two places, and work by two methods. In my comment here (and which is elaborated upon in a comment I have added to #848) what I am doing is pointing out that what the function replicate_array does is identical to what the matlab intrinsic function repelem does, and which in fact does it much faster. On my laptop, it is up to 25 times faster than the Herbert version, and up to 200 times faster than the sqw/private version). We don't need to know how it is doing the job, just that it is identical.

More specifically, and subject to checking of edge cases, any call to replicate_array:
:
vout = replicate_array (v, n)
:
can be replaced by:
:
vout = repelem (v(:), n(:))
:

Therefore this PR, whose sole purpose is to add a test routine to check the equivalence of the herbert and sqw/private versions of replicate_iarray is obsolete; the test is simply irrelevant. In a separate branch in reference to ticket #848, we can delete the sqw/private version, keep only the herbert version and replace its internal code with a single line making a call to repelem as above. Unit tests can be added on the same branch. The same can be done to replicate_iarray and replicate_logarray. These functions then become legacy functions and in the future we can just call repelem directly instead.

cmarooney-stfc:
Superseded by PR #1543

Issue #1157, opened 2023-07-28
unique object container contains number of inefficiencies


abuts:
Current implementation of `unique_object` container contains number of inefficiencies. 

1) hash, when object is added to the container, is calculated twice.
2) there is some intermittent issue with contanter initialization after `clear 'all'`
3) The container constantly grows and never shrinks when referring objects get deleted, which forces user to clear up memory manually (see 2)) 
4) saveobj operation saves `unique_references` container despite creating unique objects container anyway (does it restores reference indexes correctly?)
5) It implements singleton behaviour but implementation is different from the standard one, which makes the code difficult to read and understand.

The code would benifit from removing these inefficiencies

19/9/23: CM: noting that from item 5 above, this apparently references a class with singleton behaviour. The unique_*_container using singletons is unique_references_container. Subsequent comments here will assume that is this case, rather than unique-objects_container.
Item 5 is addressed in PR #1285.

cmarooney-stfc:
@abuts apologies, cancel the request for second hash example, it's an artefact of uniqu_references_container

abuts:
should be fixed by Re #1800

PR #1285, opened 2023-09-19
Document use of Matlab singleton pattern in unique_references_container

cmarooney-stfc:
In #1157 it is suggested that the Matlab singleton pattern is not used for unique_references_container. This is not the case, but the pattern is adapted so that a set of templated singletons, one per data type, are stored in the same class.

To document this, as suggested by Jacob, an SMG document, no.22, is added with this PR. The document is referenced in the code for the container.

As code is changed (even though only in comments) the DO_NOT_CI label is not applied.


abuts (approved):
I would rather implement standard pattern instead of discribing personal design, which does not have advantages over standard. 
As we do not have time for that, let the description exist

PR #1306, opened 2023-10-04
Fix issue with multifit on dnd objects

### LINKED ISSUE #1230: `multifit` issue with `dnd`
Christian has reported an issue with `multifit` on `dnd` objects.

```
Unrecognized method, property, or field 'data' for class 'd2d'.

Error in SQWDnDBase/mask_points (line 75)

ndim=length(win.data.pax);
```

oerc0122:
`dnd` objects were being treated in the old (pre-SQW-rewrite) way and `mask_points` hadn't been updated to address this, causing failures in creating an `mfclass` wrapping a `dnd`

Also issue with inverse order of operations on `transform_img_to_hkl`

Fixes #1230

### DIFF
--- a/horace_core/sqw/@SQWDnDBase/mask_points.m
+++ b/horace_core/sqw/@SQWDnDBase/mask_points.m
@@ -78,7 +78,7 @@ else
     data = win;
 end
 
-ndim = length(data.pax);
+ndim = dimensions(data);
 
 if ~isempty(options.keep)
     xkeep=options.keep;
--- a/horace_core/sqw/@SQWDnDBase/mask_points.m
+++ b/horace_core/sqw/@SQWDnDBase/mask_points.m
@@ -72,7 +72,13 @@ if return_with_errors
 end
 
 % Check input parameters
-ndim=length(win.data.pax);
+if isa(win, 'sqw')
+    data = win.data;
+else
+    data = win;
+end
+
+ndim = length(data.pax);
 
 if ~isempty(options.keep)
     xkeep=options.keep;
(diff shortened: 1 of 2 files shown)


horace_core/sqw/@SQWDnDBase/mask_points.m
```diff
@@ -72,7 +72,13 @@
...
 
 % Check input parameters
-ndim=length(win.data.pax);
+if isa(win, 'sqw')
+    data = win.data;
+else
+    data = win;
+end
+
+ndim = length(data.pax);
```
abuts, line 82: There is method `data.NUM_DIMS`  (hidden, do not know why it is hidden -- not my design) which is intended for this. 
Probably it should be used here.

horace_core/sqw/coord_transform/@line_proj/private/transform_img_to_hkl_.m
```diff
@@ -17,7 +17,7 @@
 else
     bmat = obj.bmatrix();
 end
-hkl_to_img = bmat*q_to_img;
+hkl_to_img = q_to_img*bmat;
 offset = bmat\offset_cc;
```
abuts, line 21: I would make a ticket and mark a concern here for further investigation (for me and Toby). 
In `aProjectionBase` class, rows 624-647 is a generic version of transform_img_to_hkl function. 

There is a generic transformation from img to pix and then transformation from pix to hkl. This defines the order of matrix operation which was there before and have been changed by the commit. 

I do not understand the reason why inverse order gives correct answer.

oerc0122: Ticket #1318 created to track.

PR #1327, opened 2023-10-12
unary operations reimplemented

abuts:
Unary operations are reimplemented using `apply`


horace_core/sqw/@sqw/sqw.m
```diff
@@ -429,6 +432,14 @@ function  save_xye(obj,varargin)
             obj.main_header.creation_date = val;
             obj.data.creation_date = val;
         end
+        %
+        function is = get.is_filebacked(obj)
+            if obj.has_pixels
+                is = obj.pix.is_filebacked;
+            else
+                is = true;
```
cmarooney-stfc, line 440: I'm surprised by this. If it has no pixels, then nothing will be stored to file and so it should work in memory. Please comment.

horace_core/sqw/PixelData/@PixelDataBase/PixelDataBase.m
```diff
@@ -674,13 +678,24 @@
...
+
     end
     %----------------------------------------------------------------------
     methods
         % return set of pixels, defined by its indexes
         pix_out = get_pixels(obj, abs_pix_indices,varargin);
-        %
+        %==================================================================
+        % Methods historically on pixels too, but useful mainly on sqw
+        % object
```
cmarooney-stfc, line 696: I don't know what this means. What do you mean by historically? If it is to alert the reader to something that was done before but is no longer done now, it would be useful to be explicit. If something else, needs clarifying.

horace_core/sqw/page_operations/PageOp_unary_op.m
```diff
@@ -0,0 +1,53 @@
...
+            end
+        end
+
+        function obj = apply_op(obj,npix_block,npix_idx)
+            signal = obj.page_data_(obj.signal_idx_,:);
+            var    = obj.page_data_(obj.var_idx_,:);
+            pg_result = obj.op_handle(sigvar(signal, var));
+
+            signal = pg_result.s;
+            var    = pg_result.e;
```
cmarooney-stfc, line 32: any particular reason for using these intermediate variables rather than assigning result directly in the next 2 lines?

horace_core/sqw/@sqw/sqw.m
abuts: yes, this is incorrect. I've already changed this 440 should be `false`

horace_core/sqw/PixelData/@PixelDataBase/PixelDataBase.m
abuts: It is still can be done, but I am not sure if it is any point in doing this. The methods exist on sqw and used and tested there, and here are left because they were there before. Do not know how to express it better.

abuts:
adresses #1324

PR #1331, opened 2023-10-16
Take Tobyfit use detector data from nxspe files

cmarooney-stfc:
Tobyfit has previously reconstructed IX_detector_arrays on the fly from detpar. Horace now processes IX_detector_array info on reading nxspe files, so there can be a different detector for each run. This PR replaces the on-the-fly detector construction with info from the experiment_info's detector arrays.

Note that the number of different detector arrays is still limited to be 1, as more work is needed to deal with the detpar array inputs as they go through dqmat etc and into resconv itself. This is a partial solution needing more input from Toby


abuts (commented):
Difficult to say -- I do not see any tests for this functionality, though big block of code is added to the codebase. 

Usually untested code is broken, so no trust to these changes. As Tobyfit is all like this, would probably accept it for the time being. 

The test failure would probably happen when you add this to sqw object and do not change anythin in sqw loading/saving/updating.

herbert_core/classes/instrument_classes/instrument/detectors/@IX_det_abstractType/IX_det_abstractType.m
```diff
@@ -77,6 +77,46 @@
         % Array of random points
         X = rand (obj, varargin)
     end
+
+    methods (Static)
+        function varargout = tobyfit_defaults(val,parms)
+        % TOBYFIT_DEFAULTS store and make available defaults for detector types
+        % e.g. He3Tube, Tobyfit_Classic, slab... so that they can be used to
+        % initialise detectors, until this information is available in nxspe input
```
oerc0122, line 85: Is this a factory method? If so, the name is confusing. Not sure what this has to do with `defaults`

herbert_core/classes/instrument_classes/instrument/detectors/@IX_det_abstractType/IX_det_abstractType.m
```diff
@@ -77,6 +77,46 @@
...
+        %             In this example case, parms will be a struct with those values
+        %
+        % OUTPUTS
+        % =======
+        % Optional (none, one or two):
+        % varargout{1} - the detector type string, if requested
+        % varargout{2} - a struct with the detector type parameters
+        
+        
+            persistent type parameters;
```
oerc0122, line 105: Why are these `persistent`, not class members? Why would subsequent calls return the last run's inputs?

herbert_core/classes/instrument_classes/instrument/detectors/@IX_det_abstractType/IX_det_abstractType.m
```diff
@@ -77,6 +77,46 @@
...
+            if nargin>0
+                type = val;
+                if nargin>1 && isstruct(parms)
+                    parameters = parms;
+                end
+            end
+            if nargout>0
```
oerc0122, line 112: Personally I would prefer `>= 1` as that's what you actually care about but this is purely stylistic.

herbert_core/classes/instrument_classes/instrument/detectors/@IX_detector_array/IX_detector_array.m
```diff
@@ -162,9 +162,36 @@
...
+                            warning("HORACE:IX_detector_array:empty detector type", ...
+                                    "probably on load from detpar");
+                            %}
+                            obj.det_bank_ = IX_detector_bank ( ...
+                                S.group, S.x2, S.phi, S.azim, ...
+                                IX_det_TobyfitClassic (S.width, S.height));
+                        else
+                            error('HORACE:IX_detector_array:invalid_type', ...
+                                  'unsupported detector type');
+                        end
```
oerc0122, line 193: Could use a `switch` here rather than an `if` chain.
```suggestion
                        switch(dtype)
                          case 'IX_det_He3tube'
                            obj.det_bank_ = IX_detector_bank ( ...
                                S.group, S.x2, S.phi, S.azim, ...
                                IX_det_He3tube (S.width, S.height, ...
                                dparms.wall, dparms.p));
                          case 'IX_det_TobyfitClassic'
                            obj.det_bank_ = IX_detector_bank ( ...
                                S.group, S.x2, S.phi, S.azim, ...
                                IX_det_TobyfitClassic (S.width, S.height));
                          case ''
                            %{
                            % omitting this warning for the moment as it
                            % causes problems with
                            % test_admin/upgrade_file_formats, but it may be
                            % useful to reinstate it at some stage
                            warning("HORACE:IX_detector_array:empty detector type", ...
                                    "probably on load from detpar");
                            %}
                            obj.det_bank_ = IX_detector_bank ( ...
                                S.group, S.x2, S.phi, S.azim, ...
                                IX_det_TobyfitClassic (S.width, S.height));
                          otherwise
                            error('HORACE:IX_detector_array:invalid_type', ...
                                  'unsupported detector type');
                        end
```

horace_core/Tobyfit/instpars_DGdisk.m
```diff
@@ -42,7 +42,7 @@
 
     inst=repmat({instruments{1}},[nrun,1]);
     for i=2:nrun
-        inst{i}=instruments{i};;
+        inst{i}=instruments{i};
     end
```
oerc0122, line 46: What is the difference between this and `inst = instruments(1:nrun)`

herbert_core/utilities/classes/@unique_references_container/unique_references_container.m
```diff
@@ -342,6 +342,14 @@
     methods % overloaded indexers, subsets, find functions
         
         function field_vals = get_unique_field(self, field)
+        %GET_UNIQUE_FIELD each of the unique objects referred to by self
+        % should have a property named 'field'. The code below takes each of the
+        % referred objects in turn, extracts the object referred to by
+        % 'field' and stores it in the unique_OBJECTS_container field_vals
+        % created here. field_vals will then contain unique copies of all
+        % the values of 'field' within the objects referred to in self, indexed
```
oerc0122, line 350: What happens if any of the `field` values are the same? Are those eliminated and the remainder collapsed down?

horace_core/Tobyfit/tobyfit_DGdisk_resconv_init.m
```diff
@@ -259,14 +259,53 @@
...
-    if use_tubes
-        detectors{iw} = IX_detector_array (detpar.group, detpar.x2(:), detpar.phi(:), detpar.azim(:),...
-            IX_det_He3tube (detpar.width, detpar.height, 6.35e-4, 10));   % 10atms, wall thickness=0.635mm
+    det = wtmp.experiment_info.detector_arrays;
+    if isempty(det) || det.n_runs == 0
+        % no detector info was available when the sqw was populated, so
+        % continue with the old detector initialisation from detpar
+        if use_tubes
+            detectors{iw} = IX_detector_array (detpar.group, detpar.x2(:), detpar.phi(:), detpar.azim(:),...
+                IX_det_He3tube (detpar.width, detpar.height, 6.35e-4, 10));   % 10atms, wall thickness=0.635mm
```
oerc0122, line 272: Thanks for the comment, but why these particular numbers? MAPS instrument?

horace_core/Tobyfit/tobyfit_DGdisk_resconv_init.m
```diff
@@ -259,14 +259,53 @@
...
     else
-        detectors{iw} = IX_detector_array (detpar.group, detpar.x2(:), detpar.phi(:), detpar.azim(:),...
-            IX_det_TobyfitClassic (detpar.width, detpar.height));
+        % make a new detector object based on value of use_tubes and insert
+        % it into the detector_array info extracted from the sqw
+        if det.n_unique_objects>1
+            error('HORACE:tobyfit_DGfermi_resconv_init:incorrect_size', ...
+                  ['all sqw runs must have identical detectors with this ', ...
+                   'implementation']);
+        else
```
oerc0122, line 284: Don't strictly need the `else` on an `error`.

horace_core/Tobyfit/tobyfit_DGfermi_resconv_init.m
```diff
@@ -258,15 +258,54 @@
         sample_coords_to_spec_to_rlu(wtmp.experiment_info);
     
     % Get detector information
-    % Because detpar only contains minimal information, hardwire in the detector type here
+    % Because detpar only contains minimal information, either hardwire in
```
oerc0122, line 261: Slightly worrying repetition, but this is in the original and due to be sorted as part of the whole `tobyfit` refactor.

oerc0122 (approved):

horace_core/Tobyfit/tobyfit_DGfermi_resconv_init.m
cmarooney-stfc: @oerc0122 unclear what the repetition is

horace_core/Tobyfit/tobyfit_DGdisk_resconv_init.m
cmarooney-stfc: true

horace_core/Tobyfit/tobyfit_DGdisk_resconv_init.m
cmarooney-stfc: @oerc0122 these numbers are in the original code and in various other tests in the codebase. I am assuming they will be representative of such tubes, and will be overridden by actual data when it is available in nxspe

herbert_core/utilities/classes/@unique_references_container/unique_references_container.m
cmarooney-stfc: @oerc0122 I am unclear what scenario you have in mind. The fields will be actual properties of the classes in the container and will be unique for each stored class object. Thus a fermi chopper instrument will have a field 'moderator' and this method will coy all the moderators and store them in a unique_objects_container indexed by the original instrument indices. So I do not know what 'same' means here

horace_core/Tobyfit/instpars_DGdisk.m
cmarooney-stfc: this is the original code. Your comment sees correct; it is possible this is teed up to add extra manipulation of the instruments. The code will be replaced when the new code is bedded down and connected up to nxspe.

The second ; has been removed

herbert_core/classes/instrument_classes/instrument/detectors/@IX_detector_array/IX_detector_array.m
cmarooney-stfc: yes, one could

herbert_core/classes/instrument_classes/instrument/detectors/@IX_det_abstractType/IX_det_abstractType.m
cmarooney-stfc: As the 2 are equivalent, I presume I care about them equally

herbert_core/classes/instrument_classes/instrument/detectors/@IX_det_abstractType/IX_det_abstractType.m
cmarooney-stfc: additional explanation has been added to the comments for this

Issue #1166, opened 2023-08-03
GUI Fitting


oerc0122:
Currently GUI single peak fit button doesn't work, but users say this is realtively unused and can be hidden for now.

This ticket relates to the hiding and a subsequent ticket should be created to cover the re-enabling or at least discussion around whether the GUI fitting is necessary.

Issue #1160, opened 2023-08-03
EPIC: Fix Horace GUI


oerc0122:
Several GUI features are currently not working. These should be fixed for MVP or soon after. 

- [x] #1161
- [x] #1162
- [x] #1163
- [x] #1164
- [x] #1165
- [x] #1166

PR #1334, opened 2023-10-18
Remove fitting button from GUI

oerc0122:
Disable fitting button to avoid issue with (apparently unused) single-peak fitting.

Addresses #1166 
Addresses #1160


cmarooney-stfc (approved):
Can't do much else

PR #1337, opened 2023-10-19
Forcibly add additional paths when starting parallel worker

### LINKED ISSUE #1292: User scripts for symmetrisation in gen_sqw not found when running clusters
When running `gen_sqw` users can specify an option, `transform_sqw` to apply a transformation, usually a symmetrisation.

The [documentation for this feature](https://pace-neutrons.github.io/Horace/v3.6.3/manual/Generating_SQW_files.html#applying-symmetry-operations-to-an-entire-dataset) states that multiple transformations can be applied using a custom user function. However, this does not work if this user m-file is not present in the default paths as it seems that the ParpoolCluster workers do not inherit the path of the main process.

In such cases, `gen_sqw` fails with an `Undefined function` error.

Original reporter: @mincentatties

Edit: The command `gen_sqw(..., 'transform_sqw',@(x)(user_symmetrisation_routine(x))` works as expected if `user_symmetrisation_routine` is in the current working folder. The `Undefined function` error only occurs if it's on the current user path but is not in the current working folder. I'm not sure if the fix for this should be to change the documentation or to modify the clusters to append the path of the parent node.

Edit: This is in Horace 3 (haven't tested it on Horace 4).

oerc0122:
Fixes #1292

### DIFF
--- a/herbert_core/classes/MPIFramework/@ClusterWrapper/ClusterWrapper.m
+++ b/herbert_core/classes/MPIFramework/@ClusterWrapper/ClusterWrapper.m
@@ -222,6 +222,13 @@ classdef ClusterWrapper
                 obj.common_env_var_('HERBERT_PARALLEL_EXECUTOR') = obj.matlab_starter_;
             end
 
+            % Split path into cell of paths
+            curr_path = split(path, pathsep);
+            % Filter array for added paths
+            curr_path = curr_path(~startsWith(curr_path, matlabroot));
+            % Join back together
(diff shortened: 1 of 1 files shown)


cmarooney-stfc (approved):

abuts (approved):
A lot of efforts were spent on configurations in assumption that the view from a node may be different from the view from the launcher. 

We currently do not have such problem, so change are ok, but I suspec SCARF having this problem.

PR #1353, opened 2023-10-24
apply

### LINKED ISSUE #1346: rewrite apply using page operations
current implementation of `apply` method on sqw uses methods, alternative to `PageOp`. The method itself may still be useful as a way of applying generic function on sqw pixels.  

This ticket is to convert `apply` ino `PageOp`.
Very similar `PageOp_sqw_eval` exist so apply should use it for its operations too.

abuts:
Fixes #1346. 

This is `apply` operation rewritten through the page_op. All its internal filebacking is replaced with the one, used in `apply_op`

Should be looked at after #1326/#1351 are completed and merged


horace_core/sqw/page_operations/PageOp_apply.m
```diff
@@ -53,19 +55,17 @@
             if obj.changes_pix_only
                 return;
             end
-            new_signal     = page_pxls.signal;           
+            new_signal     = page_pxls.signal;
```
oerc0122, line 58: Why have you swapped abbreviation? `pix` everywhere else.

abuts: its `page_pix` property, so `page_pix` -- local variable. but MATLAB does not like variable looking like property so `page_pxls`

oerc0122: This is inconsistent with the rest of Horace. If you have to, call it `local_page_pix`, but don't swap abbreviations like this.

abuts: do not see any problem in naming local variable with sope of one screen any way I like. It is meaningful anyway and have no relations to any other local variables in Horace.

oerc0122: The point is that it is not just you who will have to maintain this code in future and even small changes to consistency like this cause other to have to think twice when they're making changes.

abuts: I here mean anybody. This is local variable with scope of 20 rows. Its meaning is clean so I do not see problem of it being anything.

PR #1366, opened 2023-11-02
Simplify c_deserialise code

### LINKED ISSUE #1328: c_deserialise/c_size failure on empty structure and C++ code quality
The ticket #1324 have identified C++ code crashing on one type of empty structures. 
test `~/Horace/_tests/test_cpp_deserialize:test_ser_empty_det`  rows (579:587) illustrates that. The fixture is provided in PR #1327. 

It looks like it is pretty dirty fixture, introducing a special case. 
In addition, could `c_serial_size` code be suffering from the same problem? Have duplicated tests been added while chacing the solution?

Need careful checks & validation

oerc0122:
Remove special case and standardise `if` block, adding comment to describe reasoning.

This is not an issue in `c_serial_size` as it has the following block:
```
      if (nFields > 0) {
        mxArray* arr = const_cast<mxArray *>(input);
        mxArray* conts;
        mexCallMATLAB(1, &conts, 1, &arr, "struct2cell");
        data_size = get_size(conts);
      }
```

Fixes #1328

### DIFF
--- a/_LowLevelCode/cpp/serialiser/c_deserialize.cpp
+++ b/_LowLevelCode/cpp/serialiser/c_deserialize.cpp
@@ -309,15 +309,10 @@ mxArray* deserialize(uint8_t* data, size_t& memPtr, size_t size, bool recursed) 
     case STRUCT:
     {
         uint32_t nFields;
-        std::vector<uint32_t> fNameLens(0);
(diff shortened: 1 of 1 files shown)


abuts (approved):
If it works, it is good change. And it seems it works.

PR #1368, opened 2023-11-02
Remove spinw related files

### LINKED ISSUE #243: Remove `spinw` integration code
Initially the implementation of the SpinW within Horace was done on the Horace side with several functions in the `utilities` and `functions` folders - some of which are copies of equivalent SpinW functions. This was because the functions were needed quickly for a training course and had not passed review in SpinW and so was not included in the SpinW distribution.

Since then, however, the interfacing functionality has been implemented in a much cleaner fashion within SpinW and does not require these functions any more (although they still work, and there are unit tests for this functionality in Horace - mostly not run since it needs both Horace and SpinW installed).

Since the philosophy for PACE is that Horace should present a uniform API to third party code rather than implement specific bindings to each, and the Horace/SpinW interface functionality is now included in SpinW, the legacy SpinW interface code in Horace should be removed.

oerc0122:
Remove files related to spinw integration from Horace core repo.

Fixes #243


admin/spinw_on.m.template
```diff
@@ -1,75 +0,0 @@
-function path=spinw_on(non_default_path)
```
mducle, line 1: This file should not be removed. It's used on IDAaaS to add SpinW to the path when you do `horace_on`.

mducle (changes requested):
Please revert the removal of the `spinw_on.m.template` file - otherwise it looks good.

admin/spinw_on.m.template
oerc0122: Done that.

mducle (approved):

PR #1369, opened 2023-11-02
Add Matlab wrapper to allow legacy to new multifit

### LINKED ISSUE #1234: Add wrapper for legacy multifit converter to call from matlab
C.f. #1221 #1226

oerc0122:
Add short wrapper script into `admin` to be able to call converter from Matlab

Fixes #1234

### DIFF
--- a/horace_core/admin/mf_legacy_to_new.m
+++ b/horace_core/admin/mf_legacy_to_new.m
@@ -1,7 +1,7 @@
-function conv = mf_leg_to_new(funcline)
+function conv = mf_legacy_to_new(funcline)
 % MATLAB wrapper for python script to modernise multifit syntax
 %
-% >> mf_leg_to_new("[wfit,fitdata]=multifit_sqw(my_new_cut,@sr122_xsec,pars,pfree,pbind,'list',1);")
+% >> mf_legacy_to_new("[wfit,fitdata]=multifit_sqw(my_new_cut,@sr122_xsec,pars,pfree,pbind,'list',1);")
 % ans =
 %
 %    kk = multifit(my_new_cut);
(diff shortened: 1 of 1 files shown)


admin/mf_leg_to_new.m
```diff
@@ -0,0 +1,28 @@
...
+% J. Wilkins 8-9-2023
+
+
+
+    pth = [horace_paths.root, pathsep, 'tools'];
+    if count(py.sys.path, pwd) == 0
+        insert(py.sys.path, int32(0), pth);
+    end
+
+    conv = py.mf_leg_to_new.convert_legacy_multifit(funcline);
```
mducle, line 26: A nicer output might be:

```suggestion
    convstr = strip(char(py.mf_leg_to_new.convert_legacy_multifit(funcline)));
    if nargout < 1
        disp(convstr)
    else
        conv = convstr;
    end
```

mducle (approved):

mducle (changes requested):
Actually I just realised that the `admin/` folder is not added to the Matlab path when you do `horace_init` (or `horace_on` on IDAaaS). So it would be best to put this file in another folder which is on the path, like `horace_core/utilities`

abuts:
There is another `admin` folder, located under `horace_core` (or `herbert_core`) this folder is added to the path while top admin is intended for devalopment and installation scripts

abuts (changes requested):
move it to `horace_core\admin` folder and than it looks fine.

mducle (approved):

horace_core/admin/mf_leg_to_new.m
```diff
@@ -0,0 +1,33 @@
+function conv = mf_leg_to_new(funcline)
```
cmarooney-stfc, line 1:
```suggestion
function conv = mf_legacy_to_new(funcline)
```
Don't think missing out 3 chars is worth the uncertainty in meaning

cmarooney-stfc (changes requested):

PR #1370, opened 2023-11-04
overhaul of binary operations

abuts:
May be blocked by #1377

This is substantial overhaul of binary operations, including fixing filebacked operations.

All classes,  which have operations are now children of single operation interface `data_op_interface`, which defines all operations for `sigvar`, `IX_dataset`(s), `DnDBase` and `sqw` objects.

All various `binary_op_manager`(s) and `unary_op_manager`(s) for these classes are removed in favour of single `binary_op_manager` and `unary_op_manager` in `data_op_interface` with simple overload for `sigvar`

The spaghetti code used in old `op_manager`(s), which defines the order and priority of operations between different objects is replaced by single priority definitions in `data_op_interface`, defining the priority of operations as function of object complexity.  The priorites are testable and tests have been written. 

4 types of operations have been identified and `PageOp` classes were defined for three modes of operations where pixels are involved. 

As the result, we have single code for operations and new `PageOp` operations for data, containing pixels. The code works using the same `apply_op` interface and logic for filebacked and memory based pixels.


_developer_only/@sigvar/aaa_readme.txt
oerc0122, line 1: This should not be done in this PR make it a separate PR

herbert_core/utilities/xml_io_tools/xml_read.m
oerc0122, line 1: Why is this in this PR? Should be separate PR

oerc0122:
There are almost 300 file changes here and you have given no description of what this PR is supposed to do or how it's doing it. A very quick scan reveals a lot of changes completely unrelated to what I think this PR should be. Reduce the effort. Split it up!

herbert_core/utilities/xml_io_tools/xml_read.m
abuts: because if have failed with Matlab 2023b and I need to work with it.

_developer_only/@sigvar/aaa_readme.txt
abuts: can not agree -- I have binary operations overhaul which included sigvar, so those old sigvar is not necessary any more.

oerc0122:
This still doesn't describe the deletions of those files and also doesn't describe the "fix" for Matlab 2023 you mentioned.

abuts:
To explain the deletion of those files I need to introduce you to history. The files in _development were considered as the templates and it was assumed that if you want to write a new class with operations, you will copy these files into new class. 
There were bunch of classes with operations (sigvar, IX_dataset, DnD and SQW) which follow this rule. 

This is ugly and probably incorrect any more. In addition to that, I was searchin for duplicated files and those files were clogging my search. This is why I have deleted the files and deleted all duplicated instances of the same operations for different classes. 

Now we have common interface, single instance of operation function and single operation manager overloaded where necessary. 

I did 3 iterations until got to the state where I can reliably remove all old stuff and  avoid the code duplication, present in the previous version, instead of need to fix all code duplication and writing unit test for that if I've stopped at the first iteration.

abuts:
As for xml fix -- the code does not work with Matlab 2023b and I fixed it as by-product of other development. Its trivial fix, done in single commit

cmarooney-stfc:
@abuts Alex this PR has not been reviewed. Please revert the merge. Thanks Chris CC @tgperring

herbert_core/utilities/classes/@data_op_interface/unary_op_manager.m
```diff
@@ -27,8 +27,9 @@
 w = w1;
 for i=1:numel(w1)
     %----------------------------------------------------------------------
-    % The following block may be class specific
+    % The following block is overloaded by children classes not easy
+    % convertable to sigvar
```
cmarooney-stfc, line 31: you should say how an individual set of lines, as opposed to a function, will be overloaded

herbert_core/utilities/classes/@sigvar/binary_op_manager_single.m
```diff
@@ -0,0 +1,54 @@
...
+    error('HERBERT:sigvar:invalid_argument', ...
+        'Sizes of signal arrays for operand1(%s) and operand 2(%d) are inconsistent.', ...
+        disp2str(sz1),disp2str(sz2));
+end
+
+if ~isa(w2, 'sigvar')
+    % w1 is an instance of classname, w2 is a double
+    w2 = sigvar(w2);
+end
+if ~isa(w1, 'sigvar') % ? Never happens as this is sigvar method?
```
cmarooney-stfc, line 49: maybe we could resolve this question

herbert_core/utilities/classes/@sigvar/cos.m
```diff
@@ -1,19 +1,22 @@
...
+w1=w;
+s = cos(w.signal_);
+if ~isempty(w.variance_)
+    w1.variance_ = abs(1-s.^2).*w.variance_;     % ensure positive
+else
+    w1.variance_ = [];
+end
+w1.signal_ = s;
+w1.msk     = []; % this is how the initial operation was performed.
+%                  Is it what was intended?
```
cmarooney-stfc, line 22: maybe we could resolve this question - otherwise ticket it for further investigation

cmarooney-stfc (commented):
partial review of 1370 down to and not including SQWDnDBase/unary_op_manager. Please consider this as a Request changes review. Remainer to be looked at shortly.

_LowLevelCode/cpp/serialiser/c_deserialize.cpp
```diff
@@ -309,10 +309,15 @@ mxArray* deserialize(uint8_t* data, size_t& memPtr, size_t size, bool recursed)
     case STRUCT:
     {
         uint32_t nFields;
-
-        deser(data, memPtr, &nFields, types_size[UINT32]);
-        std::vector<uint32_t> fNameLens(nFields);
-        deser(data, memPtr, fNameLens, nFields * types_size[UINT32]);
+        std::vector<uint32_t> fNameLens(0);
+        if ((nDims == 2) && (memPtr >= size)) {
```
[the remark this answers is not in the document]
abuts: I've added my guess why it is necessary but this is just guess.  This change fixes intefimttent crash in deserialization, and chasing it properly is a separate pretty difficult tast. But change here is necessary, as I would be committing broken code otherwise.

_test/test_sqw_pageOpMethods/test_binary_ops.m
```diff
@@ -165,20 +167,48 @@ function test_adding_sqw_and_sigvar_with_sqw_1st_operand_returns_sqw(obj)
                 obj.DOUBLE_REL_TOLERANCE);
         end
 
-        function test_adding_sqw_and_sigvar_with_sqw_2nd_operand_returns_sigvar(obj)
-            signal = 1e5 * (1 + rand(size(obj.sqw_obj.data.s)));
-            variance = 1e5 * (1 + rand(size(obj.sqw_obj.data.s)));
+        function test_adding_sqw_and_sigvar_with_sqw_2nd_operand_returns_sqw(obj)
+            signal = (1 + rand(size(obj.sqw_obj.data.s)));
+            variance =(1 + rand(size(obj.sqw_obj.data.s)));
```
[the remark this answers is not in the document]
abuts: The previous behaviour of the binary operations sigar+sqw was returning sigvar. This have changed  because sqw contains more information and you can always get sigvar from sqw but the opposite is difficult. As the previous test was returning sigvar, it has been constructed to ignore sqw part (sigvar >> sqw), so within the tested tolderance sqw was ignored.  (sqw has 0-s in places where signar does not). 
As behaviour have changed, I've changed the test to have sqw and sigvar to be of the same order to properly evaluate each contribution. 

I do not see why you would write all this in the test as information about previous implementations is irrelevant in a future.

_test/test_sqw_pixels/test_PixelData_binary_ops.m
```diff
@@ -295,21 +355,31 @@ function test_add_npix_neq_num_pixels_memory(~)
...
         function test_add_dnd_neq_num_pixels_filebacked(obj)
             dnd_obj = read_dnd(obj.test_sqw_file_path);
             pix = PixelDataFileBacked(obj.test_sqw_2d_file_path);
             f = @() pix.do_binary_op(dnd_obj, @plus);
-            assertExceptionThrown(f, 'HORACE:PixelDataFileBacked:invalid_argument');
+            ME=assertExceptionThrown(f, 'HORACE:PageOp_binary_sqw_img:invalid_argument');
+            assertTrue(strncmp(ME.message,'binary op: plus',15))
+        end
+        function test_add_dnd_neq_num_pixels_filebacked_reverse(obj)
+            %something wrong with this,
```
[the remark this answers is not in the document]
abuts: That was a thougt that adding pixels to compartible DnD should return `sqw`, not `pix`
The though was ignored as 
1) this is the test for incompartible `dnd`
2) Its beyond of the scope of refactoring existing binary operations.

herbert_core/utilities/classes/@IX_dataset/IX_dataset.m
```diff
@@ -1,4 +1,4 @@
-classdef IX_dataset < serializable & data_op_interface
+classdef IX_dataset < data_op_interface & serializable
```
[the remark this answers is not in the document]
abuts: to be fair, I do not know. serializable looks like higher order inheritance but how to express this on interface -- unknown thing. Here I've decided that higher order inheritance should be furter from the class then the inheritance, more specific to the class, but it may be interpreted other way around. 

I've tried this tought here -- I do not see the difference.

herbert_core/utilities/classes/@IX_dataset/IX_dataset.m
```diff
@@ -180,9 +183,8 @@
         % Rebin an IX_dataset object or array of IX_dataset objects along
         % along the axes, defined by direction
         wout = rebin_xyz(win, array_is_descriptor,dir,varargin)
-        
-        w = unary_op_manager (w1, unary_op)
-        w = binary_op_manager(w1, w2, binary_op)
+
+        %w = unary_op_manager (w1, unary_op)
```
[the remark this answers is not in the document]
abuts: During refactoring, I was not sure if the class needs specific instance of `unary_op_manager` or not. Now I am sure it does not.

herbert_core/utilities/classes/@data_op_interface/unary_op_manager.m
abuts: ? not fully understand what you mean. I've replaced 'overloaded' by reimplemented, as the whoule method is overloaded but code is imlpemented differently

herbert_core/utilities/classes/@sigvar/binary_op_manager_single.m
abuts: shoul we spend ticket on this? This code was there before. I decided to leave it to the future. PR is big enough anyway

herbert_core/utilities/classes/@sigvar/cos.m
abuts: well, I would think that operation should keep the mask, but prevous inplementation was removing it. Probably question to Toby.

cmarooney-stfc: what is needed here is a comment that says what we're going to do, not an open-ended uncertainty. If you're going to ask Toby, ticket it, assign to Toby and refer to it here

herbert_core/utilities/classes/@sigvar/binary_op_manager_single.m
cmarooney-stfc: Comment should say something about how we resolve this. A ticket is as good as any. But if we have a question on this code, it will get lost unless we record it somehow

herbert_core/utilities/classes/@data_op_interface/unary_op_manager.m
cmarooney-stfc: reimplemented doesn't have the same problems as overloaded, so that should do

herbert_core/utilities/classes/@IX_dataset/IX_dataset.m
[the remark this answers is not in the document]
cmarooney-stfc: ok let's not make any more changes like this if it makes no difference

_test/test_sqw_pixels/test_PixelData_binary_ops.m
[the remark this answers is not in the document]
cmarooney-stfc: then we should remove the comment; it's confusing

Issue #1303, opened 2023-10-04
 filebacked Pixel binary operations are broken and non-tested


abuts:
If you look inside the tests for filebacked operations, all testing is disabled and the code for operations is invalid.
This is because many tests are written for PixelData only, Filebacked operations are not tested (or not fully tested -- when I have fundamentallly changed incorrect file hanlde construction -- nothing have changed in tests) and filebacked vs memory backed operations are not fully tested agains each other.

This ticket is fourfold:

1.  test all filebacked operations, ensure they work, check completenes and
2.   collect all filebacked operations into common code, (there are at least dozen places where similar code is used and doing looping over all pixels. This work have been started with writing apply method, but it is incomplete.

 3. Ensure that recent changes to pixel data ranges are incorporated into all pixel operations and pixel_data_range is obtained as byproduct of pixel operations.

  4. Ensure (with tests) that call to invalid data range causes recalculation of the data range where is actually necessary

PR #1375, opened 2023-11-09
1303c Cleanup typos in various files

oerc0122:
Migration of non-functional changes from #1303


abuts (approved):
what a waste of time extracting this into separate PR.  From other side, you spend mine and your time on doing this -- probably believing it is necessary. 

We clearly need discussion about working practices

PR #1377, opened 2023-11-09
1303b Move fake faccess to general function

oerc0122:
Migration of changes to fake_faccess from #1303

May block #1370


_test/test_sqw_pixels/get_pix_with_fake_faccess.m
oerc0122, line 1: Recommend moving this to `_test/common_functions` as it may be useful elsewhere.

cmarooney-stfc (approved):

abuts (approved):
this change breaks 1303 so I've reverted it there. But after it properly merged to master, it would be ok

PR #1378, opened 2023-11-09
1303a Remove sigvar and testsigvar classes from developer only folder

oerc0122:
Migration of removal of files from #1370


cmarooney-stfc (approved):

abuts (approved):
do not know how it necessary to split it into separate PR, but after current #1303 is merged, its fine as this code should not be used any more

PR #1379, opened 2023-11-10
Added PowerPoint presentation on PageOp design

### LINKED ISSUE #1374: Concerns about `PageOp`
Discussion document for `PageOp` refactor. A section on concerns is preceded by
a description of the current and prior implementations for context.

## Current Implementation

`PageOp`-class-based call sequence always follows the same pattern:

* `page_op.init` -- Parse arguments and attach as properties
* `page_op.split_into_pages` -- Binary choice of operations

  - Split on bin-boundaries
  - Split on npix chunks regardless

  Both methods are only one-line callouts to already implemented functions

* iterate over chunks:

  - `page_op.get_page_data` -- Get either one page of single `sqw` or "scroll-locked"
    pair of data for cases of binary ops.
  - `page_op.apply_op` -- method (overridden on each child class) which transforms data
    according to function requirements
  - `page_op.common_page_op` -- conditionally update pix ranges and run indices (if
    any runs have been filtered) and dump to file

* `page_op.finish_op` -- Do any necessary final actions on resulting `sqw`

## Prior implementation

Using `apply` function put the code for `init` and `finish_op` directly inline
in the called function. `apply` split the `sqw` into pages and applied a locally
defined function (effectively a combination of `apply_op` & `common_page_op`) to
the `sqw` object.

**N.B.** There is already a comment in `PageOpBase.m` to the effect of merging
`common_page_op` into `apply_op`

## Concerns

- bin-boundary splitting is implemented on the base class, and will be used
  unless overridden on the child-classes. We end up with the situation of
  having to double-check in a different file which operation is default for
  cases where this is not overridden.

- Duplicates argument parsing and handling already present in called function.

- Delocalisation of code -- Code which only affects the `sqw` object and which
  used to be encapsulated entirely within the `sqw` object is now spread across
  multiple places.

- Undoes unification efforts -- replacing (currently) 9 functions which previously
  were implemented with central call to `apply` function with 9 functions, 9
  classes + 1 base class, each with 3-5 methods on each.

- `PageOp` class structure is textbook example of a
  ["Poltergeist"](https://en.wikipedia.org/wiki/Poltergeist_(computer_programming))
  anti-pattern where a class exists only to operate as a function. The class
  cannot exist independently or be reused.

- Strange call structure -- where SQW calls an object with a reference to itself:

  ```
  page_op = page_op.init(sqw, args);
  sqw = sqw.apply_op(page_op);
  ```

  As one line to highlight strangeness:

  ```
  sqw = *sqw*.apply_op(PageOp(*sqw*, args))
  ```

- Complicates parallelisation -- Previously only needed to send function handle
  and cell array of arguments, `sqw` could be decomposed freely.

  Because `PageOp` stores a reference to `sqw` need to either:

  - construct multiple `PageOp`s to apply to multiple `sqw` segments in advance
    of `sqw.apply_op` which requires duplicated code in each parallelisable
    function
  - because `PageOp` is already constructed as part of the calling function and
    is a particular subclass `repmat` and overwrite `sqw` reference with `sqw`
    segment on each `PageOp` and distribute both as parallel classes

  Both of these lead to confusing entry-point for `parallel_call`

abuts:
Should resolve #1374 It contains the presentation on PageOp design and I believe, the answeres to concerns, rizen within the tickets

I believe, it addresses the concerns but I can of course address these questions in any other form or place


abuts:
The ticket is closed so I suspect this one can be merged without further discussion

PR #1380, opened 2023-11-10
For faccess V4 pass detector arrays through detpar

cmarooney-stfc:
There is no separate data block in faccess V4 for experiment_info/detector_arrays corresponding to those for instruments and samples. Without adding to the data blocks, we must therefore pass this information via the detpar block, and repopulate detpar after it has been read.

Note that detpar will not be required once the detectors are properly populated from nxspes, but in the interim is still used duplicating the first detector bank of the first array

CM - 1.3.24 - this PR was discontinued as it changed the file format and this was not approved. see PR #1540 for succesful alternative.


cmarooney-stfc:
closing without merge - see my comment 1.3.24 in main description comment above

PR #1391, opened 2023-11-17
userpath provided to workers

### LINKED ISSUE #1390: Matlabpath when switching Horace versions
It seems, recent additions for parallel claster initialization, which adds current path to the path of workers, cause problem when one changes Horace version, as if you run parallel cluster with one version and changed it, the path set to Matlab path remains and picks up the code from previous version. 

This should be fixed

abuts:
Fixes #1390.  Number of issues have been identified on iDAAAS related to path-es stored/not stored in MATLABPATH and other enviromental variables. Normally it is invisible but becomes an issue when one dynamically switches from one Horace version to another. Hopefully all fixed now -- switching with parallel cluster works

### DIFF
--- a/herbert_core/classes/MPIFramework/@ClusterWrapper/ClusterWrapper.m
+++ b/herbert_core/classes/MPIFramework/@ClusterWrapper/ClusterWrapper.m
@@ -222,10 +222,39 @@ classdef ClusterWrapper
                 obj.common_env_var_('HERBERT_PARALLEL_EXECUTOR') = obj.matlab_starter_;
             end
 
+            % Modify MATLABPATH to add path to user code, which user 
+            % may want to use in parallel script
+            obj = obj.add_user_path(pc);
+
+            if obj.DEBUG_REMOTE
+                obj.common_env_var_('DO_PARALLEL_MATLAB_LOGGING') = 'true';
+            else
+                obj.common_env_var_('DO_PARALLEL_MATLAB_LOGGING') = 'false';
+            end
+
+        end
+        %
+        function obj = add_user_path(obj,pc)
+            % Method adds user script path to the pathes available to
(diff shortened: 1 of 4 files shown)


herbert_core/classes/MPIFramework/@ClusterWrapper/ClusterWrapper.m
```diff
@@ -222,38 +228,83 @@
...
+            % parallel workers.
+            %
+            % The scripts should not be within Horace
+            % root folder as there files should already be on path or used
+            % in tests only.
+            %
+            % WARNING!!!
+            % It assumed that user view on the filesystem is the same as
+            % the view from parallel nodes. This probably should be
+            % changed in a future.
```
cmarooney-stfc, line 253: Please ticket this and in the ticket provide an explanation why the change should happen

herbert_core/classes/MPIFramework/@ClusterWrapper/ClusterWrapper.m
```diff
@@ -48,6 +48,12 @@
         % cluster. Property used in debugging to change default framework
         pool_exchange_frmwk_name
     end
+    properties(Dependent,Hidden)
+        % return enviromental variables used by cluster. Normally for testing
+        % It is handle, so it can be used for seting it too.
+        % (Violates class logic so avoid this)
```
cmarooney-stfc, line 54: needs more explanation what class logic violation means here, and what steps should be taken to avoid using it - e.g. if you have an alternative safer means of setting this variable, say what it is.

Also say what class this is, so we can see the handle construction logic.

cmarooney-stfc (commented):
There are a couple of places where improved commenting is desirable.

PR #1406, opened 2023-11-28
Add variable width spaghetti plot

### LINKED ISSUE #807: Feature request: Option to vary qwidth in spaghetti_plot
In `spaghetti_plot` at the moment the integration range over the perpendicular directions for each q-line segment is fixed to a single value in reciprocal angstrom (default is `0.1`). This ensures that the slices are all of uniform width but is a bit restrictive for users.

So, some users have requested the ability to specify different widths for each segment. The use case for this would be to increase or decrease the widths in directions where the excitations are (relatively) dispersionless, for example in low-dimensional magnets, in order to improve the (effective) resolution [with smaller widths] or the data quality [larger widths].

oerc0122:
- Add variable widths for q-segments to spaghetti plot
- Make noplot live up to its name (even in case of no action)
- Correct invalid_arguments to invalid_argument

Fixes #807

### DIFF
--- a/horace_core/spaghetti_plot.m
+++ b/horace_core/spaghetti_plot.m
@@ -4,20 +4,22 @@ function varargout = spaghetti_plot(varargin)
 %   >> spaghetti_plot(rlp,data_source)
 %   >> spaghetti_plot(wdisp)
 %
-%   >> spaghetti_plot(...,'labels',{'G','X',...})  % customised labels
-%   >> spaghetti_plot(...,'smooth',1)              % smooth data with this width
-%   >> spaghetti_plot(...,'smooth_shape','hat')    % smooth data with this shape
-%   >> spaghetti_plot(...,'qbin',qb)               % specify q bin size in 1/Ang
-%   >> spaghetti_plot(...,'qwidth',qi)             % specify q integration width in 1/Ang
-%   >> spaghetti_plot(...,'ebin',[elo estp ehi])   % specify energy bin in meV
-%   >> spaghetti_plot(...,'logscale')              % plots intensity in log10 scale
-%   >> spaghetti_plot(...,'clim',[cmin cmax])      % sets the colorscale (needed for logscale)
+%   >> spaghetti_plot(...,'labels',{'G','X',...})            % customised labels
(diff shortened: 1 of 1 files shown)


horace_core/spaghetti_plot.m
```diff
@@ -362,7 +379,7 @@
 
 % Plot dispersion
 %----------------
-if nargout<1 || (~present.noplot)
+if nargout<1
     plot_dispersion(arrayfun(@d2d, wdisp),opt)
 end
```
abuts, line 384: If I correctly understand the change, I would disagree with this. 
You can request plot and ask to return objects -- pretty reasonable situation. Here you allow one or anoter

horace_core/spaghetti_plot.m
```diff
@@ -139,7 +144,7 @@
...
 
 if numel(args) ~= 2
-    error('SPAGHETTI_PLOT:invalid_arguments', ...
+    error('SPAGHETTI_PLOT:invalid_argument', ...
           'Invalid number of arguments')
 end
```
abuts, line 149: I believe, we have agreement to throw 3-component exceptions. Standard exception would be useful if you, e.g. write GUI and catch all normal exceptions to display them in different place in GUI. 

I do not mind if it is not the case here, but having standard 3-component exception 'HORACE:spaghetti_plot:invalid_argument' would simplify further development

abuts (changes requested):

horace_core/spaghetti_plot.m
oerc0122: Agreed, fixed.

horace_core/spaghetti_plot.m
oerc0122: The change in this PR does not affect that at all, this is the original behaviour. 

The idea, I believe, is that if you return the object you clearly have an intent to transform it or use it elsewhere and so don't need to plot it immediately. 

This is indeed an overloaded use of the function and I am not a fan of that in general, but I recommend keeping the behaviour in this PR and if you want to discuss changing it, I would say open an issue.

abuts (approved):

cmarooney-stfc (approved):

Issue #1348, opened 2023-10-20
current implementation of paging operations in tobyfit_DGdisk_resconv/tobyfit_DGfermi_resconv does not make sence


abuts:
The code loads all pixels in memory, calculates deviations and then starts assigning the result to filebacked pages. 
What the point -- if you can load all pixels in memory, you should not bother with filebacked operations. 

Filebacked there either should be removed, or split into pages on much higher level, loading pages and calculating deviations for pages (if it is possible, pixels may go our of current bin with this approach)

abuts:
Superseeded bu #760 and #1403

PR #1420, opened 2023-12-01
Reduce `get_new_handle` usage and disable `shift_pixels`

### LINKED ISSUE #1403: remove internal write handle from sqw and pixels
The majority of filebacked operations have been rewritten using `PageOp`-type algorithms. 

This approach does not use internal write handle previously present in `sqw` objects and related to this handle `get_new_handle` method. 

The ticket is to remove these methods as they should not be used any more. 

This may lead to new tickets to rewrite algorithms using filebacked operations if some pieces of code still use `get_new_hanlde`. These pieces of code should be rewritten using `PageOp`

CM:28.11.23
End of filebacked assessment of technical debt - priority within filebacked
see epic #1302 as noted below

abuts:
This is small part of the ticket #1403 which reduces `get_new_handle` usage in the code. After this change `get_new_handle` will remain in `accumulate_cut_` only, which needs special consideration. 

The file-backed operations are removed from `sort_pix` because they should not be there. `sort_pix` (and `sort_pix_by_bins`) takes the pixels and sorts them according to the image bins. If pixels and image do not fit memory, cut takes parts of the image and calculates corresponding pixels (in memory). Thus, `sort_pix` never works with file-backed pixels and it should not.

Similarly this is done in removing `get_new_handle` from Tobyfit. Where all data are loaded in memory, operation is performed in memory and pixels are dumped to a separate file on disk. This does not make sense and is described further in ticket #1348.

If file-backed pixels are provided to Tobyfit following this PR, an error will be raised.

This PR also disables `shift_pixels` which has not been implemented for new-style `sqw` objects.

Evidence for the need to remove Tobyfit file-backing is provided by #1424

Fixes #1403


horace_core/sqw/@sqw/tobyfit_DGdisk_resconv.m
oerc0122, line 1: Why are you changing resconv here? It appears you're reducing functionality in `master` and introducing potential conflicts in where work is happening for no gain.

horace_core/utilities/sort_pix.m
oerc0122, line 1: Again, this seems to be simply removing functionality

oerc0122 (changes requested):

horace_core/sqw/@sqw/tobyfit_DGdisk_resconv.m
abuts: This is the part of removal of internal write handle in favour of external write handle which would control the way filebacked object is contstructed and written.  The structure of filebacked object produced here is different from the structure external handle is writing. (And the later will be changing). This is why I think this code should be removed now. 

But it is your ticket(branch) anyway

horace_core/utilities/sort_pix.m
abuts: This functionality does not make sence there. Again, you have array of indexes, and if the indexes fit memory, pixels will probably also fit memory. Why to go filebacked? -- memory much faster.

PR #1425, opened 2023-12-07
Rename 'mask' var to 'keep' in `mask_data_for_fit` and `mask_points_for_fit`

oerc0122:
@abuts 
Changes added to #1424 split out into own PR. 

Unsure of need.

### DIFF
--- a/herbert_core/applications/multifit/@mfclass/private/mask_data_for_fit.m
+++ b/herbert_core/applications/multifit/@mfclass/private/mask_data_for_fit.m
@@ -1,4 +1,4 @@
-function [wmask,msk_out] = mask_data_for_fit (w,msk_in)
+function [wmask,keep_out] = mask_data_for_fit (w,msk_in)
 % Mask data
 %
 %   >> [msk_out,ok,mess] = mask_data_for_fit (w,msk_in,xkeep,xremove,msk)
@@ -16,9 +16,9 @@ function [wmask,msk_out] = mask_data_for_fit (w,msk_in)
 %   wmask       Cell array (row) of masked datasets.
 %               If there was an error, wmask is an empty cell array size (1,0)
 %
-%   msk_out     Cell array (row) of mask arrays, one per data set.
+%   keep_out    Cell array (row) of mask arrays, one per data set.
 %               Same size as input data.
-%               If there was an error, msk_out is an empty cell array size (1,0)
+%               If there was an error, keep_out is an empty cell array size (1,0)
 %
 % Objects need a method sigvar_getx or mask_points. See elsewhere for required syntax.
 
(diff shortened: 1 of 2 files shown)


herbert_core/applications/multifit/@mfclass/private/mask_data_for_fit.m
```diff
@@ -16,9 +16,9 @@
 %   wmask       Cell array (row) of masked datasets.
 %               If there was an error, wmask is an empty cell array size (1,0)
 %
-%   msk_out     Cell array (row) of mask arrays, one per data set.
+%   keep_out    Cell array (row) of mask arrays, one per data set.
 %               Same size as input data.
-%               If there was an error, msk_out is an empty cell array size (1,0)
+%               If there was an error, keep_out is an empty cell array size (1,0)
```
cmarooney-stfc, line 21: It is not at all obvious why keep_out is a better name than mask out. How does keep/msk_out differ from msk_in? Does size here refer to numel(w) only or to sizes of pixel arrays?

herbert_core/applications/multifit/@mfclass/private/mask_points_for_fit_xye.m
```diff
@@ -25,7 +25,7 @@
 %           If give variances, this has the same effect (except that the function cannot check
 %           negative error bars; it does check negative variances of course)
 %
-%   mask    [Optional] array of ones and zeros, with the same number of elements as the data
+%   keep    [Optional] array of ones and zeros, with the same number of elements as the data
```
cmarooney-stfc, line 28: again, why name it keep? what does it do?

cmarooney-stfc (changes requested):
Review comments should be met possibly by changing the `keep` names, probably by adding more description to the argument comment lines

herbert_core/applications/multifit/@mfclass/private/mask_points_for_fit_xye.m
abuts: mask presumed removal of marked elements, and here we keep elements which are marked as true. 
This is why renaming occurs. It extreamly difficult to understand code, when you expect one action and it does opposite.

abuts: at least looking at inconsistency between the method name (`mask`) and the argument name `keep` you think what is actually happening.

oerc0122: Personally, I disagree, I find the concept of a positive or a negative mask equally valid. However, I think the name `keep` and certainly `keep_out` are worse names than mask. If anything `to_keep` for both would be better, however, I would argue for leaving as-is.

PR #1427, opened 2023-12-08
join using page_op in memory

abuts:
This is part of ticket #1402 which implements join using PageOp in case when all input and output data are located in memory.

PR #1428, opened 2023-12-08
join filebacked objects

abuts:
This is second step in developing PageOp-based `join` which allows source `sqw` objects and target `sqw` object be filebacked.

Should be considered after PR #1427 is merged as it is small addition to that PR.


oerc0122:
If this is a modification into #1247 you can set the base branch to merge to to be that branch (`1404_join_page_op`) rather than `master`, at which point it will only display changes relative to `1404_join_page_op` and be easier to review meaning less time in the review queue. 

If #1247 is merged first, GitHub will automatically update this to merge to `master`.

abuts:
@oerc0122  -- If I do as you say, I would not be able to automatically test the changes as pipeline is set up to test agains `master`. And the puropse of incremental changes is to keep working code ready to be merged to master.

Issue #901, opened 2022-11-15
dirty hack performance improvement in tobyfit_DGdisk_resconv/tobyfit_DGfermi_resconv


abuts:
These two functions by default convert  run_id into header index using `runid_map` which is a Matlab `containers.Map` object. 
For some reason on some machines this is extremely slow operation. 

To avoid this, a piece of the code has been inserted into these functions (tobyfit_DFdisk_resconv --   rows 119-214, tobyfit_DGfermi_resconv rows 198-209 which converts maps to arrays and uses these arrays to modify indexes.  The operation provides a order of magnitude improvement in the performance. 

This is very hack-y solution, as it relies on some assumptions, uses duplicated non-unit tested code  and should be moved somewhere within the lookup tables.

abuts:
fixed by @tgperring ticket #1419 PR #1431.

PR #1431, opened 2023-12-10
1419 tobyfit easy optimization

tgperring:
Profiling tobyfit tests highlighted a few easy optimisations for DGfermi (and similarly for DGdisk)
- Moving the test for mosaic spread of the sample from tobyfit_DGfermi_resconv (where it is called for every partial differential wrt a fitting parameter) to tobyfit_DGfermi_resconv_init and holding the result in the pre-calculated lookup lookup structure. For some reason this was a very expensive operation - reason unknown - but needlessly recalculated every time. 
- Solving issue #901 where the mapping of the run index in the pix array to the indices into the experiment_info in the sqw object was using an inefficient algorithm in parse_pixel_indices, there was a hack that was fast but not robust. There were also some bugs associated with the mapping not being done in all the places it should. This is now solved by adding a new method to the class Experiment that performs the mapping efficiently and which is used by parse_pixel_indices. One of the input argument cases to parse_pixel_indices has been removed (it was unnecessary, unused and confusing). 
- A couple of small optimisations to the serializable class have been made
-in a non-systematic comparison of a few tobyfit examples these changes have resulted in the fit time being reduced by 30-50% when tested on Toby Perring's laptop.


abuts (commented):
Generally looks good. Should be minor changes to comments. 

Small question is if `parse_pixel_indices` in `tobyfit_DG...` should be taken out off loop over sqw objects as it seems it works for array of objects. But I believe, profiling has been done to ensure it works reasonable.

horace_core/Tobyfit/tobyfit_DGdisk_resconv_init.m
```diff
@@ -5,7 +5,7 @@
 %   >> [ok,mess,lookup]=tobyfit_DGdisk_resconv_init(win)
 %
 % For specific pixels:
-%   >> [ok,mess,lookup]=tobyfit_DGdisk_resconv_init(win,indx)
+%   >> [ok,mess,lookup]=tobyfit_DGdisk_resconv_init(win,ipix)
```
oerc0122, line 8: Any strong reason for the rename? To me, intuitively, due to Fortran, `ipix` is a counter of number pixels akin to `npix` rather than pixel indices. `pix_ind` may be better if it's being renamed.

horace_core/Tobyfit/tobyfit_DGfermi_resconv_init.m
```diff
@@ -24,29 +24,21 @@
...
-%                       {{irun1,idet1,ien1}, {irun2,idet2,ien2}}
+%               - Cell array of pixel indices arrays
+%                   - only one array: applied to every sqw object, or
+%                   - several pixel indices arrays: one per sqw object
 %
 %   opt         Option: 'tables' (default) or 'notables'
-%               If 'tables', then moderator and Fermi chopper lookup tables
-%              are added to the output argument lookup
-%
+%               If 'tables', then object_lookup tables for instrumnet
```
oerc0122, line 39:
```suggestion
%               If 'tables', then object_lookup tables for instrument
```

horace_core/sqw/@Experiment/private/get_experiment_idx_.m
```diff
@@ -0,0 +1,85 @@
...
+% A full or sparse lookup array can be forced by specifying the optional
+% argument sparse_lookup as false or true.
+if ~exist('sparse_lookup','var')
+    range_key_threshold = 10000;
+    frac_sparse_threshold = 0.5;
+    if (range_keys < range_key_threshold) || (numel(keys) > frac_sparse_threshold * range_keys)
+        make_sparse_lookup = false;
+    else
+        make_sparse_lookup = true;
+    end
```
oerc0122, line 51: Could do:
```suggestion
    make_sparse_lookup =  (range_keys > range_key_threshold) && ...  % Enough keys to be worthwhile
         (numel(keys) < frac_sparse_threshold * range_keys)          % Total range < frac*100% full
```

horace_core/sqw/@Experiment/private/get_experiment_idx_.m
```diff
@@ -0,0 +1,85 @@
...
+else
+    make_sparse_lookup = logical(sparse_lookup);    % handle case of numeric input
+end
+
+if make_sparse_lookup
+    % Use a sparse array to hold the lookup array if the range of the keys is
+    % greater than a threshold and the keys are sparse than a sparcity
```
oerc0122, line 58:
```suggestion
    % greater than a threshold and the keys are more sparse than a sparcity
```

horace_core/sqw/@Experiment/private/get_experiment_idx_.m
```diff
@@ -0,0 +1,85 @@
...
+end
+
+try
+    expt_idx = lookup(run_idx - min_key + 1);
+    if issparse(expt_idx)
+        expt_idx = full(expt_idx);  % convert to full array in case of sparse lookup
+    end
+    expt_idx = reshape(expt_idx, size(run_idx));
+catch
+    error('HORACE:Experiment:invalid_argument',...
```
oerc0122, line 79: Wouldn't this be better to detect before we've built the array?

cmarooney-stfc (changes requested):

horace_core/sqw/@Experiment/private/get_experiment_idx_.m
oerc0122: Also try to keep `try`-`catch` as tight as possible in case of unexpected or API changes in future. 

In general, it's also best not to have a "bare" `try`-`catch`, `catch` an exception and check its type. You're already crashing and paying the cost of a `try-catch`, so speed is never key here, what's important is not masking unexpected errors.

horace_core/sqw/@sqw/parse_pixel_indices.m
```diff
@@ -1,212 +1,206 @@
...
-%              up indx. If indx has only one entry, then it is assumed to
-%              apply for any value of iw.
+%               - Cell array of pixel indices arrays
+%                   - only one array: applied to every sqw object, or
+%                   - several pixel indices arrays: one per sqw object
 %
 % Output:
 % -------
-%   irun        Single sqw object: Indexes of the experiments, containing
+%   irun        Single sqw object: Array of indices into the experiment_info
```
oerc0122, line 35: "single" or "scalar"? May be best to use "scalar" since `single` is a type and `sqw` objects *can* contain `single` data.

horace_core/sqw/@sqw/parse_pixel_indices.m
```diff
@@ -1,212 +1,206 @@
...
+%               Multiple sqw objects: Cell array of arrays, one per sqw object
+%                                  Cell array has the same size as win
```
oerc0122, line 37: May be worth saying "shape" rather than "size", even though in MATLAB they're conflated.

horace_core/sqw/@sqw/parse_pixel_indices.m
```diff
@@ -1,212 +1,206 @@
...
+        n_ipix = numel(ipix);
+        if n_ipix==0 || any(cellfun(@isempty, ipix))
+            error('HORACE:parse_pixel_indices:invalid_argument', ['The cell array ',...
+                '''ipix'' is empty, or one or more arrays in ''ipix'' is empty'])
+        end
+        if ~any(numel(ipix) == [1,nw])
+            error('HORACE:parse_pixel_indices:invalid_argument', ['The number of ',...
+                'arrays in ''ipix'' must be 1 or the number of sqw objects: %d'], nw)
+        end
+        if numel(ipix)==1
```
oerc0122, line 84:
```suggestion
        if isscalar(ipix)
```

horace_core/sqw/@sqw/parse_pixel_indices.m
```diff
@@ -1,212 +1,206 @@
...
         if nw == 1
-            irun = irun{1};
-            idet = idet{1};
-            ien = ien{1};
+            min_ipix = min(ipix(:));
+            max_ipix = max(ipix(:));
+            ipix_cellarray = {ipix};
+        else
+            min_ipix = min(ipix(:)) * ones(size(win));
+            max_ipix = max(ipix(:)) * ones(size(win));
```
oerc0122, line 105:
```suggestion
            min_ipix = repmat(min(ipix(:)), size(win));
            max_ipix = repmat(max(ipix(:)), size(win));
```

horace_core/sqw/@sqw/parse_pixel_indices.m
```diff
@@ -1,212 +1,206 @@
...
-    else
-        if nind ~= nw
-            error('HORACE:parse_pixel_indices:invalid_argument', 'If there is more than one indexing array, the number must match the number of sqw objects');
-        end
-        iw_internal = 1:nw;  % one indx array per sqw object
-    end
+% If no return arguments, simply have a silent return.
+% This goes against the default Matlab behaviour for functions, which is to
+% return the first output argument in a variable called 'ans'. However, that
+% could be fairly CPUtime expensive here.
```
oerc0122, line 125: Is that the expectation of functions? That is just the role of the interpreter to handle cases where arguments are returned. I think this comment can be removed provided it is documented (and it is) in the docstrings.

horace_core/sqw/@sqw/parse_pixel_indices.m
```diff
@@ -1,212 +1,206 @@
...
-% Check consistency
-for i=1:nw
-    if iscell(win)
-        wtmp = win{i};
+% Get the return arguments
+% We have repackaged ipix so that it is a cell array with size equal to the size
+% of win. In the following we only have to branch on (1) the presence or absence
+% of ipix, (2) win being an object array or a cell array of sqw objects
+if ipix_present
+    if nw == 1
```
oerc0122, line 135: I would probably invert the order of these `if`s to have related calls closer. In my view, the core test is whether we have a scalar or array for our `sqw` and the `ipix_present` just determines the exact implementation.

horace_core/sqw/@sqw/parse_pixel_indices.m
```diff
@@ -1,212 +1,206 @@
...
-        irun = irun{1};
-        idet = idet{1};
-        ien = ien{1};
-    end
+else
+    % Catch the case of no pixels for simplicity
+    irun = zeros(0,1);  % convention is column vector
+    idet = zeros(0,1);
+    ien = zeros(0,1);
 end
```
oerc0122, line 204: Depending on your thoughts on single returns from functions (single point of exit at the bottom) I would probably move this up as an early `return` to flatten this function somewhat.
```
if pix.num_pixels == 0  % Catch the case of no pixels for simplicity
    irun = zeros(0,1);  % convention is column vector
    idet = zeros(0,1);
    ien = zeros(0,1);
    return;
end
```
Or flatten the conditions as:
```
if pix.num_pixels == 0  % Catch the case of no pixels for simplicity
   ...
elseif exist('ipix', 'var')
   ...
else
   ...
end
```

horace_core/sqw/@sqw/tobyfit_DGdisk_resfun_covariance.m
```diff
@@ -78,17 +64,25 @@
...
+    % Get the indices to the runs in the experiment information block, the 
+    % detector indicies and the energy bin indices
     if all_pixels
-        [irun,idet] = parse_pixel_indices (wtmp);
+        % For all pixels in the sqw object
+        [irun,idet] = parse_pixel_indices (win(iw));
     else
-        [irun,idet] = parse_pixel_indices (wtmp,indx,iw);
+        % For the selected pixels only
+        if iscell(ipix) && numel(ipix)>1
```
oerc0122, line 79: I would merge the `else`-`if` for flow

horace_core/sqw/@sqw/tobyfit_DGdisk_resfun_covariance.m
```diff
@@ -78,17 +64,25 @@
...
-    else
-        wtmp = win(iw);
-    end
+cov_proj = cell(size(win));
+cov_spec = cell(size(win));
+cov_hkle = cell(size(win));
 
+for iw = 1:numel(win)
+    % Get the indices to the runs in the experiment information block, the 
+    % detector indicies and the energy bin indices
```
oerc0122, line 73:
```suggestion
    % detector indices and the energy bin indices
```

horace_core/sqw/@sqw/tobyfit_DGdisk_resfun_covariance.m
```diff
@@ -126,20 +120,21 @@
...
-    cov_proj{iw} = transform_matrix (cov_hkle{iw}, inv(wtmp.data.u_to_rlu));
+    cov_proj{iw} = transform_matrix (cov_hkle{iw}, inv(win(iw).data.u_to_rlu));
     rlu_to_spec = invert_matrix (spec_to_rlu);
     rlu_to_spec4(1:3,1:3,:) = rlu_to_spec(:,:,irun);
     rlu_to_spec4(4,4,:) = 1;
     cov_spec{iw} = transform_matrix (cov_hkle{iw}, rlu_to_spec4);
 end
 
-if ~iscell(win) && isscalar(win)
+if isscalar(win)
```
oerc0122, line 136: Can `win` no longer be a base `sqw` array?

horace_core/sqw/@sqw/tobyfit_DGfermi_resfun_covariance.m
```diff
@@ -78,17 +64,25 @@
...
-    
+    % Get the indices to the runs in the experiment information block, the 
+    % detector indicies and the energy bin indices
     if all_pixels
-        [irun,idet] = parse_pixel_indices (wtmp);
+        % For all pixels in the sqw object
+        [irun,idet] = parse_pixel_indices (win(iw));
     else
-        [irun,idet] = parse_pixel_indices (wtmp,indx,iw);
+        % For the selected pixels only
```
oerc0122, line 78: As above

horace_core/sqw/@sqw/tobyfit_DGfermi_resfun_covariance.m
```diff
@@ -129,20 +123,21 @@
...
+    cov_proj{iw} = transform_matrix (cov_hkle{iw}, inv(win(iw).data.u_to_rlu));
     rlu_to_spec = invert_matrix (spec_to_rlu);
     rlu_to_spec4(1:3,1:3,:) = rlu_to_spec(:,:,irun);
     rlu_to_spec4(4,4,:) = 1;
     cov_spec{iw} = transform_matrix (cov_hkle{iw}, rlu_to_spec4);
 end
 
-if ~iscell(win) && isscalar(win)
+if isscalar(win)
+    % Numeric array output if just one sqw object
```
oerc0122, line 140: As above

horace_core/sqw/parse_pixel_indices.m
```diff
@@ -0,0 +1,76 @@
...
+%   >> [irun, idet, ien] = parse_pixel_indices (win, ipix)  % selected pixels
+%
+% Note: an efficient way of checking that ipix is consistent with all the sqw
+% objects is to call without output arguments i.e.
+%
+%   >> parse_pixel_indices (win, ipix)
+%
+% This will return silently without without spending time computing output
+% argument irun if the arguments are consistent, but will throw an error with an
+% appropriate message if e.g. ipix is out of range for
```
oerc0122, line 14: "for" what?

horace_core/sqw/parse_pixel_indices.m
```diff
@@ -0,0 +1,76 @@
...
+%               Multiple sqw objects: Cell array of arrays, one per sqw object
+%                                  Cell array has the same size as win
+%
+% For each of irun, idet and ien, note that:
+% - If ipix was not given, the output arrays for each sqw object are column vectors
+% - If ipix was given, the output arrays for each sqw object have the same shape
+%   as the corresponding array indices array within ipix
+
+
+% Ensure sqw objects form a cell array of scalar sqw object(s)
```
oerc0122, line 56:
```suggestion
% Ensure win is a cell array of scalar sqw object(s)
```

horace_core/sqw/@Experiment/private/get_experiment_idx_.m
```diff
@@ -0,0 +1,85 @@
+function expt_idx = get_experiment_idx_ (obj, run_idx, sparse_lookup)
+% Get the experiment indices for an array of run indices in the pixel data
+%
```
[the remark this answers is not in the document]
tgperring: Expanded description now added to the opening comments in the function.

horace_core/sqw/parse_pixel_indices.m
```diff
@@ -0,0 +1,76 @@
...
+%   >> [irun, idet, ien] = parse_pixel_indices (win)        % all pixels
+%   >> [irun, idet, ien] = parse_pixel_indices (win, ipix)  % selected pixels
+%
+% Note: an efficient way of checking that ipix is consistent with all the sqw
+% objects is to call without output arguments i.e.
+%
+%   >> parse_pixel_indices (win, ipix)
+%
+% This will return silently without without spending time computing output
+% argument irun if the arguments are consistent, but will throw an error with an
```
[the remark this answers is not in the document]
tgperring: Done

horace_core/sqw/parse_pixel_indices.m
```diff
@@ -0,0 +1,76 @@
...
+%
+%
+% Input:
+% ------
+% Checking validity:
+%   win         Array of sqw objects, or cell array of scalar sqw objects
+%
+% [Optional]
+%   ipix        Pixel indices for which the output is to be extracted from the
+%               sqw object(s)
```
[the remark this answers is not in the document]
tgperring: Mde the change to '%               'sqw object(s). It has the form of one of:'
This appeared in six function, now all changed.

_test/test_sqw/test_parse_pixel_indices_cell_array_input.m
```diff
@@ -0,0 +1,136 @@
...
+            assertEqual(ien{2}, obj.ien2)
+        end       
+        
+        function test_columnArraySqwObject(obj)
+            % Row vector sqw object
+            w = {obj.w1; obj.w2};
+            [irun,idet,ien]=parse_pixel_indices(w);
+            assertTrue(isa(irun,'cell')&isequal(size(irun),[2,1]))
+            assertTrue(isa(idet,'cell')&isequal(size(idet),[2,1]))
+            assertTrue(isa(ien,'cell')&isequal(size(ien),[2,1]))
```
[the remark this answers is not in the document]
tgperring: DOne. I dont know why I left it in the less readable and single '&' form that I did.

_test/test_sqw/test_parse_pixel_indices_cell_array_input.m
```diff
@@ -0,0 +1,136 @@
...
+            assertEqual(ien, ien_ref)
+        end
+        
+        function test_rowArraySqwObject(obj)
+            % Row vector sqw object
+            w = {obj.w1, obj.w2};
+            [irun,idet,ien]=parse_pixel_indices(w);
+            assertTrue(isa(irun,'cell')&isequal(size(irun),[1,2]))
+            assertTrue(isa(idet,'cell')&isequal(size(idet),[1,2]))
+            assertTrue(isa(ien,'cell')&isequal(size(ien),[1,2]))
```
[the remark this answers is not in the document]
tgperring: THere was a cut-n-past. See above. Done

_test/test_experiment/test_experiment_methods.m
```diff
@@ -312,6 +312,85 @@ function test_get_subobj_by_ind(obj)
...
+            experiment.runid_map = containers.Map([20,30,40],[1,3,2]);
+            
+            % Array of valid run_idx
+            run_idx = [30,40,20,20,30];
+            expt_idx = experiment.get_experiment_idx (run_idx);
+            
+            assertEqual(expt_idx, [3,2,1,1,3])
+        end
+        
+        function test_get_experiment_idx_fullLookup_arrayRun_idx(obj)
```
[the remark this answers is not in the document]
tgperring: Yes. In an oversight I forgot to edit the 'true' to 'false'. Fortunately, the tests pass with the change to 'false' that I've now made.

_test/test_sqw/test_parse_pixel_indices_cell_array_input.m
```diff
@@ -0,0 +1,136 @@
...
+            irun_ref = {reshape(obj.irun1(ipix), size(ipix))};
+            idet_ref = {reshape(obj.idet1(ipix), size(ipix))};
+            ien_ref = {reshape(obj.ien1(ipix), size(ipix))};
+            assertEqual(irun, irun_ref)
+            assertEqual(idet, idet_ref)
+            assertEqual(ien, ien_ref)
+        end
+        
+        function test_rowArraySqwObject(obj)
+            % Row vector sqw object
```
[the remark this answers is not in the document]
tgperring: It's a separate file because there is a parse_pixel_indices that is a method of sqw (that takes an array of sqw objects as input), and a function with the same name that takes a cell array of scalar sqw objects. Not the nicest way of doing things, but this is a functionality (just like the function tobyfit_DGfermi_resconv) that is wanted for parse_pixel_indices, namely that it can take an sqw object array or a cell array. The separate file for the tests is because there are separate test folders for sqw methods and for functions in the sqw folder (not @sqw).

_test/test_sqw/test_parse_pixel_indices_cell_array_input.m
```diff
@@ -0,0 +1,136 @@
...
+            % Scalar sqw object, but without any pixels: is valid
+            w = {sqw()};
+            [irun,idet,ien]=parse_pixel_indices(w);
+            assertEqual(irun, {zeros(0,1)})
+            assertEqual(idet, {zeros(0,1)})
+            assertEqual(ien, {zeros(0,1)})
+        end
+        
+        function test_scalarSqwObject(obj)
+            % Scalar sqw object
```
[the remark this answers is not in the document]
tgperring: Indeed. Lazily I didn't update the leading comment lines on a few of the tests in this classdef. Now updated.

_test/test_sqw/test_parse_pixel_indices_cell_array_input.m
```diff
@@ -0,0 +1,136 @@
...
+        function test_columnArraySqwObject_ipixMultiple(obj)
+            % Row vector sqw object
+            w = {obj.w1; obj.w2};
+            ipix1 = [5,21,3;15,5,7];
+            ipix2 = [5,13,3;15,5,7];
+            ipix = {ipix1; ipix2};
+            [irun,idet,ien]=parse_pixel_indices(w, ipix);
+            assertTrue(isa(irun,'cell')&isequal(size(irun),[2,1]))
+            assertTrue(isa(idet,'cell')&isequal(size(idet),[2,1]))
+            assertTrue(isa(ien,'cell')&isequal(size(ien),[2,1]))
```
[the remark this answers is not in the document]
tgperring: Changed here (and a number of other locations where the same is done).

_test/test_sqw_class/test_parse_pixel_indices.m
```diff
@@ -0,0 +1,167 @@
...
+        function test_scalarSqwObject(obj)
+            % Scalar sqw object
+            w = obj.w1;
+            [irun,idet,ien]=parse_pixel_indices(w);
+            assertEqual(irun, obj.irun1)
+            assertEqual(idet, obj.idet1)
+            assertEqual(ien, obj.ien1)
+        end
+        
+        function test_scalarSqwObject_ipixEmpty_ERROR(obj)
```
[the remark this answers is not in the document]
tgperring: This test is missing from the cell_array_input tests, because the only tests in cell_array_input concern the correct handling of cell array to object array conversion (including edge cases of empty cell array, for example), and the passing of ipix as a varargin{:} in the internals of the parse_pixel_indices for cell arrays, which just calls the sqw method of the same name. Once the correct passing of the arguments has been tested, all the other tests about size of ipix in relation to the sqw object array in test_parse_pixel_indices apply. The explanation for two files was given earlier.

_test/test_sqw_class/test_parse_pixel_indices.m
```diff
@@ -0,0 +1,167 @@
...
+        end
+        
+        function test_rowArraySqwObject_ipixSingle(obj)
+            % Row vector sqw object
+            w = [obj.w1, obj.w2];
+            ipix = [5,13,3;15,5,7];
+            [irun,idet,ien]=parse_pixel_indices(w, ipix);
+            assertTrue(isa(irun,'cell')&isequal(size(irun),[1,2]))
+            assertTrue(isa(idet,'cell')&isequal(size(idet),[1,2]))
+            assertTrue(isa(ien,'cell')&isequal(size(ien),[1,2]))
```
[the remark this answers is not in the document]
tgperring: Done

_test/test_sqw_class/test_parse_pixel_indices.m
```diff
@@ -0,0 +1,167 @@
...
+        function test_rowArraySqwObject_ipixMultiple(obj)
+            % Row vector sqw object
+            w = [obj.w1, obj.w2];
+            ipix1 = [5,21,3;15,5,7];
+            ipix2 = [5,13,3;15,5,7];
+            ipix = {ipix1; ipix2};  % column: should be OK as shape wrt w not tested
+            [irun,idet,ien]=parse_pixel_indices(w, ipix);
+            assertTrue(isa(irun,'cell')&isequal(size(irun),[1,2]))
+            assertTrue(isa(idet,'cell')&isequal(size(idet),[1,2]))
+            assertTrue(isa(ien,'cell')&isequal(size(ien),[1,2]))
```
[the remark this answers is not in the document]
tgperring: Done

_test/test_sqw_class/test_parse_pixel_indices.m
```diff
@@ -0,0 +1,167 @@
...
+        function test_columnArraySqwObject_ipixMultiple(obj)
+            % Row vector sqw object
+            w = [obj.w1; obj.w2];
+            ipix1 = [5,21,3;15,5,7];
+            ipix2 = [5,13,3;15,5,7];
+            ipix = {ipix1; ipix2};
+            [irun,idet,ien]=parse_pixel_indices(w, ipix);
+            assertTrue(isa(irun,'cell')&isequal(size(irun),[2,1]))
+            assertTrue(isa(idet,'cell')&isequal(size(idet),[2,1]))
+            assertTrue(isa(ien,'cell')&isequal(size(ien),[2,1]))
```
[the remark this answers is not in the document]
tgperring: Done

herbert_core/utilities/classes/@serializable/private/to_bare_struct_.m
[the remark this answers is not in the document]
tgperring: I agree. We've discussed that now, also with Chris.

horace_core/Tobyfit/tobyfit_DGdisk_resconv_init.m
tgperring: The reason for the rename was to explicitly include 'pix' in the name. The previous name of 'indx' was as an indicator of a general index which covered a scenario of being an Nx3 array of irun-idet-ien, a case which was never used and is in fact unnecessary. It now only is used as array of pixel indices. Perhaps pix_ind is better, but given that I've made the change and to change again would require numerous edits in several iles I'd rather not now, and chalk this up to using apossibly better name for similar entities in the future.

horace_core/Tobyfit/tobyfit_DGfermi_resconv_init.m
tgperring: Done (and also in tobyfit_DGdisk_resconv_init).

horace_core/Tobyfit/tobyfit_DGfermi_resconv_init.m
```diff
@@ -226,11 +223,20 @@
...
-    % Pixel indicies
+    % Get the indices to the runs in the experiment information block, the 
+    % detector indicies and the energy bin indices
     if all_pixels
-        [irun,idet,ien] = parse_pixel_indices(wtmp);
+        % For all pixels in the sqw object
+        [irun, idet, ien] = parse_pixel_indices(wtmp);
     else
-        [irun,idet,ien] = parse_pixel_indices(wtmp,indx,iw);
+        % For the selected pixels only
```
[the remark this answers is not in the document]
tgperring: Hah! I anticipated this above . Done.

horace_core/Tobyfit/tobyfit_DGfermi_resconv_init.m
```diff
@@ -5,7 +5,7 @@
 %   >> [ok,mess,lookup]=tobyfit_DGfermi_resconv_init(win)
 %
 % For specific pixels:
-%   >> [ok,mess,lookup]=tobyfit_DGfermi_resconv_init(win,indx)
+%   >> [ok,mess,lookup]=tobyfit_DGfermi_resconv_init(win,ipix)
```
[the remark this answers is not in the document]
tgperring: Also above.

horace_core/Tobyfit/tobyfit_DGdisk_resconv_init.m
```diff
@@ -222,32 +220,27 @@
...
+    % Get the indices to the runs in the experiment information block, the 
+    % detector indicies and the energy bin indices
     if all_pixels
-        [irun,idet,ien] = parse_pixel_indices(wtmp);
+        % For all pixels in the sqw object
+        [irun, idet, ien] = parse_pixel_indices(wtmp);
     else
-        [irun,idet,ien] = parse_pixel_indices(wtmp,indx,iw);
+        % For the selected pixels only
+        if iscell(ipix) && numel(ipix)>1
```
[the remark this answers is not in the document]
tgperring: I looked at this; personally, I find it easier to think of the problem as two cases:  one when all_pixels is true, and the case when it is false; and that the case when it is false happens to have two sub-cases. Merging the else-if makes it harder for me to see the logic flow. I've changes it nevertheless (and in *DGfermi* and *resfun_covariance, where the same block appears) ; I'll see if it feels comfortable when I revisit those codes, as I am bound to do sometime.

horace_core/sqw/@Experiment/private/get_experiment_idx_.m
tgperring: I like it. Done.

horace_core/sqw/@Experiment/private/get_experiment_idx_.m
tgperring: Done

horace_core/sqw/@Experiment/private/get_experiment_idx_.m
tgperring: Agreed. Code modified.

horace_core/sqw/@sqw/parse_pixel_indices.m
tgperring: Agreed; done. It turns out that there are a large number of cases of 'single sqw' rather than 'scalar sqw' in the Horace code; we can clear then out as we come across them, or perhaps one day a documentation-clarification blitz.

horace_core/sqw/@sqw/parse_pixel_indices.m
tgperring: Yes, the conflation is annoying. But I'll leave as is because of the inherent Matlab conflation (if I'm not concentrating I'll occasionally type 'resize' when I  should be typing 'reshape', and 'shape' for the function called 'size')

horace_core/sqw/@sqw/parse_pixel_indices.m
tgperring: I've always assumed that repmat is a more expensive operation than ones. I went online and there is a surprising amount of debate about using a * ones(size(M)), a + zeros(size(M)), repmat (a, size(M)). WIthout spending more than a couple of minutes on it, it would appear there is much of a muchness regarding timings, although my guess would be repmat has more overheads because it has to check the type of what is being repeated. Personally, I find a * ones(size(M) easiest to understand, so am leaving it as is. Would be interesting for us to have a discussion one day...

horace_core/sqw/@sqw/parse_pixel_indices.m
tgperring: I added it in case someone queried the silent return. Happy to remove; done.

horace_core/sqw/@sqw/parse_pixel_indices.m
tgperring: I inverted the order of the ifs to try it on for size, and agree that it looks better. Done

horace_core/sqw/@sqw/parse_pixel_indices.m
tgperring: Usually I prefer to have a single point of exit at the end, but in this case I think that it improves the readability considerably to catch the trivial case first and returning, as you suggest in option 1. I didn't flatten the if statement as in your second option, as it places two very different tests (no pixels, and the presence of an optional input argument) into one if...end block. I think it better to separate the two tests.

horace_core/sqw/@sqw/tobyfit_DGdisk_resfun_covariance.m
tgperring: Done

horace_core/sqw/@sqw/tobyfit_DGdisk_resfun_covariance.m
tgperring: Done (and in *DGfermi*)

horace_core/sqw/@sqw/tobyfit_DGdisk_resfun_covariance.m
tgperring: Previously the code handled the case of win being a cell array of scalar sqw objects, for the case of an ancient possible use case, but this was never used, and is unnecessary anyway as it is so easy to turn a cell array of objects of the same class into an object array (modulo the discussion we had about heterogeneous arrays). The case of a cell array of sqw objects as input cannot access tobyfit_DGdisk_resfun_covariance anyway, as it is in the class definition folder. I therefore have removed what is redundant check.

horace_core/sqw/@sqw/tobyfit_DGfermi_resfun_covariance.m
tgperring: Done.

horace_core/sqw/@sqw/tobyfit_DGfermi_resfun_covariance.m
tgperring: Comment as above

horace_core/sqw/parse_pixel_indices.m
tgperring: Yes... Chris spotted my hanging edit too. Comment changed.

horace_core/sqw/parse_pixel_indices.m
tgperring: Done

oerc0122 (approved):

PR #1433, opened 2023-12-11
Add documentation on `mask_points` behaviour

### LINKED ISSUE #1409: Document `mask_points` behaviour
Following a request on Horace Help regarding the masking of data the issue was the assumption that the masking algorithm works based on bin boundaries and that if any region of the bin is within the mask region, it should be masked. 

The algorithm, however, masks bins if the bin centres are within the mask region.

This should be documented in the Horace manual.

oerc0122:
Add documentation for `mask_points`' behaviour with respect to bin-centres vs. bin-edges. 

Fixes #1409

### DIFF
--- a/documentation/user_docs/docs/manual/Reshaping_etc.rst
+++ b/documentation/user_docs/docs/manual/Reshaping_etc.rst
@@ -272,7 +272,7 @@ The inputs are:
 .. warning::
 
    It should be noted that masking through the ``xkeep`` and ``xremove``
-   arguments masks data based on the bin-centres and not through any
+   arguments will mask data based on the bin-centres and not through any
    intersection of any bin-edges. This means that for a 1-D case where:
 
    .. code-block:: matlab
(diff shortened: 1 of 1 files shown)


documentation/user_docs/docs/manual/Reshaping_etc.rst
```diff
@@ -269,6 +269,22 @@ The inputs are:
 - ``xremove`` is the range of display axes to remove. Follows the same format as
   ``xkeep``.
 
+.. warning::
+
+   It should be noted that masking through the ``xkeep`` and ``xremove``
+   arguments masks data based on the bin-centres and not through any
```
cmarooney-stfc, line 275:
```suggestion
   arguments will mask data based on the bin-centres and not through any
```
Issue here is that `masks` can be read as a noun, which is confusing. This makes a verb explicit

documentation/user_docs/docs/manual/Reshaping_etc.rst
```diff
@@ -269,6 +269,22 @@ The inputs are:
...
+   intersection of any bin-edges. This means that for a 1-D case where:
+
+   .. code-block:: matlab
+
+      bins = [1 2 3 4] % <- Defines bin-centres at: [1.5, 2.5, 3.5]
+      mask_points(w, 'keep', [1.7, 3.51])
+
+   Will remove the first bin because even though ``1.7`` lies within the first
```
cmarooney-stfc, line 283:
```suggestion
   will remove the first bin because even though ``1.7`` lies within the first
```

cmarooney-stfc (changes requested):

Issue #1439, opened 2023-12-13
combine_equvalent_zones is still used in user documentation


abuts:
`combine_equivalent_zones` routine have been removed from Horace doe base due to inefficiency, but still mentioned in user documentation. Users are confused.

abuts:
Closed through PR#1440

PR #1440, opened 2023-12-13
Removing references to redundant usage of combine_equivalent_zones

abuts:
Re #1439  -- Removed references to redundant function from user documentation.


documentation/user_docs/docs/manual/FAQ.rst
```diff
@@ -41,7 +41,7 @@ The ``sqw`` object carries around an array containing the information of each
 detector-energy pixel which underlies the result in a raw form. These pixels are
 then accumulated and binned into the intensity histogram which makes up the
 ``dnd``. ``sqw`` s also carry around details about the contributing ``.nxspe``
-files from which it was made and, in particular, allows for instrunment
+files from which it was made and, in particular, allows for instrument's
```
oerc0122, line 44: This should not be in this PR

documentation/user_docs/docs/manual/Symmetrising_etc.rst
```diff
@@ -761,149 +761,3 @@ dataset (thereby increasing the counts four-fold).
...
-the cuts (e.g ``[1,1,0]+-1``) will be accounted for twice. The direction of the
-projection should be changed to avoid this.
-
-Limitations
-===========
-
-* At present ``symmetrise_sqw``, ``combine_sqw``, and ``rebin_sqw`` work ONLY
-  for sqw objects, since they require access to individual detector pixel
-  information. The functions will work for any dimensionality of object,
-  however.
```
oerc0122, line 901: Should not be removed

oerc0122 (changes requested):

documentation/user_docs/docs/manual/Symmetrising_etc.rst
```diff
@@ -761,9 +761,20 @@ dataset (thereby increasing the counts four-fold).
...
\ No newline at end of file
+  however.
+
+.. | Re #1447 to enable  * ``combine_equivalent_zones`` has to perform some memory and hdd-access
+  intensive calculations, which should ideally be performed on `high performance
+  computing cluster
+  <http://www.isis.stfc.ac.uk/groups/excitations/data-analysis-computers/connecting-to-isiscomputendrlacuk-using-nomachine15120.html>`__. The
+  amount of memory used by the code is controlled by ``hor_config`` parameter
+  ``mem_chunk_size`` and is approximately 10 times larger then the amount,
+  specified by this parameter. | replace ::
```
oerc0122, line 780:
```suggestion
..
   Re #1447 to enable * ``combine_equivalent_zones`` has to perform some memory and
   hdd-access intensive calculations, which should ideally be performed on `high
   performance computing cluster
   <http://www.isis.stfc.ac.uk/groups/excitations/data-analysis-computers/connecting-to-isiscomputendrlacuk-using-nomachine15120.html>`__. The
   amount of memory used by the code is controlled by ``hor_config`` parameter
   ``mem_chunk_size`` and is approximately 10 times larger then the amount,
   specified by this parameter.
```

documentation/user_docs/docs/manual/Symmetrising_etc.rst
```diff
@@ -761,9 +761,20 @@ dataset (thereby increasing the counts four-fold).
    calculate magnetic form factor on or apply magnetic form factor to the
    dataset provided.
 
+.. Re# 1447 should we reimplement `combine_equivalent_zones` ? See chunk symmeterize_equivalent_zones_description.bak file in this folder when ticket resolved.
```
oerc0122, line 764:
```suggestion
..
   Re# 1447 should we reimplement `combine_equivalent_zones` ? See chunk
   symmeterize_equivalent_zones_description.bak file in this folder when ticket
   resolved.
```

oerc0122 (changes requested):
Use comment syntax

documentation/user_docs/docs/manual/Symmetrising_etc.rst
abuts: and this change does not work either. I had it initially but it work for first line so I need to use `replace` syntaxis

oerc0122: Well, it works on my machine. You should not use the replace syntax because that is not what you're doing

oerc0122: I suspect the reason it didn't work is that you lacked the newline after the `..`, c.f. [here](https://stackoverflow.com/questions/4783814/how-to-comment-a-string-in-restructured-text)

abuts: Provided reference suggests putting `..`  for each commented line. It is not what anybody wants.

abuts: replacement is exactly what I am doing ( where  possible).

oerc0122: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#comments

PR #1444, opened 2023-12-13
removing old interface to write handle

abuts:
This PR should be condidered #1435 is merged.  It addresses ticket #1403
These change remove all instances of internal write handle and `get_new_handle` functions from code base so all filebacked operations are performed through external write handle connected with PageOp class.

### DIFF
--- a/_test/test_sym_op/test_cut_sym_cube.m
+++ b/_test/test_sym_op/test_cut_sym_cube.m
@@ -24,7 +24,7 @@ classdef test_cut_sym_cube < TestCase
             obj@TestCase(name)
         end
 
-        function test_cut_sym_identity_stripped(obj)
+        function test_cut_sym_identity_stripped(~)
             tsqw = sqw.generate_cube_sqw(10);
 
             res_sqw = cut(tsqw, line_proj([1 0 0], [0 1 0]), ...
--- a/horace_core/algorithms/cut.m
+++ b/horace_core/algorithms/cut.m
@@ -66,11 +66,13 @@ if is_string(source)
     % We expect either a .sqw or .dnd file, throw an error otherwise.
     ldr = sqw_formats_factory.instance().get_loader(source);
     % If we're cutting, don't ever want to load whole file
-    sqw_dnd_obj=obj_from_faccessor(ldr, n_object, sqw_only, dnd_only);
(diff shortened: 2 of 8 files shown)


cmarooney-stfc (approved):

_test/test_sym_op/test_cut_sym_cube.m
```diff
@@ -24,7 +24,7 @@
             obj@TestCase(name)
         end
 
-        function test_cut_sym_identity_stripped(obj)
```
oerc0122, line 27: Shouldn't be in this PR.

horace_core/symop/@Symop/Symop.m
```diff
@@ -159,16 +159,8 @@ function disp(obj)
...
-                    curr_page = pix.data;
-                    for i = numel(obj):-1:1
-                        sel = ~obj(i).in_irreducible(curr_page(1:3, :), proj{:});
-                        curr_page(1:3, sel) = obj(i).transform_vec(curr_page(1:3, sel));
-                    end
-                    pix = pix.format_dump_data(curr_page);
-                end
-                pix = pix.finalise();
+                error('HORACE:Symop:not_implemented', ...
+                    'Transforming filebased pixels is not currently implemented');
```
oerc0122, line 163:
```suggestion
                    'Transforming file-backed pixels is not currently implemented');
```

horace_core/algorithms/cut.m
```diff
@@ -66,11 +66,13 @@
     % We expect either a .sqw or .dnd file, throw an error otherwise.
     ldr = sqw_formats_factory.instance().get_loader(source);
     % If we're cutting, don't ever want to load whole file
-    sqw_dnd_obj=obj_from_faccessor(ldr, n_object, sqw_only, dnd_only);
+    sqw_dnd_obj=obj_from_faccessor(ldr, n_object, sqw_only, dnd_only, ...
```
oerc0122, line 69: Shouldn't be in this PR.

_test/test_sym_op/test_cut_sym_cube.m
abuts: I would disagree with this approach.

Issue #564, opened 2021-02-12
Inconsistent Error Handling


oerc0122:
Currently the PACE project (particularly legacy code) is using multiple methods of handling try-catch paths, many of which are obscuring backtraces, reducing their utility:
E.g.
- Passing error states and messages back through arguments and optionally throwing
  ```
  function [data_out, fitdata, ok, mess] = fit (obj, varargin)

  if nargout<3
      throw_error = true;
  else
      throw_error = false;
  end

  [...]

  if ~ok_sim
    ok = false;
    if throw_error, error_message(mess), else, return, end
  end

  (@mfclass/fit)
  ```
- Throwing MATLAB errors and performing global catches and rethrows
  ```
  try
     obj = set_bfun@mfclass (obj, varargin{:});
  catch ME
      error(ME.message)
  end

  (@mfclass_tobyfit/set_bfun)
  ```

[ ] Write discussion ADR about how best to handle errors
[ ] Implement decisions from ADR and start to enforce a consistent handling scheme throughout

PR #1449, opened 2023-12-15
Remove ok, mess from `size_array_split` to avoid error obfuscation

oerc0122:
Remove `ok`, `mess` as part of #564 error handling.
Slightly refactor `size_array_split` to flatten function.
Make functions which previously used `ok`, `mess` use the error generated a as a cause.

### DIFF
--- a/herbert_core/utilities/classes/@data_op_interface/binary_op_manager.m
+++ b/herbert_core/utilities/classes/@data_op_interface/binary_op_manager.m
@@ -111,12 +111,17 @@ elseif isnumeric(w1)
     % w1 is a double array; w2 must have class 'classname'
     if ~isscalar(w1)
         size_stack1 = size(w2);
-        [size_root1, ok] = size_array_split (size(w1), size(w2));
-        if ~ok
-            error('HERBERT:data_op_interface:invalid_argument', ...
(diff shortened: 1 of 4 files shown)


herbert_core/utilities/classes/@data_op_interface/binary_op_manager.m
```diff
@@ -111,12 +111,17 @@
...
-            error('HERBERT:data_op_interface:invalid_argument', ...
-                ['Unable to resolve the numeric array into a stack of arrays, ',...
-                'with stack size matching the object array size.']);
+
+        try
+            size_root1 = size_array_split (size(w1), size(w2));
+        catch ME
+            err = MException('HERBERT:data_op_interface:invalid_argument', ...
+                             ['Unable to resolve the numeric array into a stack of arrays, ',...
+                              'with stack size matching the object array size.']);
```
abuts, line 120: I do not think you need to catch/rethrow the error here as `size_array_split` informs about the issue well enough. Checks and exception here only hide the reason. Better then before but just throwing would be clearler (and shorter)

The same opinion about couple of places below

abuts (commented):
Generally OK thouth  I would not catch error from `size_array_split` and let it propagate.. Think about it.

abuts (approved):

PR #1456, opened 2023-12-30
slim fixed

abuts:
Small changes fixing/enabling slim and unit test ensuring it works


horace_core/sqw/@sqw/rebin_sqw.m
```diff
@@ -22,118 +22,118 @@
...
-        % dimensionality
-        error('HORACE:rebin_sqw:invalid_argument', ...
-              'input object must be sqw type with detector pixel information');
-    end
-
-    if isempty(varargin)
-        error('HORACE:rebin_sqw:invalid_argument', ...
-              'requires at least one other argument');
-    end
+if ~isa(win, 'sqw')
```
oerc0122, line 25: It's not possible for `win` to not be an `sqw` here as I understand it.

horace_core/sqw/@sqw/slim.m
```diff
@@ -13,6 +13,12 @@
 % -------
 %   wout    Output sqw object or array of szqw objects
 
-wout = mask_pixels_random_fraction(win, 1/reduce);
+if reduce<1
+    error('HORACE:sqw:invalid_argument', ...
+        'reduce factor has to be bigger then 1. It is %s', ...
```
oerc0122, line 18: Message does not match condition. 
```suggestion
        'reduce factor has to be bigger than 0. It is %s', ...
```
or 
```
if reduce <= 1
```

oerc0122 (changes requested):

horace_core/sqw/@sqw/slim.m
abuts: no, 1. 1 -- all remains more then 1 -- appropriate fraction removed.

oerc0122: In this case `1.0` would be accepted which presumably is not what you want?

abuts: 1 will leave all pixels which makes no sence but may be useful for some algorithms

horace_core/sqw/@sqw/rebin_sqw.m
abuts: Correct statement but to be fair, I have not touched `rebin_sqw` -- it should be different ticket. I had a look at it dealing with slim and modified spaces as it happens almost automatically. `rebin` should be overlooked and modified properly but it is technical dept undelated to `slim`.
If you do comparison ignoring changes in spaces you will not see changes to rebin

horace_core/sqw/@sqw/slim.m
```diff
@@ -13,6 +13,12 @@
 % -------
 %   wout    Output sqw object or array of szqw objects
 
-wout = mask_pixels_random_fraction(win, 1/reduce);
+if reduce<1
+    error('HORACE:sqw:invalid_argument', ...
+        'reduce factor has to be more or equal to 1. It is %s', ...
```
oerc0122, line 18:
```suggestion
        'reduce factor has to be greater than or equal to 1. It is %s', ...
```

PR #1461, opened 2024-01-04
cache.py updated to python 3 and iDAAAS version of Python

abuts:
Existing utility updated to Python 3 and the Python avaliable on IDAAAAS

It is not much to test here, as it is not unit testable (or I do not know how to test it automatically)
It works on IDAAAS as expected -- I've just tested it. Unfortunately, does not help any more.

### DIFF
--- a/horace_core/admin/cache.py
+++ b/horace_core/admin/cache.py
@@ -89,7 +89,7 @@ def read_chunk(filename,start,size,buf_size,progress,n_workers):
     with fh as f:
         while got < size:
             bl = f.read(block)
-            got = got+block
+            got += block
             if got+block> size:
                 block = size-got
             if progress:
@@ -126,8 +126,8 @@ def process_file(argi):
     end_ch   = block_size
     if buf_size == 0:
         buf_size = block_size
-    if buf_size > sys.maxsize/1024:
-        buf_size = sys.maxsize/1024
+    if buf_size > sys.maxsize//1024:
+        buf_size = sys.maxsize//1024
 
     for i in range(0,n_threads):
         if end_ch > file_size:
--- a/horace_core/admin/cache.py
+++ b/horace_core/admin/cache.py
@@ -2,6 +2,7 @@
 import multiprocessing
 import time
 import sys
+import os
 import argparse
 #Read file using multiple threads to place file into CEPH cache.
 
@@ -113,11 +114,9 @@ def process_file(argi):
     buf_size = argi['buffer']
 
     # Estimate the file size
-    fh = open(filename,'rb',buffering=0)
-    fh.seek(0,2);
-    file_size = fh.tell()
-    #print 'File size=',nbytes
-    fh.close()
+    file_size = os.path.getsize(filename)
+    print ('File size={0:.0f}GB'.format(file_size/(1024*1024*1024)))
+
 
     # Evaluate the parameters of the file reading jobs.
     block_size = int(file_size/n_threads)+1
--- a/horace_core/admin/cache.py
+++ b/horace_core/admin/cache.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3.8
+#!/usr/bin/env python3
 import multiprocessing
 import time
 import sys
@@ -87,7 +87,7 @@ def read_chunk(filename,start,size,buf_size,progress,n_workers):
     got   = 0
     with fh as f:
         while got < size:
-            bl = fh.read(block)
+            bl = f.read(block)
             got = got+block
             if got+block> size:
                 block = size-got
--- a/horace_core/admin/cache.py
+++ b/horace_core/admin/cache.py
@@ -1,5 +1,4 @@
-#!/usr/bin/python
-#
+#!/usr/bin/python3.8
 import multiprocessing
 import time
 import sys
@@ -82,13 +81,11 @@ def read_chunk(filename,start,size,buf_size,progress,n_workers):
     if progress:
         prog_rep = progress_rep(size,n_workers)
         
-    fh = open(filename,'rb')
-    if fh<0:
-        raise ValueError("Can not open file: "+filename)
+    fh = open(filename,'rb',buffering=0)
     fh.seek(start,0)
     block = buf_size
     got   = 0
-    with open(filename, "rb") as f:
+    with fh as f:
         while got < size:
             bl = fh.read(block)
             got = got+block
@@ -97,7 +94,7 @@ def read_chunk(filename,start,size,buf_size,progress,n_workers):
             if progress:
                 prog_rep.check_progress(got)
     if progress:
-       print ""
+       print("")
     success=True
     return success
 
@@ -116,9 +113,7 @@ def process_file(argi):
     buf_size = argi['buffer']
 
     # Estimate the file size
-    fh = open(filename,'rb')
-    if fh<0:
-        raise ValueError("Can not open file: "+filename)
+    fh = open(filename,'rb',buffering=0)
     fh.seek(0,2);
     file_size = fh.tell()
     #print 'File size=',nbytes
@@ -132,10 +127,10 @@ def process_file(argi):
     end_ch   = block_size
     if buf_size == 0:
         buf_size = block_size
-    if buf_size > sys.maxint/1024:
-        buf_size = sys.maxint/1024
+    if buf_size > sys.maxsize/1024:
+        buf_size = sys.maxsize/1024
 
-    for i in xrange(0,n_threads):
+    for i in range(0,n_threads):
         if end_ch > file_size:
             block_size = file_size-start_ch
         chunk_size.append(block_size)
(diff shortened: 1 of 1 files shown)


horace_core/admin/cache.py
```diff
@@ -82,13 +81,11 @@ def read_chunk(filename,start,size,buf_size,progress,n_workers):
     if progress:
         prog_rep = progress_rep(size,n_workers)
         
-    fh = open(filename,'rb')
-    if fh<0:
-        raise ValueError("Can not open file: "+filename)
+    fh = open(filename,'rb',buffering=0)
```
oerc0122, line 84: The other form (`with`) is the standard way of doing things.

horace_core/admin/cache.py
```diff
@@ -1,5 +1,4 @@
-#!/usr/bin/python
-#
+#!/usr/bin/python3.8
```
oerc0122, line 1: Do not specify like this, this is not transferable. 

`#!/usr/bin/env python3`

horace_core/admin/cache.py
```diff
@@ -82,13 +81,11 @@ def read_chunk(filename,start,size,buf_size,progress,n_workers):
     if progress:
         prog_rep = progress_rep(size,n_workers)
         
-    fh = open(filename,'rb')
-    if fh<0:
-        raise ValueError("Can not open file: "+filename)
+    fh = open(filename,'rb',buffering=0)
     fh.seek(start,0)
```
oerc0122, line 85: Unnecessary

horace_core/admin/cache.py
```diff
@@ -116,9 +113,7 @@ def process_file(argi):
     buf_size = argi['buffer']
 
     # Estimate the file size
```
oerc0122, line 115: Python has tools for getting the file size. Why this?

horace_core/admin/cache.py
```diff
@@ -132,10 +127,10 @@ def process_file(argi):
     end_ch   = block_size
     if buf_size == 0:
         buf_size = block_size
-    if buf_size > sys.maxint/1024:
-        buf_size = sys.maxint/1024
+    if buf_size > sys.maxsize/1024:
```
oerc0122, line 130: Probably wants to be integer division
`sys.maxsize//1024`

horace_core/admin/cache.py
```diff
@@ -146,7 +141,7 @@ def process_file(argi):
     # Start parallel jobs:
     job_list = [];
     #read_chunk(filename,chunk_beg[0],chunk_size[0],buf_size,True,n_threads)
-    for nthr in xrange(n_threads):
+    for nthr in range(n_threads):
```
oerc0122, line 144: This should be done with a `map` and `with pool` not a loop

horace_core/admin/cache.py
```diff
@@ -82,13 +81,11 @@ def read_chunk(filename,start,size,buf_size,progress,n_workers):
...
         
-    fh = open(filename,'rb')
-    if fh<0:
-        raise ValueError("Can not open file: "+filename)
+    fh = open(filename,'rb',buffering=0)
     fh.seek(start,0)
     block = buf_size
     got   = 0
-    with open(filename, "rb") as f:
+    with fh as f:
```
oerc0122, line 88: This is not standard Python style, does not handle errors properly.

horace_core/admin/cache.py
```diff
@@ -82,13 +81,11 @@ def read_chunk(filename,start,size,buf_size,progress,n_workers):
...
         while got < size:
             bl = fh.read(block)
```
oerc0122, line 90: Why in `with X as f` does this refer to `fh`?

horace_core/admin/cache.py
```diff
@@ -162,10 +157,11 @@ def process_file(argi):
...
 
-    parser = argparse.ArgumentParser(add_help=True, version='0.1',description='Read file using multiple threads to place file into CEPH cache')
-    parser.add_argument('filename',action='store', type=str,default="",help='file to read to cache')
-    parser.add_argument('-n',action='store', dest='nthreads',type=int,default=16,help='number of threads to process file. Default is 16 threads')
-    parser.add_argument('-b',action='store', dest='buffer',type=int,default=4096,help='Buffer size to read each chunk of data. Default is 4096 bytes.')
+    parser = argparse.ArgumentParser(add_help=True,description='Read file using multiple threads to place file into CEPH cache')
+    parser.add_argument('filename',action='store', type=str,default="",help='file to read for caching in CEPHS')
+    parser.add_argument('--nthreads','-n',action='store', dest='nthreads',type=int,default=16,help='number of threads to process file. Default is 16 threads')
+    parser.add_argument('--buffer','-b',action='store', dest='buffer',type=int,default=4096,help='Buffer size to read each chunk of data. Default is 4096 bytes.')
+    parser.add_argument('-version',action = 'version',version = '0.1')
```
oerc0122, line 164: Standards suggest using `--` for long names

horace_core/admin/cache.py
```diff
@@ -82,13 +81,11 @@ def read_chunk(filename,start,size,buf_size,progress,n_workers):
...
-        raise ValueError("Can not open file: "+filename)
+    fh = open(filename,'rb',buffering=0)
     fh.seek(start,0)
     block = buf_size
     got   = 0
-    with open(filename, "rb") as f:
+    with fh as f:
         while got < size:
             bl = fh.read(block)
             got = got+block
```
oerc0122, line 91: Should probably use `+=`

horace_core/admin/cache.py
```diff
@@ -82,13 +81,11 @@ def read_chunk(filename,start,size,buf_size,progress,n_workers):
...
-    fh = open(filename,'rb')
-    if fh<0:
-        raise ValueError("Can not open file: "+filename)
+    fh = open(filename,'rb',buffering=0)
     fh.seek(start,0)
     block = buf_size
     got   = 0
-    with open(filename, "rb") as f:
+    with fh as f:
         while got < size:
```
oerc0122, line 89: For python 3.8, should use Walrus here

horace_core/admin/cache.py
```diff
@@ -116,9 +113,7 @@ def process_file(argi):
     buf_size = argi['buffer']
 
     # Estimate the file size
-    fh = open(filename,'rb')
-    if fh<0:
-        raise ValueError("Can not open file: "+filename)
+    fh = open(filename,'rb',buffering=0)
```
oerc0122, line 116: Again, should be `with`

oerc0122 (changes requested):

cmarooney-stfc (commented):
noting Jacob's comments on python usage, nothing further from me to add. OK when Jacob approves

horace_core/admin/cache.py
abuts: some machines (old machines) on IDAAAS still use 3.6 so for compartibility better to leave as it is.

horace_core/admin/cache.py
abuts: That's useful

horace_core/admin/cache.py
abuts: ? opening file? How?

horace_core/admin/cache.py
abuts: yes, and I am using it in when reading chunk-of-file but for single seek using it looks confusing.

oerc0122: You do not need the seek and this file is only coincidentally closed which is very confusing

oerc0122 (changes requested):

horace_core/admin/cache.py
oerc0122: Then the seek should be in the with block

horace_core/admin/cache.py
abuts: probably Python 2 did not have it.

horace_core/admin/cache.py
oerc0122: I just realised the original comment got cut off it should read
> Unnecessary `whence` argument, 0 is the default and regardless this should be using the [`enum`s defined in `os`](https://docs.python.org/3/library/io.html#io.IOBase.seek)

horace_core/admin/cache.py
oerc0122: No, [it has](https://docs.python.org/3/library/os.html#os.stat) been in there since at least 2.2 which is where the docs history stops

abuts: modified.

horace_core/admin/cache.py
abuts: Do not think so. This is the procedure for single thread .

horace_core/admin/cache.py
abuts: the errors with file opening've dealt above.

horace_core/admin/cache.py
abuts: that's what is done

PR #1471, opened 2024-01-09
Chunking Tobyfit algorithm

### LINKED ISSUE #760: Running out of memory on resolution calc for large npix
C.f. #1348 for other necessary changes to Tobyfit
Blocked by #1340 

When trying to run a resolution convolution calculation ("Tobyfit") with a medium sized dataset of 88M pixels, I run out of memory in an IDAaaS large VM session (120GB RAM). 

The memory error is in the calculation of the `dq_mat` by the [dq_matrix_DGdisk](https://github.com/pace-neutrons/Horace/blob/master/horace_core/Tobyfit/dq_matrix_DGdisk.m#L67) function (`dq_matrix_DGfermi` also has this issue). 

`dq-mat` is a stack of 4x11 matrices (a `(4,11,npix)`-sized array), but it is formed by multiplying a 4x6 matrix with a 6x11 matrix. These are currently all allocated in memory so we need `(4*6 + 6*11)*npix=90*npix` elements allocated in memory plus the end product of `44*npix` elements. For 88M pixels this is 47GB. If we chunk the intermediate matrix-matrix multiplication the total memory requirement would be less and could allow this calculation to proceed.

Before the `dq_matrix_*` function is called some intermediate matrices (four 3x3x`npix` matrix stacks `d_mat`, `f_mat`, `s_mat` and `spec_to_rlu`) and vectors (five `npix`-length vectors `wi`, `wf`, `xa`, `x1` and `x2`) are already calculated, which already uses `32*npix` elements of memory (11GB), and during the calculation an addition 3x3x`npix` matrix stack and 7 `npix`-length vectors are created (6GB).

The calculation started with about 60GB memory already used so this pushed it beyond the system's limit.

oerc0122:
Compute main array computations of Tobyfit in chunks to reduce high-water memory load on system.

Switch to computing `qw` (pixel location) on-the-fly and only storing `en`(ergy) in the lookup tables

Separate computation of the resolution-convolution function into its own respective functions (for DGdisk/DGfermi) paving the way towards using `PageOp`-, `apply`-, or method-wise implementations.

Naive implementation does a lot of allocate-deallocate of (smaller) memory chunks and a couple of unnecessarily repeated (though light-load) operations.

Decisions to be made over chunking ratio currently specified as `hpc_config.mem_chunk_size / 10` for simplicity, despite algorithm ultimately resulting in `44`x`npix` sized arrays. 

Performance not confirmed.

Added tests for chunked computation.
Loosened tolerances on tests to make them not seed-dependent. 

Fixes #760


abuts (commented):
Changes look reasonable, but I still do not understand why the code for `DGdisk` and `DGfermi` have to be different. At the firsts glance it can be the same resconv function with disk or fermi input. 

This will make code much more maintainable and changes more tranceable

horace_core/sqw/@sqw/tobyfit_DGfermi_resconv.m
```diff
@@ -115,16 +117,63 @@
...
+    if reset_state
+        if ~isempty(state_in{i})
+            rng(state_in{i})
+        end
+    else
+        state_out{i} = rng;     % capture the random number generator state
+    end
+
+    npix = win(i).data.npix(:);
+    [npix_chunks, idxs] = split_vector_fixed_sum(npix, max_pix_size);
```
tgperring, line 156: Return argument 'idxs' is not used anywhere, so can be deleted.

horace_core/sqw/@sqw/tobyfit_DGfermi_resconv.m
```diff
@@ -115,16 +117,63 @@
...
+                                  refine_moderator, modshape, ...
+                                  pix_bin_regions(j), pix_bin_regions(j+1));
+
+    end
+    wout(i) = recompute_bin_data(wout(i));
+end
+
+end
+
+function sqw_obj = compute_resconv(sqw_obj, iw, lookup, sqwfunc, pars, ...
```
tgperring, line 172: I like to see a clear demarcation between one function and another in a file - typically two blank lines and then a line of 80 minus signs. It helps to make the code readable.

horace_core/sqw/@sqw/tobyfit_DGfermi_resconv.m
```diff
@@ -160,136 +239,79 @@
...
-        [win(i), pars{1}] = refine_crystal_strip_pars (win(i), xtal, pars{1});
-
-        % Update s_mat, spec_to_rlu, alatt and angdeg because in general the
-        % crystal orientation and lattice parameters will have changed
-        [~,s_mat,spec_to_rlu,alatt,angdeg]=sample_coords_to_spec_to_rlu(win(i).experiment_info);
-
-        % Recompute Q because crystal orientation will have changed (don't need to update qw{4})
-        qw(1:3) = calculate_q (ki(irun), kf, detdcn, spec_to_rlu(:,:,irun));
+% Recompute Q on-the-fly
+qw = calculate_q(ki(irun), kf, detdcn, spec_to_rlu(:,:,irun));
```
tgperring, line 244: The cell array 'qw' no longer contains energy (which is the origin of the w standing in for an omega) - rename to, say, just 'q' to avoid confusion.

horace_core/sqw/@sqw/tobyfit_DGfermi_resconv.m
```diff
@@ -138,11 +187,41 @@
...
+npix = idx_end - idx_start + 1; %sqw_obj.pix.num_pixels;
+[irun, idet] = parse_pixel_indices(sqw_obj, (idx_start:idx_end)');     % returns column vectors
+
+% Get detector information for each pixel in the sqw object
+% size(x2) = [npix,1], size(d_mat) = [3,3,npix], size(f_mat) = [3,3,npix]
+% and size(detdcn) = [3,npix]
+[x2, detdcn, d_mat, f_mat] = detector_table.func_eval_ind (iw, irun, idet, @detector_info);
+
+% Catch case of refining crystal orientation
+if refine_crystal
```
tgperring, line 216: The if refine_crystal ... end block (lines 216 - 224) only need to be called once for any one sqw object, not repeated for every chunk.  Accordingly, the block can be moved to immediately before line 155. If I recall correctly, the call to sample_coords_to_spec_to_rlu is quite time consuming due to the digging down into unique_objects_container &/or unique_reference_container, so this likely will have a practical effect on performance as well as it logically belonging in the main function.

horace_core/sqw/@sqw/tobyfit_DGfermi_resconv.m
```diff
@@ -138,11 +187,41 @@
...
+    [~,s_mat,spec_to_rlu,alatt,angdeg]=sample_coords_to_spec_to_rlu(sqw_obj.experiment_info);
 
-% Perform resolution broadening calculation
-% -----------------------------------------
-% Package parameters as a cell for convenience
-if ~iscell(pars), pars={pars}; end
+end
 
 % Catch case of refining moderator parameters
 if refine_moderator
```
tgperring, line 227: The if refine_moderator ...end block only needs to be done once in the call to tobyfit_DGfermi_resconv. This block can be moved to immediately before line 139. Like the case of if refine_crystal ... end  I think it affects performance, as well as logically belonging earlier as it is repeating the same assignment for each chunk of each sqw object.

horace_core/sqw/@sqw/tobyfit_DGfermi_resconv.m
```diff
@@ -160,136 +239,79 @@
...
-        % Strip out crystal refinement parameters and reorient datasets
-        [win(i), pars{1}] = refine_crystal_strip_pars (win(i), xtal, pars{1});
-
-        % Update s_mat, spec_to_rlu, alatt and angdeg because in general the
-        % crystal orientation and lattice parameters will have changed
-        [~,s_mat,spec_to_rlu,alatt,angdeg]=sample_coords_to_spec_to_rlu(win(i).experiment_info);
-
-        % Recompute Q because crystal orientation will have changed (don't need to update qw{4})
-        qw(1:3) = calculate_q (ki(irun), kf, detdcn, spec_to_rlu(:,:,irun));
+% Recompute Q on-the-fly
```
tgperring, line 243: Performing in-the-fly calculation of q has a pleasing side effect. Because calculate_q is no longer called by tobyfit_DGfermi_resconv_init, the file ...\Horace\horace_core\Tobyfit\calculate_q.m can be deleted. The present method of sqw (i.e. tobyfit_DGfermi_resconv) uses the identical file in the @sqw private folder ...\Horace\horace_core\sqw\@sqw\private\calculate_q.m. We can therefore keep the sqw private function,  delete the copy in  ...\Horace\horace_core\Tobyfit, and thereby eliminate a technical debt banana skin.

horace_core/sqw/@sqw/tobyfit_DGdisk_resconv.m
```diff
@@ -115,16 +117,63 @@
...
+    if reset_state
+        if ~isempty(state_in{i})
+            rng(state_in{i})
+        end
+    else
+        state_out{i} = rng;     % capture the random number generator state
+    end
+
+    npix = win(i).data.npix;
+    [npix_chunks, idxs] = split_vector_fixed_sum(npix, max_pix_size);
```
tgperring, line 156: Return argument 'idxs' is not used anywhere, so can be deleted.

horace_core/sqw/@sqw/tobyfit_DGdisk_resconv.m
```diff
@@ -115,16 +117,63 @@
...
+                                  refine_moderator, modshape, ...
+                                  pix_bin_regions(j), pix_bin_regions(j+1));
+
+    end
+    wout(i) = recompute_bin_data(wout(i));
+end
+
+end
+
+function sqw_obj = compute_resconv(sqw_obj, iw, lookup, sqwfunc, pars, ...
```
tgperring, line 172: I like to see a clear demarcation between one function and another in a file - typically two blank lines and then a line of 80 minus signs. It helps to make the code readable.

horace_core/sqw/@sqw/tobyfit_DGdisk_resconv.m
```diff
@@ -166,132 +243,78 @@
...
-    % Set random number generator if necessary, and save if required for later
-    if reset_state
-        if ~isempty(state_in{i})
-            rng(state_in{i})
-        end
-    else
-        state_out{i} = rng;     % capture the random number generator state
-    end
+% Recompute Q on-the-fly
+qw = calculate_q(ki(irun), kf, detdcn, spec_to_rlu(:,:,irun));
```
tgperring, line 248: The cell array 'qw' no longer contains energy (which is the origin of the w standing in for an omega) - rename to, say, just 'q' to avoid confusion.

horace_core/sqw/@sqw/tobyfit_DGdisk_resconv.m
```diff
@@ -141,10 +190,38 @@
...
+npix = idx_end - idx_start + 1; %sqw_obj.pix.num_pixels;
+[irun, idet] = parse_pixel_indices(sqw_obj, (idx_start:idx_end)');     % returns column vectors
+
+% Get detector information for each pixel in the sqw object
+% size(x2) = [npix,1], size(d_mat) = [3,3,npix], size(f_mat) = [3,3,npix]
+% and size(detdcn) = [3,npix]
+[x2, detdcn, d_mat, f_mat] = detector_table.func_eval_ind (iw, irun, idet, @detector_info);
+
+% Catch case of refining crystal orientation
+if refine_crystal
```
tgperring, line 216: The if refine_crystal ... end block (lines 216 - 224) only need to be called once for any one sqw object, not repeated for every chunk.  Accordingly, the block can be moved to immediately before line 155. If I recall correctly, the call to sample_coords_to_spec_to_rlu is quite time consuming due to the digging down into unique_objects_container &/or unique_reference_container, so this likely will have a practical effect on performance as well as it logically belonging in the main function.

tgperring (changes requested):

horace_core/sqw/@sqw/tobyfit_DGdisk_resconv.m
```diff
@@ -115,16 +117,63 @@
...
+% Perform resolution broadening calculation
+% -----------------------------------------
+% Package parameters as a cell for convenience
+if ~iscell(pars)
+    pars={pars};
+end
+
+reset_state=caller.reset_state;
+
+% Factor 10 because of 44 x npix array
```
cmarooney-stfc, line 139: not sure "because" is correct. It's a provisional one-size fits all value. You may wish to refer to the pr number or paste your comment from the pr here

_test/test_TF_components/test_tobyfit_cuts.m
```diff
@@ -66,7 +66,7 @@
             obj.fe_arr = fe_arr;  % three short const-E cuts, 140-160, 160-180, 180-200 meV
             obj.rb_arr = rb_arr;    % two const-Q cuts, 2-10 meV
 
-            tol_sig = 0.25;        % tolerance as multiple of st. dev. of reference value
+            tol_sig = 0.5;        % tolerance as multiple of st. dev. of reference value
```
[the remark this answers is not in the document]
oerc0122: I think not taken in isolation this changes much. It's of note here because it is changing, but going forward I'm not sure the reasoning is necessarily relevant.

horace_core/Tobyfit/tobyfit_DGfermi_resconv_init.m
```diff
@@ -95,9 +95,9 @@
 %       Cell array of widths of energy bins, one array per dataset
 %           dt          Time widths for each pixel (s), size [npix,1]
 %
-%       Cell arrays of q,w and transformation arrays, one array per dataset
-%           qw          Cell array size [1,4] with components of momentum (in rlu) and energy
-%                       for each pixel [Columns of length npix]
+%       Arrays of w and transformation arrays, one array per dataset
```
tgperring, line 98: The line has inherited what was already an erroneous comment about transformation arrays: for consistency with e.g. line 95:
''%    Cell array of energy bin centres, one array per dataset"
Do we want this as a column instead of a row?

horace_core/Tobyfit/tobyfit_DGdisk_resconv_init.m
```diff
@@ -93,9 +93,9 @@
 %       Cell array of widths of energy bins, one array per dataset
 %           dt          Time widths for each pixel (s), size [npix,1]
 %
-%       Cell arrays of q,w and transformation arrays, one array per dataset
-%           qw          Cell array size [1,4] with components of momentum (in rlu) and energy
-%                       for each pixel [Columns of length npix]
+%       Arrays of w and transformation arrays, one array per dataset
```
tgperring, line 96: The line has inherited what was already an erroneous comment about transformation arrays: for consistency with e.g. line 93:
''%    Cell array of energy bin centres, one array per dataset"
Do we want this as a column instead of a row?

tgperring (approved):

cmarooney-stfc (approved):

Issue #1460, opened 2024-01-04
update cache.py to python 3 and verify if it works as expected.


abuts:
`cache.py` is simple python utility which does multithread reading of sqw file to place it to CEPHS cache. 
its written on Python-2 which is not available on IDAAAS any more. In addition, simple modifications to this utility are not working properly on IDAAAS, so there is the question if they are working at all. 

The utility should be updated and checked if it operates as expected.

abuts:
closed through #1461

PR #1480, opened 2024-01-17
Pythonise cache.py

oerc0122:
- Convert to modern Python (min 3.6; f-strings) 
- Add type hints

Re #1460


abuts (approved):
It looks like it works so I do not mind having these changes. 
We do not have agreed code standards for Python and my knolege of it is rusty, so whatever works

PR #1496, opened 2024-01-19
head error

abuts:
These are the simple changes to `head` which warn user about the fact he uses old format `head`. May be it worth running this format but the necessary changes are simple so I think the user would manage


horace_core/sqw/@sqw/head.m
oerc0122, line 1: All of these changes apart from those in this file are completely unrelated to the stated PR's purpose.

abuts: well, I may reword the ticket as code quality changes as the bug here only with old code (which is not running anyway -- just bug a bit werid)

PR #1504, opened 2024-01-23
fixed issue with xml_write on higher Matlab versions

### LINKED ISSUE #1503: xml_write is broken with higher versions of MATLAB
like xml_read before, version check in xml_write have similar piece of code and is broken on higher versions  of MATLAB

abuts:
fixes #1503. Added common check for Matlab version. Its third party code, so I do not bother with fixing it too much.

### DIFF
--- a/herbert_core/utilities/xml_io_tools/check_MATLAB_version.m
+++ b/herbert_core/utilities/xml_io_tools/check_MATLAB_version.m
@@ -1,0 +1,15 @@
+function check_MATLAB_version()
+%CHECK_MATLAB_VERSION verifies if the MATLAB version is sufficient to run
+%xml_read/xml_write
+
+v = ver('MATLAB');
+vs = regexp(v.Version, '\d\d.\d','match','once'); % new Matlab after 2023a
+if isempty(vs)
+    vs = regexp(v.Version, '\d.\d','match','once'); % Matlab before 2023a
+end
+version = str2double(vs);
+if (version<7.1)
+    error('HERBERT:xml_io_toold:runtime_error', ...
+        'Your MATLAB version is too old. You need version 7.1 or newer.');
+end
+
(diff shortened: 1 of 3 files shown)


cmarooney-stfc (approved):

herbert_core/utilities/xml_io_tools/check_MATLAB_version.m
```diff
@@ -0,0 +1,15 @@
+function check_MATLAB_version()
+%CHECK_MATLAB_VERSION verifies if the MATLAB version is sufficient to run
+%xml_read/xml_write
+
+v = ver('MATLAB');
+vs = regexp(v.Version, '\d\d.\d','match','once'); % new Matlab after 2023a
+if isempty(vs)
+    vs = regexp(v.Version, '\d.\d','match','once'); % Matlab before 2023a
+end
```
oerc0122, line 9:
```suggestion
vs = regexp(v.Version, '\d\d?.\d','match','once'); 
```

oerc0122: But why do it this way? There's a [`verlessthan`](https://uk.mathworks.com/help/matlab/ref/verlessthan.html) function. Why would we implement our own?

oerc0122 (changes requested):

herbert_core/utilities/xml_io_tools/check_MATLAB_version.m
cmarooney-stfc: if you click on this link, it will tell you at the top that verlessthan is deprecated, and gives you the replacement

oerc0122: Deprecated as of 2020b, and we support back to 2018b

oerc0122: It is still supported, i.e. will work, it is just deprecated. That means it may be removed in future releases, but is not yet.

abuts: This is minor issue to fix third party code. I think what author did is good enough and further discussion is not worth time spend, so I am closing this issue.

PR #1509, opened 2024-01-24
1157 fix multiple hashing and experiment initialisation

cmarooney-stfc:
As per point 1 of #1157, objects are hashed twice, unnecessarily, for unique_object_containers. This does not happen for these containers themselves, but when used in unique_references_containers, this does happen, and is fixed by this PR.

This investigation also shows that sqw had its experiment_info incorrectly initialised by a property definition rather than a part of the constructor. This may also fix point 2 about clearing the container although as there are no instructions for reproducing it I cannot be certain.

It also shows that unique_references_containers are misleadingly tested by the present tests, as earlier tests perform initialisation of the persistent data which then allow later tests to proceed which would otherwise fail if run in isolation.

### DIFF
--- a/herbert_core/utilities/classes/@unique_objects_container/private/add_single_.m
+++ b/herbert_core/utilities/classes/@unique_objects_container/private/add_single_.m
@@ -1,4 +1,4 @@
-function [self,nuix] = add_single_(self,obj)
+function [self,nuix] = add_single_(self,obj,ix,hash)
 %ADD_SINGLE_ Add single object to the unique objects container
 
 
@@ -6,7 +6,24 @@ function [self,nuix] = add_single_(self,obj)
 % returned as the index to the object in the container.
 % hash is returned as the hash of the object. If ix is empty
 % then the object is not in the container.
-[ix,hash] = self.find_in_container(obj);
+%
+% Input
+% -----
+% self - the unique_objects_container in question
+% obj  - the object to be added to the container
+% ix   - the unique index of the object if it is already in the container. 
+%        this will have been found from a previous call of find_in_container
+% hash - the hash for obj previously made by that call to find_in_container
+% 
+% Output
+% ------
+% self - the modified container (modified by adding obj)
+% nuix - the insertion index at which obj is added in the container
+
+% if ix and hash are not specified, call find_in_container to get them
+if nargin<=2
+    [ix,hash] = self.find_in_container(obj);
+end
 
 % If the object is not in the container.
 % store the hash in the stored hashes
--- a/herbert_core/utilities/classes/@unique_references_container/unique_references_container.m
+++ b/herbert_core/utilities/classes/@unique_references_container/unique_references_container.m
@@ -498,10 +498,10 @@ classdef unique_references_container < serializable
                 error('HERBERT:unique_references_container:incomplete_setup', ...
                     'global name unset');
             end
-            [glindex, ~] = self.global_container('value',self.global_name_).find_in_container(inobj);
+            [glindex, hash] = self.global_container('value',self.global_name_).find_in_container(inobj);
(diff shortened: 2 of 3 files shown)


horace_core/sqw/@sqw/sqw.m
```diff
@@ -65,7 +65,7 @@
         % The class providing brief description of a whole sqw file.
         main_header_ = main_header_cl();
 
-        experiment_info_ = Experiment();
+        experiment_info_ = []; %Experiment();
```
oerc0122, line 68: So the array has the right type.
```suggestion
        experiment_info_ = Experiment.empty();
```

herbert_core/utilities/classes/@unique_references_container/unique_references_container.m
```diff
@@ -498,10 +498,10 @@
...
                     'global name unset');
             end
-            [glindex, ~] = self.global_container('value',self.global_name_).find_in_container(inobj);
+            [glindex, hash] = self.global_container('value',self.global_name_).find_in_container(inobj);
             if isempty(glindex)
                 glcont = self.global_container('value',self.global_name_);
-                [glcont,glindex] = glcont.add(inobj);
+                [glcont,glindex] = glcont.add_single_(inobj,[],hash);
                 if glindex == 0
                     % object was not added
```
oerc0122, line 506: What is the "standard" signal that an object has not been added? Reading this is appears that empty `glindex` or `glindex == 0`, but that is two different signals. Is this correct?

oerc0122 (changes requested):

abuts (commented):
This is a comment:
As object hash is caclulated in `find_in_container` I would assume that external hash should be propagated up to `fild_in_container` to return if provided. 

This looks like clearer code and less problem for maintaining it. 

Bearing in mind idea of storing hash with object, current change addresses local problem so fine.

herbert_core/utilities/classes/@unique_objects_container/private/add_single_.m
tgperring, line 1: I am not competent to give a review without spending time familiarising myself with the structure of the problem.

horace_core/sqw/@sqw/sqw.m
cmarooney-stfc: But I don't want it to have the right type. I want it to fail as obviously as possible if this variable has not been reinitialised correctly, probably in the constructor.

herbert_core/utilities/classes/@unique_references_container/unique_references_container.m
cmarooney-stfc: Very well spotted - it's an issue - but what is happening is that if isempty(glindex), the object has not been found and is added as a new object. If glindex==0 then nothing has been added to the container by add_single_, so the call has failed - not the same as not being found. However, a look at add_single_ shows that an object will always be added, so the if glindex==0 block is paranoia, and should be removed - I will do so.

herbert_core/utilities/classes/@unique_objects_container/private/add_single_.m
```diff
@@ -1,12 +1,29 @@
...
+%        this will have been found from a previous call of find_in_container
+% hash - the hash for obj previously made by that call to find_in_container
+% 
+% Output
+% ------
+% self - the modified container (modified by adding obj)
+% nuix - the insertion index at which obj is added in the container
+
+% if ix and hash are not specified, call find_in_container to get them
+if nargin<=2
```
oerc0122, line 24: Suggestion:
I prefer not using `nargin` to check arguments as it limits future function signature changes. Instead, prefer explicitly checking `exist('variable', 'var')`

oerc0122 (approved):

Issue #1466, opened 2024-01-04
H4UX: uoffset in old Horace and uoffset in new Horace are different things


abuts:
Horace 4 defines property offset as expressed in `hkl`  where `hkl` are Miller indexes on primary reciprocal lattice and `uoffset` -- hidden property mapped to `img_offset` where `img_offset` expressed in the coordinate system related to the cut projection.  (taret coordinate system)

Horace-3 defines `uoffset` as offset in image_coordinate system?

The script below generates and images below shows four cuts done in Horace 3.6.3 and Horace 4.0 made from sqw file produced by horace_demo.

It looks like mapping old `uoffset` into `offset` assuming old `uoffset` was provided in `hkl` would sort majority of the problems

``` MATLAB
sqw_file = fullfile(pwd,'fe_demo.sqw');
hv = horace_version('-num');
%hv = 0;
if hv < 400
    proj = struct();
    proj.u = [1,0,0];
    proj.v = [0,1,0];
    proj.type = 'rrr';
else
    proj = line_proj;
    proj.type = 'rrr';
end
cc2o=cut_sqw(sqw_file,proj,[-3,0.05,3],[-3,0.05,3],[-0.1,0.1],[180,220]);
plot(cc2o);
keep_figure;
proj.u = [1,1,0];    
proj.v = [1,-1,0];        

if hv<400
    proj.uoffset = [0,-2,0];
else
    proj.offset = [0,-2,0];
    %proj.uoffset = [0,-2,0];
end
cc2o_s1=cut_sqw(sqw_file,proj,[-3,0.05,3],[-1,0.05,5],[-0.1,0.1],[180,220]);
plot(cc2o_s1);
keep_figure;

if hv<400
    proj.uoffset = [1,-1,0];
    proj1 = proj;
    proj1.uoffset= [1,0,0];
else
    proj.offset = [1,-1,0];
    proj1 = proj;
    proj1.offset = 0;
    proj1.img_offset = [1,0,0];    
end
cc2o_sa=cut_sqw(sqw_file,proj,[-3,0.05,3],[-1,0.05,5],[-0.1,0.1],[180,220]);
plot(cc2o_sa);
keep_figure;
cc2o_sb=cut_sqw(cc2o_s1,proj1 ,[-3,0.05,3],[-1,0.05,5],[-0.1,0.1],[180,220]);
plot(cc2o_sb);
keep_figure;

```
Horace 3 cuts:
[...]

PR #1521, opened 2024-02-01
Removed img_offset from projection

abuts:
This is agreed decision of #1466 consisting of rewiring old `uoffset` to `offset`
In addition projection code is simplified removing img_offset. This parameter may be provided later as part of cut parameters.


cmarooney-stfc (approved):

herbert_core/utilities/strings/str_to_iarray.m
oerc0122, line 1: This should not be in this PR

herbert_core/utilities/strings/str_trim_cellstr.m
oerc0122, line 1: This should not be in this PR

horace_core/sqw/@DnDBase/change_crystal.m
```diff
@@ -40,12 +40,17 @@
         rlu_corr = this_alignment.get_corr_mat(obj.proj);
         rlu_to_u = wout(i).proj.bmatrix();
         proj = wout(i).proj;
-        proj.img_offset(1:3)=rlu_corr*wout(i).img_offset(1:3)';
-        proj = proj.set_ub_inv_compat(rlu_corr/rlu_to_u);
-        wout(i).proj = proj;
+        %proj.img_offset(1:3)=rlu_corr*wout(i).img_offset(1:3)';
```
oerc0122, line 43: What's the differece between l. 48 and l. 43 besides an intermediate variable? If so, do we need this comment here?

herbert_core/utilities/strings/str_trim_cellstr.m
```diff
@@ -32,3 +32,10 @@
 
 all_non_empty = all(non_empty);
 cstr = cstr(non_empty);
+
+
+function is = str_is_cellstr(strin)
```
oerc0122, line 37:
```suggestion
function is = str_is_cellstr(string)
```
And elsewhere

oerc0122 (changes requested):
Beyond the bits fiddling with the utilities this all looks fine

horace_core/sqw/@DnDBase/change_crystal.m
abuts: 43 is the same as 5 other rows but using `img_offset`. It is removed so can not be used, but having it would be benefitial and more clear code. As single row is summary of 5 others with clear meaning, it is worth keeping it (and adding more comments)

abuts:
fixed thorugh #1521

Issue #1495, opened 2024-01-19
H4UX: replicate_dnd should produce reasonable warning wrt replicate_sqw(___,'-set_pix') argument


abuts:
`replicate(sqw)` got new `-set_pix` argument, which allow it to build image, fully aligned with pixels. `repliace(dnd)` called with this parameters fails with unhelpful error message. 
`replicate(dnd,___'-set_pix')` should warn about extra parameters and proceed ignoring it.

It '-set_pix' option should also work with `dnd` replicated on reference `sqw`

abuts:
is addressed through Re #1527

PR #1527, opened 2024-02-05
replicate has all possible options and correct warnings

abuts:
Addresses #1495. 

These changes addresses small inconvenience in `replicate` providing common interface and common behaviour for `sqw` and `dnd` objects


horace_core/sqw/@sqw/sqw.m
```diff
@@ -390,6 +390,8 @@ function  save_xye(obj,varargin)
         function npix = get.num_pixels(obj)
             npix = obj.pix_.num_pixels;
         end
+        % if sqw object has actual pixels or pixelles object
```
cmarooney-stfc, line 393:
```suggestion
        % if sqw object has actual pixels or pixels object
```
avoid confusion that there might be boy pixels and girl pixels

horace_core/sqw/@DnDBase/replicate.m
```diff
@@ -18,11 +18,16 @@
 %           - The data is expanded along the plot axes of wref that are
 %           integration axes of win.
 %           - The annotations etc. are taken from the reference dataset.
+% Optional:
+% '-set_pix' -- if provided, wref object should be sqw objects with pixels.
+%               in this case result would be sqw object(s) with pixels set
+%               to reprouce replicated image, defined by input dnd object.
```
cmarooney-stfc, line 24:
```suggestion
%               to reproduce replicated image, defined by input dnd object.
```
You have said both that wref should be sqw and that input object is dnd. Something is inconsistent. Also as the current class is dndbase, checking if this is appropriate for sqw input/output

horace_core/sqw/@DnDBase/replicate.m
```diff
@@ -18,11 +18,16 @@
...
 %
 % Output:
 % -------
 %   wout    Output dataset object (or array of objects). It is dnd object
-%           with the same dimensionality as wref.
+%           with the same dimensionality as wref or sqw object with the
+%           same dimensionality as input dnd object and
```
cmarooney-stfc, line 30: comment stopped short without completion

cmarooney-stfc (changes requested):

horace_core/sqw/@sqw/sqw.m
abuts: hm -- pixel-less - no pixels are present. "Pixelles" team may higack useful name but I would not even bother thinking about them. `pixels object` seems is something different from pixelles

PR #1543, opened 2024-02-12
Refactor replicate_array

tgperring:
This PR:
(1) refactors the code of the Herbert utility functions replicate_array, replicate_iarray and replicate_logarray;
(2) creates unit tests for these functions;
(3) deletes the redundant function @sqw/private/replicate_array. 

Refers to ticket #848 . This PR makes PR #1284 redundant and that PR should be deleted.

The purpose is to:
(1) resolve the lack of any tests of replicate_array and the companion functions;
(2) remove a technical debt banana skin of two identically named and identical purpose functions but which use different methods, and eliminate a third method used for integer arrays in replicate_iarray.
(3) increase performance. 

The refactoring uses the Matlab intrinsic function repelem, which has almost identical function. Comprehensive speed tests for different sizes of the input arguments show that it is between a factor of 1 and 25 faster than the faster of the two replicate_array functions on TGP's XPS 15 laptop (vintage: 2023), which is in turn a factor of 2 to 8 times faster than the slower of the two versions. In the case of replicate_iarray the speed improvement is between a factor of 1 and 2.  All three functions replicate_array, replicate_iarray and replicate_logarray now make a single line call to repelem; the only difference is that replicate_array and companions enforce an additional check on the size of the input arguments.

In general, the call 
      :
      vout = replicate_array (vin, n)
      :
can in any new coding be substituted by
      :
      vout = repelem (vin(:), n(:))
      vout = vout(:)
      :
and similarly for the companion function. These refactored function are essentially now just legacy functions.


herbert_core/utilities/maths/replicate_logarray.m
```diff
@@ -1,26 +1,44 @@
...
+%   >> vout = replicate_logarray (v, n)
+%  use:
+%   >> vout = repelem (logical(v(:)), n(:)) % conversion to logical is needed
+%                                           % if the class of v is not 'logical'
+%
 %
 % Input:
 % ------
 %   v       Array of logical values i.e. true or false
+%           If the class of v is not 'logical', an attempt to convert is mode.
```
cmarooney-stfc, line 19:
```suggestion
%           If the class of v is not 'logical', an attempt to convert is made.
```

herbert_core/utilities/maths/replicate_logarray.m
```diff
@@ -1,26 +1,44 @@
 function vout = replicate_logarray (v, n)
 % Replicate logical array elements according to list of repeat indicies
 %
-%   >> ivout = replicate_logarray (v, n)
+%   >> vout = replicate_logarray (v, n)
```
cmarooney-stfc, line 4: given this is legacy back comptaibility, I guess we'll leave the name, but log could mean logarithm

herbert_core/utilities/maths/replicate_array.m
```diff
@@ -3,6 +3,15 @@
 %
 %   >> vout = replicate_array (v, n)
 %
+% This is a legacy function retained for backwards compatibility. The same
+% behaviour is accomplished with the Matlab intrinsic function repelem which was
+% introduced in release R2015a:
+%  Instead of:
+%   >> vout = replicate_array (v, n)
+%  use:
+%   >> vout = repelem (v(:), n(:))
```
cmarooney-stfc, line 12: basically all 3 replicate functions now have identical implementation. Guess ok as it's legacy back compatibility

_test/test_utilities_herbert/test_replicate_array.m
```diff
@@ -0,0 +1,344 @@
...
+    % Test behaviour of replicate_array, replicate_iarray and replicate_logarray.
+    % The original algorithms of thes functions have been superseded by the
+    % Matlab intrinsic function repelem.
+    
+    methods
+        %--------------------------------------------------------------------------
+        % Test replicate_array
+        %--------------------------------------------------------------------------
+        function test_array_empty_v_empty_n (~)
+            % Empty input array, empty n ==> zerp(0,1) output
```
cmarooney-stfc, line 11:
```suggestion
            % Empty input array, empty n ==> zero(0,1) output
```
the zerp function is too important for this

cmarooney-stfc (changes requested):

herbert_core/utilities/maths/replicate_logarray.m
tgperring: Done

herbert_core/utilities/maths/replicate_array.m
tgperring: They are not quite identical: replicate_logarray has a type conversion as well. It is certainly my goal to use repelem in the future. However, there are several places where they are used around Horace and in my own code (although repelem was introduced in 2015a, I only learned of its existence a week or two ago.

_test/test_utilities_herbert/test_replicate_array.m
tgperring: A quick search reveals that there is no publicly known zerp function. We have the opportunity to create it ourselves and launch it on an expectant world.

cmarooney-stfc (approved):

cmarooney-stfc (approved):

PR #1546, opened 2024-02-13
Refactor sawtooth_iarray and add unit tests

tgperring:
Add copious unit tests of Herbert utility function sawtooth_iarray, correct an inconsistency of output array size if the output is empty, and make robust to non-integer input array elements with consistent behaviour with Matlab colon operator.

sawtooth_iarray is called alongside replicate_iarray in some Horace methods that manipulate sqw objects. It is appropriate that sawtooth_iarray is made similarly robust and tested.

The refactoring of sawtooth_iarray has left the speed of the function essentially unchanged. Fairly wide testing yields same time +/- 35%

### DIFF
--- a/_test/test_utilities_herbert/test_sawtooth_iarray.m
+++ b/_test/test_utilities_herbert/test_sawtooth_iarray.m
@@ -81,29 +81,23 @@ classdef test_sawtooth_iarray < TestCase
             vout_ref = [1;2;3;4;1;2];
             assertEqual(vout, vout_ref);
         end
-                
-        function test_sawtooth_iarray_real_n (~)
-            % Reals rounded down to nearest integer
-            n = 3.7;
-            vout = sawtooth_iarray (n);
-            vout_ref = [1;2;3];
-            assertEqual(vout, vout_ref);
-        end
-                
-        function test_sawtooth_iarray_real_n_ltZero (~)
-            % Negative reals equivalent to zero
-            n = -0.7;
-            vout = sawtooth_iarray (n);
-            assertEqual(vout, zeros(0,1));
+        
+        function test_negative_n_ERROR (~)
+            % Throw an error because one of the elements of n is negative
+            n = [4,-2,2];
+            f = @()sawtooth_iarray(n);
+            ME = assertExceptionThrown (f, 'HERBERT:sawtooth_iarray:invalid_argument');
+            assertTrue(contains(ME.message, ...
+                'The elements of the input argument must all be nonnegative integer values'))
         end
         
-        function test_sawtooth_iarray_realArray_n (~)
-            % Reals rounded down to nearest integer; array input that exercises
-            % this in several cases
-            n = [-0.7, 3.7, 2, 0, 4.0, -0.001];
-            vout = sawtooth_iarray (n);
-            vout_ref = [1;2;3;1;2;1;2;3;4];
-            assertEqual(vout, vout_ref);
+        function test_real_n_ERROR (~)
+            % Throw an error because one of the elements of n is non-integer
+            n = [4,1.8,2];
+            f = @()sawtooth_iarray(n);
+            ME = assertExceptionThrown (f, 'HERBERT:sawtooth_iarray:invalid_argument');
+            assertTrue(contains(ME.message, ...
+                'The elements of the input argument must all be nonnegative integer values'))
         end
                                 
     end
--- a/_test/test_utilities_herbert/test_sawtooth_iarray.m
+++ b/_test/test_utilities_herbert/test_sawtooth_iarray.m
@@ -1,0 +1,110 @@
+classdef test_sawtooth_iarray < TestCase
+    % Test behaviour of Herbert utility function sawtooth_iarray.
+    
+    methods
+        function test_sawtooth_iarray_empty_n (~)
+            % Empty input array, empty n ==> zero(0,1) output
+            n = [];
+            vout = sawtooth_iarray (n);
+            assertEqual(vout, zeros(0,1));
+        end
+        
+        function test_sawtooth_iarray_zero_n (~)
+            % Output should be zero(0,1)
+            n = 0;
+            vout = sawtooth_iarray (n);
+            assertEqual(vout, zeros(0,1));
+        end
+        
+        function test_sawtooth_iarray_scalar_n (~)
+            % Output should be a column independent of shape of input
+            n = 3;
+            vout = sawtooth_iarray (n);
+            vout_ref = [1;2;3];
+            assertEqual(vout, vout_ref);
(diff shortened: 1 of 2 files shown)


herbert_core/utilities/maths/sawtooth_iarray.m
```diff
@@ -1,38 +1,47 @@
 function ivout = sawtooth_iarray (n)
-% Create array of successive runs [1,2,3..n(1),1,2,3...n(2),...]'
+% Create column vector [1;2;3..n(1); 1;2;3...n(2); ...] from integer array n
 %
 %   >> ivout = sawtooth_iarray (n)
 %
+% This is a 'sawtooth' array as each element n(i) defines a right angle triangle
+% with length n and values 1,2,3,...n(i).
```
cmarooney-stfc, line 7: this right angle statement baffles me. Sawtooth (other array descriptions of this type say jagged, but that's fine) does not seem compatible with this triangle picture

herbert_core/utilities/maths/sawtooth_iarray.m
```diff
@@ -1,38 +1,47 @@
...
+%
+%
 % Input:
 % ------
-%   n       List of length of each sawtooth section (integer)
-%          (Zeros correspond to zero length section i.e. they are
+%   n       Array with the lengths of each sawtooth section
```
cmarooney-stfc, line 12: The array you create requires n to elucidate its structure. Having this as a class would make sense as then you can embed n into it. It would also make n(i)==0 sensible as you can pad unused rows. Otherwise banning n(i)==0 may be more sensible,

herbert_core/utilities/maths/sawtooth_iarray.m
```diff
@@ -1,38 +1,47 @@
...
-% Original author: T.G.Perring
+% EXAMPLES
+%   ivout = sawtooth_iarray(n)
 %
+%   n                 ivout
+% ----------------------------------------------------------------------
+%   0               zeros(0,1)      % column vector with zero length
+%   [2,3,1]         [1;2;1;2;3;1]
+%   [2,0,1]         [1;2;1]         % zero length middle sawtooth section
+%   [2.3,-0.7,3.99] [1;2;1;2;3]     % reals rounded downwards
```
cmarooney-stfc, line 35: it would be preferable if non-integer values could be banned

_test/test_utilities_herbert/test_sawtooth_iarray.m
```diff
@@ -0,0 +1,110 @@
...
+            n = [2;3;1];
+            vout = sawtooth_iarray (n);
+            vout_ref = [1;2;1;2;3;1];
+            assertEqual(vout, vout_ref);
+        end
+        
+        function test_sawtooth_iarray_arr_v_arr_n (~)
+            % Output should be a column independent of shape of input
+            n = [2,3,1;...
+                4,5,6];
```
cmarooney-stfc, line 46: allowing this gives rise to an ambiguity in the order of accessing n, this examples could be 2,4,3,5,1,6

cmarooney-stfc (changes requested):

herbert_core/utilities/maths/sawtooth_iarray.m
tgperring: I was hoping the text would help, but probably the description on line 2 i.e. it is the array
[1;2;3..n(1); 1;2;3...n(2); ...]
is more useful. I've deleted the description. on lines 6 and 7

herbert_core/utilities/maths/sawtooth_iarray.m
tgperring: I don't really understand this comment. The output is a column vector, so all rows have length one; the section for n(i) has size [n(i),1]; n(i)=0 results in a section with size [0,1] which is equivalent to being ignored. This behaviour is consistent with replicate_iarray (v, n) where n(i) results in a section of the output column vector with n(i) repeats of the value v(i) i.e. size [n(i),1]. In the case of n(i) = 0 the is a section of size [0,1].

The result is that replicate_iarray(v,n) and sawtooth_iarray(n) can be used together as they always produce a column vector with the same length = sum(n(:)), as is done in @PixelDataBase, and cases of n(i)=0 for one or more i are handled as required by some multifit commands. I've added a comment that the size of the output is [sum(n(:)), 1]

I think defining a class just adds complication and performance overheads to what is a straightforward function.

_test/test_utilities_herbert/test_sawtooth_iarray.m
tgperring: It is assumed that the order of the elements is always the standard Matlab (and Fortran) order i.e. in this case 2,4,3,5,1,6. This is standard Matlab behaviour.

herbert_core/utilities/maths/sawtooth_iarray.m
```diff
@@ -1,38 +1,37 @@
...
-nn0=nn0(:);
-if ~isempty(nn0)
-    nend=cumsum(nn0);
-    tmp=ones(nend(end),1);
-    ix=1+nend(1:end-1);
-    if ~isempty(ix)
-        tmp(ix)=tmp(ix)-double(nn0(1:end-1));
-    end
-    ivout=cumsum(tmp);
+n = n(n>0);
```
cmarooney-stfc, line 29: not really needed after the error on line 25

herbert_core/utilities/maths/sawtooth_iarray.m
```diff
@@ -1,38 +1,37 @@
...
+if isempty(n)
+    ivout = zeros(0,1);
+elseif isscalar(n)
+    ivout = (1:n)';
 else
-    ivout=[];
+    offset = cumsum([0;n]);
+    ivout = (1:offset(end))' - repelem(offset(1:end-1), n);
```
cmarooney-stfc, line 36: I have stared at this blankly for some time. I'm sure it works, but a line of explanation would be good

cmarooney-stfc (changes requested):

cmarooney-stfc (changes requested):
Noticing you have requested a review but the comments at the end of sawtooth_iarray have not been addressed. maybe comments and request crossed

herbert_core/utilities/maths/sawtooth_iarray.m
tgperring: I've added an explanation of the algorithm. Note that I branch on the case of empty and scalar n to avoid the overhead of calling repelem for these simple cases. It keeps the algorithm optimal.

cmarooney-stfc (approved):

cmarooney-stfc (approved):

PR #1547, opened 2024-02-13
1350c cm detectors save load correctly

cmarooney-stfc:


cmarooney-stfc:
Merged by PR #1540 by Jacob (thanks) following incorrect early merge. This is now a preliminary draft to be closed without merging. All changed checked to have been incorporated into master except for test parallel parameters which are unchanged from previous master.

PR #1558, opened 2024-02-20
Respond to David's comments on documentation except ResConv

oerc0122:
- Migrate section on `Symop` to end and mark as advanced
- Suggest DGS planner 
- Rewrite parallel path issue
- Restate `symmetrise_sqw` restrictions

### DIFF
--- a/documentation/user_docs/docs/manual/Planning_a_Horace_scan.rst
+++ b/documentation/user_docs/docs/manual/Planning_a_Horace_scan.rst
@@ -2,6 +2,12 @@
 Planning a Horace scan
 ######################
 
+.. note::
+
+   While Horace includes its own planner it is worth noting that the
+   `MANTID DGS
+   planner <https://docs.mantidproject.org/v3.8.0/interfaces/DGSPlanner.html>`__
+   exists [1]_.
 
 horace_planner: a simple scan planner
 -------------------------------------
@@ -130,3 +136,35 @@ in exactly the same way as you would a real dataset. For example:
    :alt: 2d simulation
 
    Sample plot of fake dataset
+
+
+.. [1] The main feature differences at the time of writing are:
+
+   - The crystal orientation can be specified as a UB matrix instead
+     of as the u and v vectors which define the horizontal scattering
+     plane.
+
+   - It allows rotations about horizontal rotation axes (Horace only
+     allows rotation about the vertical axis). Which is not supported
+     by Horace.
+
+   However, the GUI is significantly different:
+
+   - Plots
+
+     * The Mantid DGS has a single plot window with a single projection
+       of the 3D q-space into 2D.
+
+     * Horace has 3 plots showing all three projections in to the
+       possible combinations of axes.
+
+   - Rescaling
+
+     * The Mantid DGS GUI is resizeable so is usable on small laptop
+       screens.
+
+     * The Horace GUI is not resizeable so is very difficult to view in
+       small laptop screens.
+
+   The result of this is that if you have a small screen it may be
+   favourable to use the MANTID planner rather than Horace's native one.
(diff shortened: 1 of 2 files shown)


documentation/user_docs/docs/manual/Planning_a_Horace_scan.rst
```diff
@@ -2,6 +2,12 @@
 Planning a Horace scan
 ######################
 
+.. warning::
+
+   While Horace includes its own planner it is potentially less feature complete
+   than the [MANTID DGS
+   planner](https://docs.mantidproject.org/v3.8.0/interfaces/DGSPlanner.html)
```
cmarooney-stfc, line 9: can we say something about when to use which of these two? Otherwise it's just frustrating

documentation/user_docs/docs/manual/Symmetrising_etc.rst
```diff
@@ -368,6 +260,13 @@ pixels according to the symmetry operations and accumulating the transformed
 pixels into the bins appropriately. This is done through the ``symmetrise_sqw``
 function, the signature for which is below:
 
+.. warning::
+
+   Due to restrictions related to the `irreducible region`_, ``symmetrise_sqw`` is
+   only defined for ``SymopReflection`` and ``SymopRotation`` and **NOT**
```
cmarooney-stfc, line 266:
```suggestion
   only defined for ``SymopReflection`` and ``SymopRotation`` and **NOT** for
```

documentation/user_docs/docs/manual/Symmetrising_etc.rst
```diff
@@ -763,7 +667,7 @@ dataset (thereby increasing the counts four-fold).
 
 ..
    Re# 1447 should we reimplement `combine_equivalent_zones` ? See chunk symmeterize_equivalent_zones_description.bak_Re#1447 file in this folder when ticket resolved.
-   
+
```
cmarooney-stfc, line 670: for previous 2 lines, put hash next to 1447 and preferably spell symmetrize correctly

documentation/user_docs/docs/manual/Symmetrising_etc.rst
```diff
@@ -773,11 +677,122 @@ Limitations
...
   ``mem_chunk_size`` and is approximately 10 times larger then the amount,
-  specified by this parameter.
\ No newline at end of file
+  specified by this parameter.
+
+
+Symop Methods - Advanced
+========================
+
+``Symop`` objects have methods to transform a variety of methods to transform
```
cmarooney-stfc, line 693:
```suggestion
``Symop`` objects have methods to transform a variety of 
```

cmarooney-stfc (changes requested):

documentation/user_docs/docs/manual/Planning_a_Horace_scan.rst
mducle: They're basically the same, there are only two features which Mantid has extra:

* It allows the crystal orientation to be specified as a UB matrix instead of as the `u` and `v` vectors which define the horizontal scattering plane
* It allows rotations about horizontal rotation axes (Horace only allows rotation about the vertical axis).

These are really minor features - Horace does not allow computing `sqw` files generated by rotation about either horizontal axes (parallel to or perpendicular to the incident beam), so it doesn't make sense to support the second feature (Mantid `MDWorkspace`s can be generated from scans about the horizontal axes).

I think actually how the GUI is presented is much more important:

* The Mantid DGS has a single plot window with a single projection of the 3D q-space into 2D. Horace has 3 plots showing all three projections in to the possible combinations of axes.
* The Mantid DGS GUI is resizeable so is actually viewable on small laptop screens. The Horace GUI is not resizeable so is very difficult to view in small laptop screens.

So, personally I would say - if you have a laptop with a small screen use Mantid, if not use Horace.

documentation/user_docs/docs/manual/Symmetrising_etc.rst
oerc0122: Sadly, that's the name of the file (which causes other issues on Linux), but renaming it isn't for here. This is in a comment and not visible in the rendered document.

PR #1576, opened 2024-02-26
Add parser for bracket-matching

### LINKED ISSUE #1505: H4UX: mf_leg_to_new is not very user friendly and does not converts binds properly
`mf_leg_to_new` is a python function which converts legacy Tobyfit string into modern Tobyfit. 

The description is very scetchy so inexperienced user does not know what it is doing. 

Why it is Python? Nothing there which benifits Python only. 

the binds converted using this function do not work (e.g.:

```MATLAB
pars_in = [0.4,-0.8,0.07, 0.5,-0.22,0.07, 0.5,0.22,0.07, 0.4,0.8,0.07, 0.4,1.2,0.07];
pars_free = [1,   1,   0,   1,    1,   0,   1,   1,   0,   1,  1,   0,   1,  1,   0];
pars_bind = {{2,11,0,-1}, {5,8,0,-1}}; % ensures symmetry about x=0
% The syntax above is that each binding is a cell array of 4 numbers:
% {i1, i2, func_id, ratio}, where i1 and i2 are indices of the parameters
% to be bound, func_id is the index of the function (0 for foreground, 1
% for background), and ratio is their ratio. So {2,11,0,-1} means that
% parameter 11 will always be -1*(value_parameter_2). If the last two
% numbers are omitted, the ratio used will be the ratio of the starting
% values of those parameters.
% type: help fit_func to get a more detailed explanation
%
%[wfit, fitdata] = fit_func(my_cut(1)-0.35, @multigauss, pars_in, pars_free, pars_bind);
kk = multifit(my_cut(1)-0.35);
kk = kk.set_fun(@multigauss);
kk = kk.set_pin(pars_in);
kk = kk.set_free(pars_free);
kk = kk.set_bind(pars_bind);
[wfit, fitdata] = kk.fit();
```
This does not work

oerc0122:
Replace regex argument extraction with rudimentary bracket-matching algorithm to catch multiply nested arguments correctly

Fixes #1505

### DIFF
--- a/tools/mf_leg_to_new.py
+++ b/tools/mf_leg_to_new.py
@@ -6,6 +6,8 @@ J. Wilkins 8-9-2023
 
 import sys
 import re
+from typing import Iterator
+
 
 ARG_TYPE = ("fun", "pin", "free", "bind")
 OLD_OPT_TO_NEW = {'fit': 'fit_control_parameters',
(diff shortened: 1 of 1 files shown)


tools/mf_leg_to_new.py
```diff
@@ -14,10 +16,53 @@
...
+    accum = ""
+    str_mode = False
+
+    for ind, char in enumerate(string):
+        if char in STRDELIM:
+            if not str_mode:
+                stack.append(char)
+                str_mode = True
+
+            elif stack[-1] == char:  # Stay in string mode if not matching quote
```
cmarooney-stfc, line 37: the comment is what you're not doing, which is confusing

cmarooney-stfc (changes requested):

Issue #962, opened 2023-03-02
Tests for `shift_pixels` insufficient


oerc0122:
`Horace/_test/test_sqw_class/test_migrated_apis.m` has tests for `shift_pixels` which are likely insufficient.

abuts:
Wery strange algorithm shift energy bins. 
E.g, assume 1D-dispersion 
x=1:10
y = 25-(x-5).^2

Then, following the algorithm one gets:
xx = x+y
![image](https://github.com/pace-neutrons/Horace/assets/1109606/d620aad1-2626-4b85-a766-5739ad6a0415)
orange line: x-y, 
blue line: xx-y. What the idea?

abuts:
That's certainly correct. I have no doubt about it. The question is why would you need this transformation.

PR #1579, opened 2024-02-27
Extracted shift pixels into separate test file

abuts:
This is partial work on ticket #962 which is currently low priority ticket awaiting discussion. 

Despite that, this work is done.

The changes take the tests for `shift_pixels` from `migrated_api` (currently disabled) and put them into separate test_file `test_shift_pixels` (still disabled)


cmarooney-stfc (changes requested):
i would be happy to approve this based on the title and the code.

but I cannot understand the description. Please can you make it clear why these actions fix the ticket.

PR #1580, opened 2024-02-27
Fix rel3.6 jenkins

oerc0122:
Should solve problem on Rocky8 with old Jenkinsfile

### DIFF
--- a/tools/build_config/Jenkinsfile
+++ b/tools/build_config/Jenkinsfile
@@ -87,7 +87,8 @@ pipeline {
 
   environment {
     MATLAB_VERSION = utilities.get_param('MATLAB_VERSION', pli.matlab_release)
-    CMAKE_VERSION = utilities.get_param('CMAKE_VERSION', '3.19')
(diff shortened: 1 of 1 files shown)


tools/build_config/Jenkinsfile
```diff
@@ -87,6 +87,7 @@ pipeline {
 
   environment {
     MATLAB_VERSION = utilities.get_param('MATLAB_VERSION', pli.matlab_release)
+    MATLAB_MODULE = utilities.get_param('MATLAB_VERSION', pli.matlab_release.minus('R'))
     CMAKE_VERSION = utilities.get_param('CMAKE_VERSION', '3.19')
```
abuts, line 91: looks good except it does not have 3.19 cmake on the machine. To avoid unnecessary compexity I would add `latest` instead of `3.19` here. 

https://anvil.softeng-support.ac.uk/jenkins/job/Examples/job/List%20of%20available%20software/OS%20Image=rocky8/lastSuccessfulBuild/artifact/module-list.txt

I believe cmake runs previous versions?

abuts (approved):

cmarooney-stfc (commented):
ok

PR #1599, opened 2024-03-08
Removing unneeded hashes from the container population stages of Horace

cmarooney-stfc:
(1) Further use of existing hashes to prevent further hash calculations
(2) Remove duplicate calculations of Experiment contents

### DIFF
--- a/_test/test_experiment/test_experiment_cnstrct_and_properties.m
+++ b/_test/test_experiment/test_experiment_cnstrct_and_properties.m
@@ -67,6 +67,7 @@ classdef test_experiment_cnstrct_and_properties < TestCase
             exp(3).filename = 'a3';
             detectors = repmat(IX_detector_array(),1,numel(exp));
 
+            lastwarn('');
             exper= Experiment(detectors,instruments,samples,exp);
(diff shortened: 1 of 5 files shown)


_test/test_experiment/test_experiment_cnstrct_and_properties.m
```diff
@@ -75,6 +76,7 @@ function test_nontrivial_runid_map(~)
                 'Samples in experiment are defined but their lattice is undefined');
             assertEqual(b, ...
                 'HORACE:Experiment:lattice_undefined');
+            lastwarn('');
```
abuts, line 80: Minor: you do not need to do `lastwarn`  here as next warning will superseed it. 
also 
```
clob = set_temporary_warning('off','whatever was issued for tests') 
```
should be used to reduce noise while running tests (the warning will be issued anyway, but would not be displayed)

abuts (approved):
looks fine. minor notes provided

tgperring (approved):
I have no comments on the code, and am happy with the PR. I think my best test is going to be seeing if I can cause it to break in testing!

PR #1602, opened 2024-03-15
ubmat proj

abuts:
Further development and test fixing change things in such a way that it is full solution . One issue remain:

1) Image scales during alignment are seems incorrect.


horace_core/sqw/coord_transform/@ubmat_proj/get_line_proj.m
```diff
@@ -0,0 +1,12 @@
+function lp = get_line_proj(obj)
```
oerc0122, line 1: If we can get a `line_proj`, why do we need to define `ubmat_proj` in the first place? Could it not just be a function?

horace_core/sqw/coord_transform/@ubmat_proj/check_combo_arg.m
```diff
@@ -0,0 +1,19 @@
...
+% Generic method. Needs specific private function checkfields
+
+% Original author: T.G.Perring
+%
+% 	15 August 2009  Pass w to checkfields, so that checkfields can alter fields
+%                   of object. Because checkfields is a private method, the fields
+%                   can be altered using w.x=<new value> *without* calling
+%                   set.m. (T.G.Perring)
+%  03/04/2017       Checking only combo aruments as ivalid arguments can
+%                   not be set up separately using class setters
```
oerc0122, line 18:
```suggestion
%  03/04/2017       Checking only combo aruments as invalid arguments can
%                   not be set up separately using class setters
```

Also not convinced this was written in 2017

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
```diff
@@ -0,0 +1,387 @@
...
+    %  Defines coordinate transformations, used by cut_sqw when making
+    %  Horace cuts defined by rotation matrix
+    %
+    %  Object that defines the ortholinear projection operations
+    %
+    % Input accepting the structure:
+    %   >> proj = ubmat_proj(proj_struct)
+    %             where proj_struct is the
+    %             structure, containing any fields, with names, equal any
+    %             public fields of the line_proj class.
```
oerc0122, line 14: Why `line_proj` not `ubmat_proj`?

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
```diff
@@ -0,0 +1,387 @@
...
+    %   plus any of other arguments, provided as key-value pair e.g.:
+    %
+    %   >> proj = ubmat_proj(...,'offset',offset,...)
+    %   >> proj = ubmat_proj(...,'label',labelcellstr,...)
+    %   >> proj = ubmat_proj(...,'lab1',labelstr,...)
+    %                   :
+    %   >> proj = ubmat_proj(...,'lab4',labelstr,...)
+    %
+    % Minimal fully functional form:
+    %   >> proj =  ubmat_proj(u_to_rlu,'alatt',latice_parameters,'angdeg',lattice_angles_in_degrees);
```
oerc0122, line 37: It says above that you can truncate at any argument. How is this, then, the minimal fully functional form?

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
```diff
@@ -0,0 +1,387 @@
...
+    %
+    %IMPORTANT:
+    % if you want to use ubmat_proj as input for the cut algorithm, it needs
+    % at least one input parameter u_to_rlu, (or its default value) as
+    % the lattice parameters for cut will be taken from sqw object
+    % if not provided with projection.
+    %
+    % For independent usage u_to_rlu and lattice parameters (minimal fully
+    % functional form) needs to be specified. Any other parameters have
```
oerc0122, line 46: This also contradicts what's above

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
```diff
@@ -0,0 +1,387 @@
...
+        % ortho-ortho transformation to identify cells which may contribute
+        % to a cut. Correct value is chosen on basis of performance analysis
+        convert_targ_to_source=true;
+    end
+
+    properties(Access=protected)
+        % Cached properties value, calculated from input u_to_rlu matrix
+        type_  = 'aaa';
+        nonorthogonal_ = false;
+        uvw_cache_     = eye(4);
```
oerc0122, line 100: Why does `uvw_cache` have 4 4-vectors. It can't just be `uvw` in that case.

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
```diff
@@ -0,0 +1,387 @@
...
+        %
+        function v = get.v(obj)
+            v = obj.uvw_cache_(:,2)';
+        end
+        %
+        function w = get.w(obj)
+            w = obj.uvw_cache_(:,3)';
+        end
+        function cell = get.unit_cell(obj)
+            cell = [obj.uvw_cache_,[0;0;0];[0,0,0,1]];
```
oerc0122, line 144: Will this work? `uvw_cache` is defined as a 4x4 above, you can't append these

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
```diff
@@ -0,0 +1,387 @@
...
+            [obj,axes] = align_proj@aProjectionBase(obj,alignment_info,axes);
+        end
+    end
+    %======================================================================
+    methods(Access = protected)
+        function is = get_proj_aligned(obj)
+            is = obj.proj_aligned_;
+        end
+        function obj = set_proj_aligned(obj,val)
+            obj.proj_aligned_ = logical(val);
```
oerc0122, line 265: A bare set here seems dangerous. What if I pass a vector?

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
```diff
@@ -0,0 +1,387 @@
...
+            end
+            obj.uoffset_ = val(:)';
+            if obj.do_check_combo_arg_
+                obj = obj.check_combo_arg();
+            end
+        end
+        %------------------------------------------------------------------
+        function   contrib_ind= get_contrib_cell_ind(obj,...
+                cur_axes_block,targ_proj,targ_axes_block)
+            % get indexes of cells which may contributing into the cut.
```
oerc0122, line 293:
```suggestion
            % get indices of cells which may contribute to the cut.
```
What are "cells" though? "bins"?

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
```diff
@@ -0,0 +1,387 @@
...
+            %
+            if obj.convert_targ_to_source && isa(targ_proj,class(obj))
```
oerc0122, line 295:
```suggestion
            if obj.convert_targ_to_source && isa(targ_proj, 'ubmat_proj')
```
Try not to obscure what you mean by indirect references

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
```diff
@@ -0,0 +1,387 @@
...
+            %
+            % sets up target projection as the parent method.
+            % In addition:
+            % sets up matrices, necessary for optimized transformations
+            % if both projections are line_proj
+            %
+            obj = check_and_set_targ_proj@aProjectionBase(obj,val);
+            % if isa(obj.targ_proj_,'line_proj') && ~obj.disable_srce_to_targ_optimization
+            %     obj = set_ortho_ortho_transf_(obj);
+            % else
```
oerc0122, line 317: Debugging statement?

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
oerc0122, line 1: Where did `ubmatrix` previously fit into Horace 3.6? If `projaxes` became `line_proj`, what warrants this being considered a projection in its own right.

This does not seem to be a true projection, but some kind of means to recover a `line_proj` or at least a `line_proj`'s functionality. Is this the case?

Furthermore: 
- How do you intend this to be used? 
- Will a user ever need to instantiate one of these to perform a cut on a legacy sqw? 
- What sense will it make to a user particularly in how it differs from a `line_proj`? and if they use the "wrong" one, will they just get nonsense results?

oerc0122 (changes requested):
Pausing review here for clarification before continuing to review other files.

horace_core/sqw/coord_transform/@ubmat_proj/get_line_proj.m
abuts: The code, which is in `ubmat_proj` was previously in `line_proj`. The code is responsible for processing legacy data and distribured in over more than one functions. Though it is theoretically possible to move the changes to set of functions, class is the best pattern for that as encapsulates and separates legacy data and legacy functions together. 

Now, if we do not want to maintain legacy data, we just drop `ubmat_proj`. In addition, it is possible and may be necessary to add specific code to this class. (was planning to do that but now see low priority for this kind of operations)

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
abuts: Because if you have not provide these three arguments you can not use it. It will need more parameters

Description modified

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
abuts: do not see any contradictions

horace_core/sqw/coord_transform/@ubmat_proj/ubmat_proj.m
abuts: because it is how `u_to_rlu` was written in old code

PR #1604, opened 2024-03-18
Remove combo-arg checking while distributing IX_dataset objects

### LINKED ISSUE #1603: Failure on `distribute` with `IX_dataset`
Due to changes in underlying `serializable` object, `distribute` no longer works with `IX_Dataset` due to protections on `check_combo_arg`. 

Fix: wrap block in `obj.do_check_combo_arg = false;` and reset.

oerc0122:
Fixes #1603


abuts (approved):
Approach itself is questionable but reasonable action within this approac.

PR #1605, opened 2024-03-18
Avoid using `cell2mat` in `symmetrise_sqw`

### LINKED ISSUE #1598: Issue with multiple Symops in symmetrise
MATLAB  `cell2mat` fails to turn a cell array to a matrix if cell array contains certain things. 

Failure in `symmetrise_sqw` with array of Symops in some versions of matlab. Replace with `cat` in `validate_sqw`

oerc0122:
Fixes #1598

### DIFF
--- a/horace_core/sqw/@sqw/symmetrise_sqw.m
+++ b/horace_core/sqw/@sqw/symmetrise_sqw.m
@@ -264,7 +264,7 @@ function [sym, fold] = validate_sym(sym)
 
         elseif all(cellfun(@(x) isa(x, 'SymopReflection') || ...
                                 isa(x, 'SymopIdentity'), sym))
-            sym = cell2mat(sym);
+            sym = cat(1, sym{:});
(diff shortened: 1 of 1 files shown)


horace_core/sqw/@sqw/symmetrise_sqw.m
```diff
@@ -264,7 +264,7 @@
 
         elseif all(cellfun(@(x) isa(x, 'SymopReflection') || ...
                                 isa(x, 'SymopIdentity'), sym))
-            sym = cell2mat(sym);
+            sym = cat(1, sym{:});
```
cmarooney-stfc, line 267: noting that this forces a column vector, which is probably what you want, but cell2mat might give you a row vector under the same conditions. 

Once you've checked this, this is really an approval, go ahead and merge

cmarooney-stfc (changes requested):

Issue #1621, opened 2024-03-25
set_alignment_martix for PixelDataMemory


abuts:
set_alignment matrix for pixel data memory should align pixels and set unit alignment matrix. 

This what is happened anyway, as this operation recalculates pixel ranges so applies alignment to pixels. 

This is bug and not an improvement as misaligned memory based data are stored in sqw file as aligned but metadata state that they are misaligned, so when data are loaded, they are rotated (additionaly aligned) again.

abuts:
Meged to 1592_ubmat_proj and will be closed indirectly through Re #1592 ; PR #1602

Issue #1592, opened 2024-03-04
ubmat_proj class


abuts:
create projection class which is defined by ubmatrix directly

abuts:
> Please can you explain what this means and why it is needed.

It appears to be convenient way of separating legacy code. I though I will use it for more, but found other ways around the issues with alignment. 

It also may be useful for building projections for people who prefer to work with ubmatrix as source, but this is the request I remember from Mantid. 

Fixed through Re #1602

PR #1622, opened 2024-03-25
pix alignment

abuts:
This fixed the bug #1621, related to saving aligned pixel data to file.

### DIFF
--- a/_test/test_sqw_pixels/test_PixelAlignment.m
+++ b/_test/test_sqw_pixels/test_PixelAlignment.m
@@ -12,6 +12,33 @@ classdef test_PixelAlignment < TestCase & common_pix_class_state_holder
             hc = hor_config;
             obj.stored_config = hc.get_data_to_store();
         end
+        function test_save_filebacked_applies_alignment_on_sqw(~)
+            ref_sqw = sqw.generate_cube_sqw(10);
+            test_file_mal = fullfile(tmp_dir,'test_save_filebacked_misal.sqw');
+            test_file_al = fullfile(tmp_dir,'test_save_filebacked_align.sqw');
+            clOb = onCleanup(@()del_memmapfile_files({test_file_mal,test_file_al}));
+            save(ref_sqw,test_file_mal);
+
+            clCf = set_temporary_config_options(hor_config, 'mem_chunk_size', 500);
+            clW  = set_temporary_warning('off','HOR_CONFIG:set_mem_chunk_size');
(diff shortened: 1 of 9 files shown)


_test/test_sqw_pixels/test_PixelAlignment.m
```diff
@@ -12,6 +12,33 @@
...
+            test_sqw = sqw(test_file_mal,'file_backed',true);
+            % ensure we are testing paged filebacked object
+            assertTrue(test_sqw.is_filebacked);
+            assertEqual(size(test_sqw.pix.data,2),500);
+
+            al_matr = rotvec_to_rotmat2([0, 0, pi/4]);
+            % you should not do that with real sqw, as this invalidates
+            % pix-image correspondence. But fine for this test.
+            test_sqw.pix.alignment_matr = al_matr;
+            assertTrue(test_sqw.pix.is_misaligned);
```
cmarooney-stfc, line 33: Comparing this with line 79 maybe a problem, see comment there

_test/test_sqw_pixels/test_PixelAlignment.m
```diff
@@ -47,13 +76,13 @@ function test_finalize_alignment_membacked(~)
             al_matr = rotvec_to_rotmat2([0, pi/4, pi/4]);
 
             pdm.alignment_matr = al_matr;
-            assertTrue(pdm.is_misaligned);
+            assertFalse(pdm.is_misaligned);
```
cmarooney-stfc, line 79: for this, assigning non-eye alignment_matr makes misaligned false as presumably the alignment is assigned by the set. However, above line 33 assigning non-eye makes misaligned true. There seems to be an inconsistency here. Even if this is explicable, it makes the use of setting alignment problematic for the average user

cmarooney-stfc (changes requested):

_test/test_sqw_pixels/test_PixelAlignment.m
abuts: No inconsistency. In one situation, asignment done for memory-based object which is aligned on-fly. 
For other case, it is filebacked object and aligning it is presumably takes a long time. To avoid this, we make it misaligned and will perform alignment as side-operation along with other operations.

abuts:
Will not be merged directly as `crystal_alignment` tests are failing and this PR is part of PR #1602. It is merged to branch 1592_ubmat_proj and will be indirectly closed through ticket #1592

Issue #1620, opened 2024-03-25
Pixels not fully transformed in cut_sym


oerc0122:
Currently `cut` with symmetry does not transform the pixels when they are rebinned in a symmetrised cut. This means that the pixels are binned appropriately, but the connection between pixels and image is broken.

The solution to this is to pass the symmetry operations all the way through to `cut_accumulate_data_` which should transform the pixels following the cut akin to how `symmetrise_sqw` transforms the pixels.

abuts:
A projection uses generic `bin_pixels`  and generic `transform_pix_to_img` to bin pixels
`pix_transformed = obj.transform_pix_to_img(pix_candidates);` 
`transform_pix_to_img` should be modified to form:
`[pix_transformed,pix_candidates] = obj.transform_pix_to_img(pix_candidates);` 

and if projection has symmetry operation attached to it (in additional property), `pix_candidates` should be transformed using symmetry too. 

Alternatively,  overload `line_proj`  to `line_proj_rotation` & `line_proj_folding`  which have these methods specialized.

PR #1623, opened 2024-03-27
Add error on attempting to cut-sym with pix

oerc0122:
Quick fix for cut-sym with pixels

Partially addresses #1620


horace_core/sqw/@sqw/private/cut_accumulate_data_.m
```diff
@@ -168,6 +168,10 @@
     targ_proj, targ_axes, npix, s, e, ll, ...
     keep_precision, pixel_contrib_name)
 
+if numel(targ_proj) > 1
+    error('HORACE:cut:not_implemented', 'Cannot cut sym and return pixels')
+end
+
```
abuts, line 174: I would add words "not implemented yet" to the error message but hopefully not for long so this way ok too. 

You need to add '-nopix' to tests for cut_sym for tests to pas

abuts (approved):
fix the build as described and it would be decent temporary solution to merge ASAP

PR #1626, opened 2024-04-02
Fix issue with cut_sym not transforming pixels

### LINKED ISSUE #1620: Pixels not fully transformed in cut_sym

oerc0122:
Fixes #1620

### DIFF
--- a/horace_core/sqw/@sqw/private/cut_accumulate_data_.m
+++ b/horace_core/sqw/@sqw/private/cut_accumulate_data_.m
@@ -1,5 +1,5 @@
 function [s, e, npix, pix_out, unique_runid] = ...
-    cut_accumulate_data_(obj, targ_proj, targ_axes, keep_pixels, log_level)
+    cut_accumulate_data_(obj, targ_proj, targ_axes, keep_pixels, log_level, sym)
 %%CUT_ACCUMULATE_DATA Accumulate image and pixel data for a cut
 %
 % Input:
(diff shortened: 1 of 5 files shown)


horace_core/sqw/@sqw/private/cut_accumulate_data_.m
```diff
@@ -225,6 +221,7 @@
         [npix, s, e, pix_ok, unique_runid_l, pix_indx, selected] = ...
             targ_proj(i).bin_pixels(targ_axes(i), candidate_pix, npix, s, e);
 
+        candidate_pix = sym.transform_pix(candidate_pix, {}, selected);
         candidate_pix = candidate_pix.tag(selected);
```
abuts, line 225: As usual, do not understand the logic here. You loop over projections which project various areas of the sqw file to target area. But use only one sym operation for transforming pixels, while each projection should have its own. 
werid.

abuts (commented):
When it fixed it seems ok. Do not like why one should provide two interconnected parameters (`sym{i}` and `proj{i}`) separately and not to bind them together, but as it works -- fine with this.

horace_core/sqw/@sqw/private/cut_accumulate_data_.m
oerc0122: Fixed that

cmarooney-stfc (approved):

Issue #954, opened 2023-02-23
Verify the spherical coordinate system meaning and description


abuts:
Physical verification of the meaning of the spherical coordinate system, if it is correctly oriented vrt. the instrument and the description corresponds to the actual formulas used. See `spher_proj` class description (rows 5-...)

The question to answer are:
1) is the suggested coordinate system convenient for users. (e.g. phi calculated with correct sign and from proper direction)
2) Is the description provided for the spheirclal projection correct and comprehensive
3) Do we neeed to allow change in the spherical coordinate system -- redirect spheciral axes along other coordinate axes (not beam direction)
4) Do we want to allow offset (not the centre of coordinate system)
5) do we need other units for Q (currently A^-1) only (mod(hkl) ? nonorthogonal, eliptic)

6) spherical cut from dnd is probably incorrect or should be tested. If there are any cut from dnd (not axes aligned), do we need spherical cut too?

cmarooney-stfc:
@abuts to return to Christian Balz to arrange a face to face meeting to resolve this

abuts:
fixed by Re #1638

PR #1638, opened 2024-04-15
Curvilinear proj verification

abuts:
This PR partially deals with Re #954 and verifies some components of spherical/cylindrical projection operations. It is not total validation but the changes here are big enough to have PR and close this particular ticket. Other validation ticket will follow. 

The PR added number of tests for cylindrical cuts, some have physical meaning so it is obvious that they are correct. Some reproduce the results produced by previous implementation for cylindrical cuts. 

In addition, this ticket deals with validation of target ranges identification algorithm. Code have been changed and improved and number of unit tests were written to validate and support its operations. It still not ideal and can be done better, but current implementation is good enough.


abuts:
fixes Re #954

PR #1641, opened 2024-04-16
1D plots are plotted on a single image

### LINKED ISSUE #1619: plotting array in case of 1D objects does not behaves as expected
plotting arrays of 1D objects in Horace creates mutliple images and the additional images do not registered with Horace

Should plot array of 1D objects on single picture

abuts:
fixes #1619 
array of 1D object is plotted on single plot with line parameters choosen from specified variable. The old behaviour when you may support array of plot parameters is also supported. 

fixed separate plots in 1D case


abuts:
resolves #1619

PR #1645, opened 2024-04-23
[WIP] Dump parallel failure to log unless "debug" is "true"

oerc0122:


cmarooney-stfc (changes requested):
needs a test

herbert_core/classes/MPIFramework/@JobDispatcher/private/display_fail_jobs_.m
```diff
@@ -1,107 +1,143 @@
...
 
+end
 
+function debug_print(varargin)
+    if debug
+        fprintf(varargin{:});
+    else
+        fprintf(fh, varargin{:});
+    end
+end
```
abuts, line 143: completely agree with need for test. Absolutely not-obvious that `fh` is global variable and it propagated correctly.

The idea looks reasonable.

abuts (changes requested):
needs test

cmarooney-stfc (dismissed):

herbert_core/classes/MPIFramework/@JobDispatcher/private/display_fail_jobs_.m
```diff
@@ -1,107 +1,142 @@
...
-        elseif isfield(outputs{i},'error') && isa(outputs{i}.error,'MException')
-            mEXceptions_outputs(i) = true;
-            fprintf('Task N%d failed. Reason: %s\n',...
-                i,outputs{i}.fail_reason);
-            
+        if isa(outputs{i}, 'MException')
+            MExceptions_outputs(i) = true;
+            debug_print('Task %d failed. Error %s; Message %s\n', ...
+                    i, outputs{i}.identifier, outputs{i}.message);
+
```
abuts, line 46: I was playing with various failing jobs during the week and would argue that this should be reconsidered. 
When job fails, it prints short list of messages stating worker failed (one row)
Then it prints detailed information. 
The detailed information should go into log, that's correct. 
But short info about job failure should remain. Then it should be one string 
`see file .... for details of the issue`. 

I do not want to investigate this issue too deep, but have a feeling that modified changes are too brief. The suggested solution would provide much better user experience.

abuts (commented):
suggestion done below

Issue #1609, opened 2024-03-18
User Request: Condense parallel warnings


oerc0122:
Parallel failures dump every process' warning to the console. This is intimidating and unhelpful to users. Errors should be condensed helpfully or not printed if not in some kind of debug mode.

PR #1647, opened 2024-04-23
Add `warnonce` function to trigger a warning only once

oerc0122:
Related to #1609 

Does not use this function anywhere, but provides function for later PRs

### DIFF
--- a/_test/test_utilities/test_warnonce.m
+++ b/_test/test_utilities/test_warnonce.m
@@ -1,0 +1,66 @@
+classdef test_warnonce < TestCase
+
+    methods
+        function obj = test_warnonce(~)
+            obj@TestCase('test_warnonce');
+        end
+
+        function setUp(~)
+            warnonce('clear', 'HORACE:panic:panic')
+            warnonce('clear', 'HORACE:panic:arr')
+        end
+
+        function tearDown(~)
+            warnonce('clear', 'HORACE:panic:panic')
+            warnonce('clear', 'HORACE:panic:arr')
+        end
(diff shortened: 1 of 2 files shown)


_test/test_utilities/test_warnonce.m
```diff
@@ -0,0 +1,29 @@
...
+        function obj = test_warnonce(~)
+            obj@TestCase('test_warnonce');
+        end
+
+        function test_warnonce_warns_once_defaults(~)
+            out = evalc("warnonce('HORACE:panic:panic', 'Panic at the disco')");
+            msg = lastwarn;
+
+            assertEqual(msg, 'Panic at the disco');
+            assertTrue(contains(out, 'Panic at the disco'))
```
abuts, line 13: The test will fail if run on the same machine twice. 

This is pain for maintenance. 
Reset mechanism should be added for testing purposes.

abuts (changes requested):
idea ok, but tests would fail if run on the same machine twice. This should be avoided
simple reset 
(e.g. call with:
 `warnonce('-reset')` should be implemented for testing purposes. Or `warnonce('')` like `lastwarn` even better

_test/test_utilities/test_warnonce.m
oerc0122: Double check you're happy with the new mechanism and we can start putting this in places where it's used.

cmarooney-stfc (approved):
on the basis this is infrastructure only, needs propagating

PR #1652, opened 2024-04-26
Hashing for unique-objects-container to use external MD5 function

cmarooney-stfc:


_LowLevelCode/cpp/GetMD5/GetMD5.c
```diff
@@ -0,0 +1,919 @@
...
+  Method_t Method = BINARY_m;
+  mwSize   nByte;
+  
+  // Check number of inputs and outputs: ---------------------------------------
+  if (nrhs == 0 || nrhs > 3) {
+     ERROR("BadNInput", "1 to 3 inputs required.");
+  }
+  if (nlhs > 1) {
+     ERROR("BadNOutput", "Too many output arguments.");
+  }
```
abuts, line 812: All our mex files contain standard output, returned when mex is called with no inputs and single output. The `mex`-es all included in `check_horace_mex` which is called at initialization and verifies if mex files are available and compiled correctly. If they are not, mex gets disabled. 

I think this mex should follow standard practice.

abuts (changes requested):
It supporsetly should be completed implementation which uses mex/nomex md5 hasher. 

as such, it has to be build according to standard (described in the code below) and tested separately 
should be mex/nomex unit test and couple of tests validating hashes against standard values.

abuts (changes requested):
Well, almost there. 
GetMD5.c should be removed as it is not used any more -- this is requests

Others are suggestions:
1) I would question using class as the wrapper for hasher and the initialization of Java engine in case of mex is used. As I understand, it is initialized anyway, regardless of mex beeing used or not and it is unnecessary. 
2) You use persistent variable to hold java engine which means that class is not necessary -- everything can be placed within your hasher.
3) Though I am not sure we should keep build windows mex-es in the repository we still do. Some of them have been updated and new one have not been added. I think we should either add GetMD5.mex64 or remove all mexes.

_test/test_unique_objects_container/test_unique_objects.m
```diff
@@ -269,12 +269,13 @@ function test_add_different_types(obj)
         end
         %----------------------------------------------------------------
         function test_change_serializer(obj)
+            skipTest('not changing serialiser any more');
```
oerc0122, line 272: Not sure what you mean by this.

_test/test_unique_objects_container/thingy.m
```diff
@@ -1,4 +1,4 @@
-classdef thingy
+classdef thingy < serializable
```
oerc0122, line 1: Do these want to go in the general `common_functions` folder with more general names so we can use them globally in tests and keep the testing folder cleaner?

herbert_core/utilities/classes/@Hashing/Hashing.m
```diff
@@ -0,0 +1,76 @@
+classdef Hashing
+    %HASHING Independent class providing hash creation functions
```
oerc0122, line 2: Is it a class or a mixin?

herbert_core/utilities/classes/@serializable/private/to_bare_struct_.m
```diff
@@ -32,7 +32,11 @@
 for j = 1:numel(obj)
     obj_tmp = obj(j);   % get pointer to jth object to save expensive indexing
     for i = 1:numel(field_names)
+        try
         field_name = field_names{i};
+        catch
+            disp('')
+        end
```
oerc0122, line 39: Debugging statement?

horace_core/DLL/_PCWIN64/accumulate_cut_c.mexw64
oerc0122, line 1: Why are all these being changed?

herbert_core/utilities/classes/@Hashing/Hashing.m
```diff
@@ -0,0 +1,76 @@
...
+                % mex version to be used, use it
+                hash = GetMD5(bytestream);
+            else
+                % mex version not to be used, manually construct from the
+                % Java engine
+                Engine.update(bytestream);
+                hash0 = Engine.digest;
+                hash1 = typecast(hash0,'uint8');
+                %using the above typecast to remedy that dec2hex
+                %does not work with negative numbers before Matlab 2020b
```
oerc0122, line 68:
```suggestion
                %dec2hex does not work with negative numbers before Matlab 2020b
                hash1 = typecast(hash0,'uint8');
```

_LowLevelCode/cpp/GetMD5/GetMD5.c
cmarooney-stfc: thanks, should be done

_test/test_unique_objects_container/test_unique_objects.m
cmarooney-stfc: skip message modified

_test/test_unique_objects_container/thingy.m
cmarooney-stfc: The class looks pretty specific to its location to me. I have added a comment about why it is now subclassed

herbert_core/utilities/classes/@serializable/private/to_bare_struct_.m
cmarooney-stfc: deleted

herbert_core/utilities/classes/@Hashing/Hashing.m
cmarooney-stfc: saves the reader from trying to work out why this is an issue. In fact a bit more is needed and has been added

horace_core/DLL/_PCWIN64/accumulate_cut_c.mexw64
cmarooney-stfc: exceedingly good question. Why would be local compile be different from what is already in the repo? (Admitted that they should not have been pushed)

cmarooney-stfc:
Re Alex' comment
```
Well, almost there.
GetMD5.c should be removed as it is not used any more -- this is requests

Others are suggestions:

I would question using class as the wrapper for hasher and the initialization of Java engine in case of mex is used. As I understand, it is initialized anyway, regardless of mex beeing used or not and it is unnecessary.
You use persistent variable to hold java engine which means that class is not necessary -- everything can be placed within your hasher.
Though I am not sure we should keep build windows mex-es in the repository we still do. Some of them have been updated and new one have not been added. I think we should either add GetMD5.mex64 or remove all mexes.
```
GetMD5.c only kept until .cpp version known to work - now removed.
mex file now added but let's remove them all systematically at some point
engine initialization now made conditional on use_mex value
In an ideal world Matlab would have class variables and we would not need to use persistent here. However, if we are going to separate hashing from unique_objects_container, as previously requested, then I prefer to have it in a class from the start as it provides growing room if this needs further modification e.g. for uses outside the unique containers.

herbert_core/utilities/classes/@Hashing/Hashing.m
cmarooney-stfc: It is a class providing a static member. It is not intended presently that unique_objects_container derive from it. See unique_objects_container.hashify(...)

herbert_core/utilities/classes/@Hashing/Hashing.m
cmarooney-stfc: ok realised this was about the position of the comment but it takes out too much text - fixing with alternative text

_test/test_unique_objects_container/test_unique_objects.m
```diff
@@ -269,12 +269,14 @@ function test_add_different_types(obj)
         end
         %----------------------------------------------------------------
         function test_change_serializer(obj)
+            skipTest('serialiser function is now fixed and may not be changed');
```
oerc0122, line 272:
```suggestion
            skipTest('serialiser function is now constant and may not be changed');
```

oerc0122 (approved):

abuts (approved):

abuts:
I think it is only the beginning of the changes as you hasing serializable and this is unnecessary as fields for hash are the same. 
Hash should be build from values of the fields and be stored with object (`hashable`)

Issue #1656, opened 2024-05-07
upgrade_file_format incorrectly uses `-upgrade_range` option.


abuts:
`upgrade_file_format` have range of issues very inconvenient to users.

1) if you specify `-upgrade_range` option incorrectly, it silently ignores it. File format get upgraded but ranges are not calculated. 
2) if you call `upgrade_file_format` on recent version of sqw file with incorrect ranges, and the file is not misaligned, ranges are not recalculated. There are no curent way of upgrading ranges for this kind of file.

abuts:
fixed through Re #1661

Issue #1498, opened 2024-01-19
H4UX: update_file_format is not very user friendly


abuts:
I've tride to run `update_file_format` on test iron file and was not very happy. It looks like it works, but is slow and silent not reporting the progress for 2-3 minutes. It also looks like doing pix range calculations twice (well, it re-aligns data) but seems all this can be done better and more transparent

abuts:
Superseeded by Re #1656 and fixed by Re #1661

Issue #1195, opened 2023-08-19
More checks needed for upgrade_legacy_alignment


abuts:
Ticket #1182 brough `upgrade_file_format` routine which modifies old Horace sqw files into new sqw file format. It also adds routine `apply_alignment` which apply alignment to misaligned modern sqw file and make it aligned but without any alignment information. 
there is also routine `upgrade_legacy_alignment` which converts legacy alignment into modern alignment. 

For the user convenience, `upgrade_file_format` does all above working over all old format and aligned files converting all these into normal `sqw` files. 

The path in `upgrade_file_format`  when one has legacy aligned file and tries to upgrade it into modern (not-aligned) file needs more tests. It is indicated in `test_admin->test_upgrade_file_format:test_upgrade_legacy_alignment_with_lattice`

abuts:
Fixed via Re #1156 through RP #1661

PR #1661, opened 2024-05-07
upgrade file format

abuts:
This solves #1656. It looks like it also addresses Re #1498 and Re #1195


herbert_core/configuration/log_config.m
```diff
@@ -60,8 +60,8 @@
     end
 
     properties(Access=protected, Hidden=true)
-        info_log_print_time_ = 30 % normally info log should be printed
-        %                         % every 30 secodd
+        info_log_print_time_ = 300 % normally info log should be printed
```
oerc0122, line 63: Why? I understand that this comment wasn't added this PR, but why should the log be printed every X timeperiod?

horace_core/admin/upgrade_file_format.m
```diff
@@ -46,20 +41,24 @@
...
             if nargout > 0
-                sqw_list{i} = sqw(ld_new,'file_backed',true);
+                sqw_list{i} = finalize_alignment(ld_new);
+            else
+                finalize_alignment(ld_new);   % Will do nothing if the file is not aligned && ranges are valid
             end
-            ld_new.delete();
+        elseif nargout>0
+            sqw_list{i} = sqw(ld_new,'file_backed',true);
         end
```
oerc0122, line 60: Would prefer like elements grouped together
```suggestion
        if nargout > 0
            if upgrade_ranges %
                sqw_list{i} = finalize_alignment(ld_new);
            else
                sqw_list{i} = sqw(ld_new,'file_backed',true);
            end
        else
            finalize_alignment(ld_new);   % Will do nothing if the file is not aligned && ranges are valid
        end
```

horace_core/sqw/PixelData/@PixelDataBase/finalize_alignment.m
```diff
@@ -11,7 +11,7 @@
     outfile = '';
 end
 targ_obj = copy(obj);
-if ~targ_obj.is_misaligned
+if ~targ_obj.is_misaligned && targ_obj.is_range_valid
```
oerc0122, line 14: This seems to be a repeated condition with a defined meaning. Possibly want a getter for this combination with a clear name.

horace_core/sqw/PixelData/@pix_metadata/pix_metadata.m
```diff
@@ -58,6 +58,20 @@
...
+            % input is valid, i.e. not equal to empty range;
+            if nargin == 1 % check the whole range
+                idx = PixelDataBase.FIELD_INDEX_MAP('all');
+            elseif iscell(fld)
+                idx = cellfun(@(fl)PixelDataBase.FIELD_INDEX_MAP(fl),fld);
+            else
+                idx = PixelDataBase.FIELD_INDEX_MAP(fld);
+            end
+            invalid = obj.data_range_(:,idx) == PixelDataBase.EMPTY_RANGE(:,idx);
+            is = ~any(invalid(:));
```
oerc0122, line 72: Minor thing and subjective, but given the function is in the positive would probably find it easier to rationalise if this were in the positive too:
```suggestion
            valid = obj.data_range_(:,idx) ~= PixelDataBase.EMPTY_RANGE(:,idx);
            is = all(valid(:));
```

oerc0122 (changes requested):

_test/test_admin/test_upgrade_file_format.m
```diff
@@ -68,7 +65,9 @@ function test_upgrade_single_sqw_filebacked_upgrade_range_no_warning(obj)
             w2  = sqw(targ_f{1});
 
             [~,e] = lastwarn;
-            assertEqual(e,'TESTS:my_warning');
+            % second warning comes on windows as memmapfile is locked
+            % despite beeing deleted.
+            assertTrue(isequal(e,'TESTS:my_warning')||isequal(e,'MATLAB:DELETE:Permission'));
```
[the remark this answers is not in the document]
abuts: no masking -- it has been deleted but warning was issued.

_test/test_sqw_pixels/test_PixelAlignment.m
```diff
@@ -39,6 +39,28 @@ function test_save_filebacked_applies_alignment_on_sqw(~)
...
+            pdm = PixelDataFileBacked(pix_data);
+            
+            assertFalse(pdm.is_misaligned);
+            assertTrue(pdm.is_range_valid);
+
+            ref_range = pdm.data_range;
+            pdm.data_range = PixelDataBase.EMPTY_RANGE;
+            assertFalse(pdm.is_range_valid);
+
+            pdm = pdm.finalize_alignment();
```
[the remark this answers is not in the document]
abuts: well, yes, but I would not doing it now.

herbert_core/configuration/log_config.m
abuts: I was working with alignment and found this issue. This number controls log string (`printed N#out of #F`). Dots are printed with every page

horace_core/admin/upgrade_file_format.m
abuts: I am very reluctant to change order of operations here on a whim. The operations have costs there and in suggested case you will read metadata regardless of your intentions on aligning them. This change needs much more thinking than I am ready to spend on it

horace_core/sqw/PixelData/@PixelDataBase/finalize_alignment.m
abuts: they appear in various independent places. Its not so big and complex thing, to define new API for it. The code is much clearer this way.

PR #1684, opened 2024-05-10
changes to fix idaaas crash

abuts:
This contains various debugging changes and modifications done in attempt to fix IDAAAS kernel crash when running gen_sqw job.
Though the reason for crash was the issue in kernel, here I put bunch of convenience changes and fixes made during debugging, namely:
1) logging while running parallel framework. Previous code was botching FUSE file system. In addition, it is exteamly inconvenient to see and look at dots. User wants to get estimate when his parallel job completes and he wants this estimate ASAP, not in half an hour.
2) presumably faster pix writer
3) log while gen_sqw bulds rundata files. Before it was sitting telling user nothing for long time when sqw is build from number of nxspe files. 
4) Possibility to modify `sqw_format_factory` to have different default file writer for specified objects. Previous file accessors do not work entirely correctly but may be used successfully in number of differen situations. 
5) bunch of small formatting and convenience changes.

### DIFF
--- a/_test/test_sqw_file/test_faccess_sqw_v3_3.m
+++ b/_test/test_sqw_file/test_faccess_sqw_v3_3.m
@@ -201,6 +201,8 @@ classdef test_faccess_sqw_v3_3< TestCase
             so = faccess_sqw_v3_3(tf);
             so = so.reopen_to_write();
 
+            % file will be probably broken but actually faccessor recovers
+            % it from metadata (binary reader will not)
             so = so.put_num_pixels(10);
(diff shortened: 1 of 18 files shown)


oerc0122 (changes requested):
It would be very unwise to hotfix in so many changes. Split it into the bit which you think actually addresses the problem on IDAaaS and let's see if we can get that in ASAP.

abuts:
These changes are relatively urgent as reply to user request. I do not have time dividing them in pieces and discussing them separately.

Issue #1690, opened 2024-05-16
minor addition to head


abuts:
recently I have modified `head` so adding the description of the changes to doc may be useful

abuts:
Sorted through Re #1691

PR #1691, opened 2024-05-17
Small addition to description of `head` command

abuts:
Resolves Re #1690. Added example of `head` command output. 
Not very useful, important or pretty but why not.


documentation/user_docs/docs/images/Head_output_table.csv
```diff
@@ -0,0 +1,23 @@
+Scientific view:  ``head(file);`` ,Programmer view:  ``struc = head(file)``
```
cmarooney-stfc, line 1: Having looked at the file preview, it is very confusing as it stands. I keep thinking that there is a correlation between the items in each row, but there isn't. Could you break this into 2 tables, one with just the first column, one with just the second. Thanks

cmarooney-stfc (changes requested):

documentation/user_docs/docs/images/Head_output_table.csv
abuts: There is some correlation between rows but you are probably right about confusion

documentation/user_docs/docs/manual/Special_sqw_information.rst
```diff
@@ -48,6 +48,64 @@ an ``sqw`` or a ``dnd`` object. It also returns general information about the co
 this object, i.e. data ranges, number of pixels in ``sqw`` file etc., whatever developers decided 
 most important for user to know about correspondent object.
 
+Note, that output from calling ``head(filename);`` and ``res = head(filename)`` differs:
```
cmarooney-stfc, line 51:
```suggestion
Note, that the outputs from calling ``head(filename);`` and ``res = head(filename)`` differ:
```
sorry, minor pedantry

documentation/user_docs/docs/manual/Special_sqw_information.rst
```diff
@@ -48,6 +48,64 @@ an ``sqw`` or a ``dnd`` object. It also returns general information about the co
...
+
+>> ``head(filename);`` **"Scientific" view:**
```
cmarooney-stfc, line 53: I don't understand why this one is called scientific and the next one is called programmers. They are different uses of the function; one gives printed output to the screen, one gives a variable (a struct) with similar but not identical information. A better description of the two views and a reason why they come out that way would be useful

cmarooney-stfc (changes requested):

documentation/user_docs/docs/manual/Special_sqw_information.rst
abuts: To be fair, I would throw away "scientifc" output and would leave only the second one. But I am certainly looking into it as a programmers.

But historically, this is what was displayed by Horace/disp and this constrans the information useful for a scientist. (not sufficient for any other purposes) I can not explain why this selection was done and do not like it but as it was probably reason behind it, I am maintaining it.

cmarooney-stfc: ok if there are good historical reasons for this, let it stay

cmarooney-stfc (approved):

Issue #1703, opened 2024-06-03
build GetMD5.mex using horace script


abuts:
main build channel for creating mex files is cmake script, but for using it you need cmake and mpich libraries. 

If you need to build some mex files or not all prerequests for cmake build are met, for building mex files it may be convenient to deploy old fashioned `horace_mex` script. 

This ticket is to modify `horace_mex` script  to build new Horace mex file `GetMD5`

abuts:
fixed by Re #1704

PR #1704, opened 2024-06-03
Add possibility to build GetMD5.mex to horace_mex script

abuts:
simple solution for Re #1703. No unit test but it is unclear how to write them for admin tasks. 
Otherwise, verified manually and it works fine.


horace_core/admin/check_horace_mex.m
```diff
@@ -35,8 +35,8 @@
     'compute_pix_sums  : ', ...
     'combine_sqw       : ', ...
     'mtimesx_mex       : ', ...
-    'hdf_mex_reader    : ', ...
-    'GetMD5            : '  ...
+    'GetMD5            : ',  ...
```
cmarooney-stfc, line 38: Nothing wrong here but it is not clear why these lines have to be swapped. a comment saying why would clarify this for future developers. thanks

cmarooney-stfc (changes requested):

horace_core/admin/check_horace_mex.m
abuts: hdf mex reader is not used, does not work for current versions of Matlab and should probably be dropped.

I spent a lot of time on it but do not see it usage in any forceable future, so I am dropping it now.

abuts:
solves Re #1703

Issue #1415, opened 2023-11-29
spaghetti_plot on bin centres


oerc0122:
Request from Radu

Radu has a modified version of `spaghetti_plot` which always slices such that segment edges are calculated as bin-centres, but overlapping plots are used to conceal the other half of the bin, this means that when a new segment boundary is inserted into the middle of a segment, the resulting graph is identical rather than partially displaced.

PR #1711, opened 2024-06-19
Fix issue with 4.0.0 spaghetti plot

### LINKED ISSUE #1709: Issues with horace_demo_script
On Matlab R2024a I get the following error when running horace_demo_script:

```

*** Completed cat_pixels using 1 pages in 0.0095046 sec.
*** Step 1 of 1; Read data for   246124 pixels -- processing data... ----->  retained     97798 pixels
*** Completed cat_pixels using 1 pages in 0.0043894 sec.
*** Step 1 of 1; Read data for   324097 pixels -- processing data... ----->  retained    202794 pixels
*** Completed cat_pixels using 1 pages in 0.010904 sec.
Index exceeds the number of array elements. Index must not exceed 1.

Error in spaghetti_plot (line 398)
    out{2} = varargout{2};

Error in horace_demo_script (line 89)
wsp=spaghetti_plot(hklline+repmat([2 2 0],size(hklline,1),1),sqw_file,'labels',bzpts, ...


```

Further down in the script, an example fit is made. However, it calls a function '`fit_sqw`' which doesn't appear to exist

### LINKED ISSUE #1710: Spaghetti plot differs between 3&4
Vivi's user has found an issue between Horace 3 & Horace 4 Spaghetti plot with the rework. 

> Attached is the spaghetti plot script including one by one plots and plot-together. It looks the same (Fig.1).

 > The problem we met when using spaghetti_plot is that we got different plots from the normal horace plots:

download

oerc0122:
Issue was with which space vectors were computed in. 

- Fixed and simplified spaghetti plot using Horace 4 features. 
- One notable difference is that `qwidth` now specifies a full-width rather than a half-width confirmed by Duc to be the original intent, however, this means that `spaghetti_plot` `qwidths` in new Horace are half those of Horace 4. **N.B.** Can back this change out.

Fixes #1710 
Fixes #1709

### DIFF
--- a/documentation/user_docs/docs/user_guide/Horace3_to_4.rst
+++ b/documentation/user_docs/docs/user_guide/Horace3_to_4.rst
@@ -69,6 +69,13 @@ Passing a pair of vectors will still be accepted and be internally
 transformed into the correct ``Symop``, however, it is still best to
 switch.
 
+Spaghetti Plot
+--------------
+
+``qwidth`` in ``spaghetti_plot`` arguments now correctly refers to a full-width of the perpendicular
+bins. This means that to achieve the same results in Horace 4 as Horace 3, a user-specified
+``qwidth`` should be twice as large. Default ``qwidths`` have been updated to reflect this change.
+
 Crystal Alignment
 -----------------
 
--- a/horace_core/spaghetti_plot.m
+++ b/horace_core/spaghetti_plot.m
@@ -21,10 +21,10 @@ function varargout = spaghetti_plot(varargin)
 %   >> [wdisp,cuts] = spaghetti_plot(...)                    % generates a set of 1D cuts
 %   >> [wdisp, cuts] = spaghetti_plot(...,'withpix')         % return cuts as sqw rather than d1ds
 %   >> [wdisp,cuts,fig,axes,plots] = spaghetti_plot(...)     % in addition
-%                                    to  1D cuts retun references to the
+%                                    to 1D cuts return references to the
 %                                    plot objects i.e. figure handle, axes
-%                                    handle and array of hanles to each
-%                                    line plots displayed by spaghetti plot
+%                                    handle and array of handles to each
+%                                    line plot displayed by spaghetti plot
 % Input:
 % ------
 %   rlp             Array of r.l.p. e.g. [0,0,0; 0,0,1; 1,0,1; 1,0,0];
@@ -123,7 +123,7 @@ function varargout = spaghetti_plot(varargin)
 % Set defaults
 % ------------
 arglist = struct('qbin', 0.05, ...
-    'qwidth', 0.1, ...
+    'qwidth', 0.2, ...
     'ebin', [], ...
     'labels', '', ...
     'noplot', false, ...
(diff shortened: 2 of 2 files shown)


mducle (approved):
Looks good. Thanks for simplifying the code around line 306 and spotting the error in calculating the widths in line 337 / 312. 

Just to re-iterate `qwidth` was always meant to be the full-width but in the original implementation, the integration range `[v - qwidth, v + qwidth]` was used (line 337 in old code) which meant that `qwidth` was actually the half-width. Now the integration range is `[-qwidth, qwidth] ./ 2` (new line 313).

I'm happy with this breaking change, but @tgperring what do you think? Should the previous (half-width) behaviour be kept with just the documentation updated?

documentation/user_docs/docs/user_guide/Horace3_to_4.rst
```diff
@@ -69,6 +69,13 @@ Passing a pair of vectors will still be accepted and be internally
 transformed into the correct ``Symop``, however, it is still best to
 switch.
 
+Spaghetti Plot
+--------------
+
+``qwidth`` in ``spaghetti_plot`` arguments now correctly refers to a full-width of the perpendicular
+bins. This means that to achieve the same results in Horace 4 as Horace 3, a user-specified
+``qwidth`` should be twice as large. Default ``qwidths`` have been updated to reflect this change.
```
cmarooney-stfc, line 77: defaults where? For unspecified arguemnts? In documentation? etc

horace_core/spaghetti_plot.m
```diff
@@ -273,73 +276,51 @@
...
 
-bx_norm = sqw_proj.transform_hkl_to_pix([1;0;0]);
-bx_norm = bx_norm ./ norm(bx_norm);
-if nargout>1
+if return_cuts
     out_1d_cuts = cell(1,nseg);
 end
 
 for i = 1:nseg
+    % Choose u1 along the user desired q-direction
```
cmarooney-stfc, line 290:
```suggestion
    % Choose u1 along the user -desired q-direction
```

horace_core/spaghetti_plot.m
```diff
@@ -273,73 +276,51 @@
...
 
-    q_dir_abs = sqw_proj.transform_hkl_to_pix(q_dir_rlu')';
-    q_dir_abs = q_dir_abs ./ norm(q_dir_abs);
-
-    q_start_abs = sqw_proj.transform_hkl_to_pix(rlp(i, :)')';
-    q_end_abs = sqw_proj.transform_hkl_to_pix(rlp(i+1, :)')';
-
-
+    q_dir_cart = sqw_proj.transform_hkl_to_pix(q_dir_rlu')';
+    % Choose u2 to be either perpendicular to the plane of all the rlp (previous determined)
```
cmarooney-stfc, line 294:
```suggestion
    % Choose u2 to be either perpendicular to the plane of all the rlp (previously determined)
```

cmarooney-stfc (commented):
it would be nice to know what actually fixes the problem. a lot of renaming obscures the activ changes

abuts (approved):
As I understand, the selected implementation of the spaghetty plot follows the option 1

image

I think this should be included in description (e.g.:
"Note: the cuts performed between each reciprocal lattice points are ovelapped at the bin centers. See more details in HoraceHelp:spaghetty_plot (difficult to explain without image)".
If this is option 2 -- still needs to be specified
 Something like this.

documentation/user_docs/docs/user_guide/Horace3_to_4.rst
oerc0122: Defaults are usually for unspecified arguments, aren't they? I'm not sure what "defaults in documentation" might be and I'm not sure what clarification you want.

oerc0122, line 77:
```suggestion
``qwidth`` should be twice as large. Default ``qwidth``s (if unspecified) have been updated to reflect this change.
```
?

oerc0122:
> As I understand, the selected implementation of the spaghetty plot follows the option 1
> 
> image
> 
> I think this should be included in description (e.g.: "Note: the cuts performed between each reciprocal lattice points are ovelapped at the bin centers". If this is option 2 -- still needs to be specified Something like this.

This suggestion is addressed in issue #1415 however, a brief discussion with Adroja suggests that the behaviour they want/expect is what you have in diagram 1. 

At some point in the future we may implement #1415 as a boolean option (following this refactor it is now very simple to do).

Issue #1721, opened 2024-08-05
crystal alignment fails if parallel multifit is enabled


abuts:
as in Title, crystal alignment tries to deploy parallel multifit and it fails. 

The issue is twofold. 
1) crystal_alignment do not need parallel multifit. Its waste of time for crystal_alignemtn.
2) failure in splitting crystal alignment job to do parallel mutlifit. 

This ticket is to fix 1) and write unit test confirming 2). to fixing this unit test, ticket #1724 is written

abuts:
fixed by Re #1725

Issue #1724, opened 2024-08-05
Parallel multifit fails for some set of input parameters


abuts:
The ticket Re #1721 fixes `crystal_alignment` for case when parallel_multifit is enabled by disabling parallel mutlifit option in crystal alignment.  

This is probably fine for `crystal_alignment` as it does not need to run in parallel. But similar issue may occur for different multifit user cases so splitting procedure in multifit should be smarter to identify such parameters and either do correct split or (better) be smart enough to identify that parallel multifit in this case is unnecessary and multifit should not run in parallel. 

Unit test written and disabled in Re #1721: `Horace\_test\test_multifit_herbert\test_multifit_1.m`, row 362  should be fixed.

PR #1725, opened 2024-08-05
parallel in crystal alignment

abuts:
The PR fixes Re #1721 by disabling parallel multifit in crytal_alignment. It also adds the test, which demonstrates that parallel multifit fails for crytal_alignment case.


_test/test_multifit_herbert/test_multifit_1.m
```diff
@@ -324,5 +331,45 @@
...
+            kk = kk.set_fun (@reciprocal_space_deviation, {pars,rlu_expected});
+            kk = kk.set_options ('list', 0);
+            kk = kk.set_options ('fit',[1e-4,50,-1e-6]);
+
+            [distance,fitpar] = kk.fit();
+            assertEqualWithSave(obj,distance,'tol',1.e-4);
+            assertEqualWithSave(obj,fitpar,'tol',1.e-3);
+            clear clOb;
+
+            skipTest('Re #1724 This fit fails and should be fixed as part of the ticket #1724')
```
cmarooney-stfc, line 366: this suggests that the fix is going to be enabling parallel multifit. But the fix in this PR is to disable parallel multifit. At the least this message needs modifying

horace_core/lattice_functions/refine_crystal.m
```diff
@@ -167,9 +167,12 @@
 end
 
 kk = multifit (vcryst0_3, zeros(3*nv,1), 0.01*ones(3*nv,1));
+% function to caclulate distance between actual and expected bragg peak
+% positions in reciprocal space.
 kk = kk.set_fun (@reciprocal_space_deviation, {pars,rlu_expected}, pfree, pbind);
 kk = kk.set_options ('list', 0);
 kk = kk.set_options ('fit',[1e-4,50,-1e-6]);
+clOb = set_temporary_config_options('hpc_config','parallel_multifit',false);
```
cmarooney-stfc, line 175: this appears to be the only part of this PR that actually makes a change; the rest is refactoring and testing. There is no comment here to say - parallel multifit disabled here due to ticket 1724. The reader should be aware why this line is here and be able to trace it through the system. Please add the comment

cmarooney-stfc (changes requested):

horace_core/lattice_functions/refine_crystal.m
oerc0122: `set_temporary_config_options` is only available during testing (it's in `test/common_functions`. It cannot be used outside.

_test/test_multifit_herbert/test_multifit_1.m
abuts: parallel multifit is disabled in `crystal_alignment`. This test shows that parallel multifit fails for a task, similar to `crystal_aligmnent` but not actually it. You may see that no `crystal_alignment` is invoked here.

horace_core/lattice_functions/refine_crystal.m
oerc0122: Or you have moved it in this PR without saying that anywhere

_test/test_change_crystal/test_change_crystal_bragg_coarse.m
```diff
@@ -632,6 +634,36 @@ function test_rotmat_recovery(~)
 
             assertElementsAlmostEqual(corr.rotmat,rotmat_rec);
         end
+        function test_corrections_work_in_parallel(obj)
+            % testing if multifit works in parallel (actually no parallel 
+            % should be invoked) for this kind of task
```
oerc0122, line 639: I do not understand this comment which contradicts itself.

horace_core/lattice_functions/refine_crystal.m
abuts: I've moved it into `herbert/utilities`

_test/test_change_crystal/test_change_crystal_bragg_coarse.m
abuts: multifit should work regardless of parallel_multifit option is enabled or not. How it is achieved -- up to ticket Re #1724

oerc0122: The comment still doesn't make any sense to me.

abuts: modified

Issue #1723, opened 2024-08-05
Horace 4 - sliceomatic outputs full struct everytime you click something


mducle:
If you plot a 3D sqw object or d3d you get at `sliceomatic` figure in Horace 4 you see a stream of output in the command window which is not present in Horace 3.

This is because the code for the [sliceomatic function]( https://github.com/pace-neutrons/Horace/blob/master/herbert_core/graphics/sliceomatic/sliceomatic.m) was changed to return a structure `d` whereas the [original code](https://github.com/pace-neutrons/Herbert/blob/master/herbert_core/graphics/sliceomatic/sliceomatic.m) returned nothing. Note that outputs of GUI functions are not silenced so any outputs from them will be echoed.

So, `sliceomatic` should be (re)rewritten to **not** return any outputs.

abuts:
Fixed by Re #1726

PR #1726, opened 2024-08-09
remove persistiong output in sliceomatic

abuts:
Should fix  Re #1732 

Instantiate Horace\_test\test_dnd_class\test_plot_dnd.m  put breakpoint at row 97 and when code stops at it run:
`plot(d3d_obj)`

sliceomatic appears but each time you move or add view pannel, output is displayed on Matlab screen.

after applying this change, the output disappears


mducle (commented):
Code looks good and solves issue #1723

:shipit:

mducle (approved):

Issue #1727, opened 2024-08-09
Retrieve run number if it is stored in nxspe files


abuts:
Horace loader uses complex and unreliable procedure to identify run number. If nxspe file contains run number, this procedure becames substantially simpler. 

The purpose of this ticket is to complement the exisitng run number identification procedure with possibility to retrieve run number from nxspe file if this number is available there.

abuts:
fixed by Re #1728

PR #1728, opened 2024-08-13
runid in nxspe

abuts:
fixes Re #1727 

Looks like trivial change but I was not able to resist the temptation to refactor loaders a bit. There was even ticket for that but probably not worth doing it now. I've learn a lot from the times when I was writing this code so loaders can be substantially improved. This is a step in the direction of improvement. 

and run_id of course retrieved from the data files if it is present there.


herbert_core/classes/data_loaders/@a_loader/a_loader.m
```diff
@@ -34,13 +34,17 @@
         % all data should be loaded
         file_name
 
+        run_id % the number with identifies run on instrument
```
cmarooney-stfc, line 37:
```suggestion
        run_id % the number which identifies run on instrument
```

_test/test_data_loaders/test_loader_utilites.m
```diff
@@ -34,7 +34,7 @@ function test_FormatCurrentlyNotSupported(this)
         function test_CorrectHeader(this)
             [result,version] = find_root_nexus_dir(f_name(this,'MAP11014.nxspe'),'NXSPE'); % file name has loose relation to the result
             assertEqual(result,'/11014.spe');
-            assertEqual(version,'1.1');
+            assertEqual(version,1.1);
```
cmarooney-stfc, line 37: the potential for narrow mishits for floats may make it preferable to use the char version

_test/test_data_loaders/test_loader_nxspe.m
```diff
@@ -470,8 +478,20 @@ function test_loader_nxspe_defines(obj)
...
+        
+        function test_runid_from_filename(obj)
+            dat_file =f_name(obj,'MAP11014.nxspe');
+            lx = loader_nxspe(dat_file);
+            assertEqual(lx.run_id,11014)
+        end
+
         function test_load(obj)
-            dat_file =f_name(obj,'MAP11014v2.nxspe');
+            dat_file =f_name(obj,'MAP11014v3.nxspe');
```
cmarooney-stfc, line 494: here and in the other tests it would be useful to be able to change the filename at a single point, to save a lot of inspecting changes

cmarooney-stfc (changes requested):

_test/test_data_loaders/test_loader_utilites.m
abuts: From now on I have regular comparison for version beeing bigger than the specific number, so this change comes.  If Matlab works correctly, it should use the same `str2double` transformation while interpreting numbers typed in workspace and converting one, returned by hdf, so I do not expect equality being problem too.

Issue #1729, opened 2024-08-15
filebacked cut from file with nopix fails


abuts:
got this letter from user:

```
Dear Alex,

I'm having a new issue with using cut_sqw where it tries to write the cut to a tmp file in a completely different directory for some reason. I've been running the same code many times without issue until now. The cut happens inside the function PlotOnePanel. 
```
I had a check and indeed -- it fails as it actually does filebacked cut with pixels but then do not combine them together. 
need fixing

abuts:
This is invalid ticket caused by incorrect code relesed on idaaas. Nothing to fix -- correct code does not have an issue

PR #1730, opened 2024-08-15
filebacked cut with nopix fixture

abuts:
should fix Re #1729. 

Seems fixed the issue but have no idea how to write unit test for that. Thertainly bad design, but not the time to fix the design.


cmarooney-stfc (changes requested):
Hi Alex - this is a fix by changing data, so not something that we can review for correctness in Horace. The change appears to be a sign in the first column. Could you say why this situation arose, what will fix it, what we can detect in Horace to fail gracefully. Can you also attach the messages associated with the failure in the ticket, and the failure point in Horace. Thanks

abuts:
apparently the issue is more complicated that I have thought. 
Its not fixed yet

PR #1737, opened 2024-09-16
Fix SymopGeneral in cut_sym

### LINKED ISSUE #1735: USER REPORT: Issues with SymopGeneral in cut_sym
User reported issues with SymopGeneral attempting to call `in_irreducible` during a cut sym (which should not need it). 

image
image

oerc0122:
Add extra option to `trust` the `selection` of atoms found by cut to be in the cut range and not check whether it also lies in the irreducible zone. 

Adds test for SymopGeneral cuts.

Fixes #1735

### DIFF
--- a/_test/test_sym_op/test_cut_sym_cube.m
+++ b/_test/test_sym_op/test_cut_sym_cube.m
@@ -178,6 +178,30 @@ classdef test_cut_sym_cube < TestCase
 
         end
 
+        function test_cut_sym_general(obj)
+            data = sqw.generate_cube_sqw(10);
+
+            proj = line_proj([1 0 0], [0 1 0]);
+            ubin = {[0.5 1 4.5], [0.5 1 4.5], [-5 5], [-5 5]};
+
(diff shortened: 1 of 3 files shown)


abuts (commented):
Comment: The solution of the problem as currently done fine with me, but more general objection which may be the point of another ticket is:

1) What a benifits of supporting heterogeneous arrays for symop? 
If they are present, then `transform_pixels` and may be `in_irreducable` should be purely virtual, generic transformation defined as special function and called when appropriate as content of `transform_pixels` operation to avoid code duplication or overloaded appropriately. 
2) If there are no visible benifits of heterogeneous arrays, just use cellarrays and do simple overloading for both `transform_pixels` and `in_irreducable`.

Everything done here using `if` and additional parameters can be implemented using standard virtual function overloading -> much simple and clearer code.

cmarooney-stfc (approved):

cmarooney-stfc (approved):

PR #1751, opened 2024-09-24
Fix issues with calculate_qw_pixels2 for recent horace

### LINKED ISSUE #1152: Missing tests for `calculate_qw_pixels2`

oerc0122:
Update function to work with unique_references_container, struct-array header and use established functions. 

This will likely be used in the on-the-fly re-computation of pixels.

Fixes #1152


horace_core/sqw/@sqw/calculate_qw_pixels2.m
```diff
@@ -79,13 +84,15 @@
     kf = ki;
 end
 
-qw = cell(1, 4);
-qw(1:3) = calculate_q (ki, kf, detdcn(:, idet), spec_to_rlu(:, :, irun));
+qw = cell(4, 1);
+qw(1:3) = calculate_q(ki, kf, detdcn(:, idet), spec_to_rlu(:, :, irun));
 qw{4} = eps_diff;
+qw = cat(2, qw{:})';
```
cmarooney-stfc, line 90: be good to comment what these lines do as they look rather obscure

cmarooney-stfc (changes requested):

horace_core/sqw/@sqw/calculate_qw_pixels2.m
oerc0122:
```suggestion
% Join cell array into 4xN mat
qw = cat(2, qw{:})';
```

horace_core/sqw/@sqw/calculate_qw_pixels2.m
```diff
@@ -36,8 +36,11 @@
 idet = idx(2,:)';
 ien = idx(3,:)';
 
+remap = containers.Map(unique(irun), 1:numel(win.header));
+irun = arrayfun(@(x) remap(x), irun);
+
 if ~iscell(win.header)
-    header = {win.header};
+    header = num2cell(win.header)';
```
abuts, line 43: To be fair, we should not use `header` in new code. This is legacy method, necessary for writing old format `sqw` files and may be in some printing (also legacy, no time to throw it from there)

This is probably another ticket though.

horace_core/sqw/@sqw/calculate_qw_pixels2.m
```diff
@@ -79,13 +84,16 @@
...
+qw = cat(2, qw{:})';
+qw = win.data.proj.transform_hkl_to_pix(qw);
 
 end
 
-function [d_mat, detdcn] = spec_coords_to_det (detpar)
+function detdcn = spec_coords_to_det (detpar)
 % Matrix to convert coordinates in spectrometer (or laboratory) frame into detector frame
 %
 %   >> d_mat = spec_coords_to_det (detpar)
```
abuts, line 99: This is code duplication as something similar is in `rundatah/convert_to_sqw`. Which one is better I do not know, as see some optimization here too. Probably other ticket again.

abuts (commented):
Just comment:
solves the problem so no direct objections but quality of changes is low as opportunity to remove substantial and very important code duplication is missed.

horace_core/sqw/@sqw/calculate_qw_pixels2.m
oerc0122: A whole chunk of this including `calculate_qw_pixels2` is duplicated in `rundatah` (c.f. `rundatah.calc_qspec`) and the whole thing needs to be split off into utilities at some point, but like you say, not this ticket.

horace_core/sqw/@sqw/calculate_qw_pixels2.m
oerc0122: What is the new alternative?

abuts: `Experiment.IX_experiment` or ` sqw.experiment_info.expdata`

Issue #1753, opened 2024-09-27
horace_mex and issues with cpp_communicator


abuts:
users complain that loading Horace-4 from distribution and trying to install it on clean windows machine causes problem. `cpp_communicator` does not have all its references (dll-s) satisfied on the system. 

To at least partially address this:
1) make check for `cpp_communicator` an independent on other `check_horace_mex` checks to allow mex to be enabled regardless to mpi. 
2) add validation for cpp_communicator to attempt to enable mpi cluster  (`mpiexec_mpi`)
3) fix issues with `horace_mex` to allow users who is not familiar with `cmake` building libraries on their systems.

abuts:
fixed through #1754

PR #1752, opened 2024-09-26
bunch of minor issues and comments

abuts:
while working with Horace, I've identified bunch of minor annoying issues. 
These are the fixes for these issues.

PR #1754, opened 2024-09-27
fixing horace mex and disentangling various mex options.

abuts:
Fixed Re #1753.

These are small changes necessary for using simple mex and MPI mex independently. 

also fixed horace_mex to be able to help users without `cmake`

should be dealt with after Re #1752 is merged


abuts:
should be dealt with after Re #1752

horace_core/admin/check_horace_mex.m
```diff
@@ -1,17 +1,29 @@
-function [rez, n_errors, can_use_mex_4_combine] = check_horace_mex()
+function [rez, n_errors, can_use_mex_4_combine,can_use_MPI] = check_horace_mex()
 % function checks if horace mex files are compiled correctly and returns
 % their version string.
 %
 % Usage:
 %   >>[rez, n_errors] = check_horace_mex();
-%   >>[rez, n_errors, can_use_mex_4_combine] = check_horace_mex();
+%   >>[rez, n_errors, can_use_mex_4_combine,can_use_herbert_mpi] = check_horace_mex();
```
cmarooney-stfc, line 7: unclear why herbert is appearing here in the description

horace_core/configuration/@hor_config/hor_config.m
```diff
@@ -296,14 +296,20 @@
             if use
                 % Configure mex usage
                 % --------------------
-                [~, n_errors, can_combine_with_mex] = check_horace_mex();
+                [~, n_errors, can_combine_with_mex,can_use_herbert_mpi] = check_horace_mex();
```
cmarooney-stfc, line 299: as above unclear why herbert is appearing here

cmarooney-stfc (changes requested):
one query about mpi labelling as herbert

horace_core/admin/check_horace_mex.m
abuts: because parallel computing toolbox uses Matlab MPI and Herbert may use Poor man MPI  which is not an MPI but pretends to be such (writes/reads/deletes files pretending to be MPI messages). And may use `cpp_communicator` which supports proper MPI implementation compiled for Herbert

horace_core/configuration/@hor_config/hor_config.m
abuts: may be gave this variable better name

horace_core/admin/check_horace_mex.m
```diff
@@ -1,17 +1,29 @@
-function [rez, n_errors, can_use_mex_4_combine] = check_horace_mex()
+function [rez, n_errors, can_use_mex_4_combine,can_use_custom_MPI] = check_horace_mex()
```
oerc0122, line 1: What is `custom` here? What makes it `custom`? Do you mean system?

herbert_core/classes/MPIFramework/@parallel_config/parallel_config.m
```diff
@@ -491,7 +491,13 @@
         function obj = set.parallel_threads(obj,n_threads)
             n_threads = floor(n_threads);
             n_workers = get_or_restore_field(obj, 'parallel_workers_number');
-            n_poss_threads = floor(obj.n_cores/n_workers);
+            if ispc % let's assume pc is Intel so threades are double number of cores
```
oerc0122, line 494: What do you mean "assume it's intel"? Do you mean it has hyperthreading? You should be careful
a) because hyperthreading my be either captured in `n_cores` or it may not. Sometimes you have hyperthreading mapped as a logical (but not a physical) processor other times it may just be a hidden feature handled internally as in some ARM systems. 
b) Because hyperthreading may not always be advantageous to use if you're already saturating the processors with work and can result in slowdown.

herbert_core/classes/MPIFramework/@parallel_config/parallel_config.m
```diff
@@ -491,7 +491,13 @@
...
+                % this estimeat will not be too wrong even if it is AMD,
```
oerc0122, line 495:
```suggestion
            if ispc % let's assume pc is Intel so threads are double number of cores
                % this estimate will not be too wrong even if it is AMD,
```

admin/horace_mex_mpi.m
```diff
@@ -5,7 +5,7 @@
 % Manually modify this script to specify the mpi libraries location in your
 % system.
 %
-use_her_mpich = false;
+use_her_mpich = true;
```
oerc0122, line 8: Why are we changing defaults here?

horace_core/configuration/@hor_config/hor_config.m
```diff
@@ -296,14 +296,20 @@
...
                     warning('HOR_CONFIG:set_use_mex',...
                         ' mex files can not be initiated, Use mex set to false');
                 end
                 if ~can_combine_with_mex
-                    config_store.instance().store_config(obj,'combine_sqw_using','matlab');
+                    config_store.instance().store_config('hpc_config','combine_sqw_using','matlab');
+                end
+                if ~can_use_custom_mpi
+                    pc = parallel_config;
+                    if ismember(pc.parallel_cluster,{'mpiexec_mpi','slurm_mpi'})
```
oerc0122, line 310: Why if these two are present in `parallel_cluster` would it default to herbert? What is this `if` statement actually checking?

oerc0122 (changes requested):

horace_core/configuration/@hor_config/hor_config.m
abuts: because `~can_use_custom_mpi` means that `cpp_communicator` does not work. These two frameworks use `cpp_communicator` as the transport, so they are not avaliable if it does not work.

admin/horace_mex_mpi.m
abuts: because I was trying to build `cpp_communicator` using the library, we provide with `Herbert` and it worked. Probably it is a bit easier to build `cpp_communicator`  with the libraties we provide with Herbert, than install external MPI on a system and then build `cpp_communicator` (`use_her_mpic=false` -- look for MPI installed on a system)

There are actually little difference between these two options as compiling `cpp_communicator` is a simpler task than making it to work, which I will try to explain in a fresh ticket.

horace_core/admin/check_horace_mex.m
abuts: custom means created by us, not provided by Matlab.

cmarooney-stfc (commented):
Agreeing with Jacob that the issues he comments on need explaining. Not adding any changes required myself.

I think this needs a more comprehensive explanation going back to first principles; the changes and comments assume the reader knows something about these issues, which may not be the case. Please work out how to tell this story in the code.

admin/horace_mex_mpi.m
cmarooney-stfc: As from Slack earlier: two things. One, if it needs that much explanation, then it should be documented in the code. Please add a substantial comment saying what is going on there. Two, I don't understand your explanation. It sounds more as if there was a temporary glitch in the build process rather than something fundamental that requires this change. Please can you write a paragraph for comments and run them by us. If need be we can use the sprint review slot tomorrow to discuss it.

abuts: There are generally not advisable to describe in words the thing the code is doing. You should comment on meanings and describe intentions. This requests violates this principle, but it is not difficult to add what the code is doing adding couple of sentences.

Issue #1758, opened 2024-10-07
horace_install to install spinW


abuts:
Current Horace Install modifies `horace_on.template` and sets up path-es to Horace. `horace_on` script also contains path to spinW but this path have to be modified manually. 

as we now enable Horace-spinW integration tests  which may use `horace_install` to enable Horace, horace_install should also set up `spinW` path if such is provided.

abuts:
fixed by PR #1759

PR #1759, opened 2024-10-07
horace install accepts spinW path

abuts:
fixes Re #1758 

`horace_install` accepts path to `spinW` package and if such  option is available, modifies `horace_on` to include path to spinW.

Modified unit tests for `horace_install` to remove common code and test new option.


cmarooney-stfc (approved):
if the test works, this is fine

Issue #1762, opened 2024-10-21
horace_demo_script is not working due to reference to fit_sqw


henrikjacobsenfys:
This is issue is reported earlier in #1289 which is closed, but it's still present. I suggest changing the line

```
[wfit,fitdata]=fit_sqw(cc2a,@demo_FM_spinwaves,[250 0 2.4 10 5],[1 0 1 0 1],...
%     @constant_background,[0.05],[1],'list',2,'fit',[0.001 30 0.001]);
```

to
```
kk = multifit_sqw (cc2a);
kk = kk.set_fun (@demo_FM_spinwaves);

kk = kk.set_pin ([250 0 2.4 10 5]); %input parameters
kk = kk.set_free ([1 0 1 0 1]); %fitting parameters 1, 3 and 5

kk = kk.set_bfun (@constant_background); % set_bfun sets the background functions
kk = kk.set_bpin ([0.05]);   % initial background constant 
kk = kk.set_bfree ([1]);    % fix the background
[wout pout]=kk.fit();
```

abuts:
Dear @henrikjacobsenfys. Where did you find that? As far as I can see on IDAAAS, the suggested replacement have already been done.

Regards,
Alex

henrikjacobsenfys:
Hi @abuts 

Sorry for not being clear about this. It's in the Master branch of the repository: https://github.com/pace-neutrons/Horace/blob/master/demo/horace_demo_script.m 

Henrik

abuts:
fixed by Re #1763

PR #1763, opened 2024-10-22
removed old fit_sqw from horace demo

abuts:
fixes Re #1762. 
Old `fit_sqw` interface have been removed from Horace, but left in demo sctipt and now fails. This PR fixes this issue (noticed by user).

There is the question why not to support old `fit_sqw` interface -- it is ugly but works in old scripts. 

There is the question about spinW-Horace comparison as current spinW does not reproduce result obtained from Horace, but formaly demo script is fixed and spinW-Horace comparison is the question for another ticket.
The comparison question redirected to spinW:
 https://github.com/SpinW/spinw/issues/207

### DIFF
--- a/admin/horace_install.m
+++ b/admin/horace_install.m
@@ -19,11 +19,11 @@ function [init_folder,hor_init_dir,use_old_init_path,modified_hor_on_contents] =
 %                  to be installed. If missing,
 %                  <path to Horace code>/../ISIS folder is selected.
 %                  This folder will be added to Matlab search path.
-% spinW_folder --  key with value, which shows where spinw is located. 
-%                  Allows to initalize horace together with spin_w
+% spinW_folder --  key with value, which shows where spinW is located. 
+%                  Allows to initialize Horace together with spin_w
 %                  For this option to work, spinw_on.m.template file have
-%                  to be available in Horace and spinw package installedn and 
-%                  be avaiable on filesystem.
+%                  to be available in Horace and spinW package installed and 
+%                  be available on filesystem.
 %
 % test_mode   -- if true, do not install Horace but return installation
 %              folders, i.e. the folder where Horace and horace_on
@@ -33,7 +33,7 @@ function [init_folder,hor_init_dir,use_old_init_path,modified_hor_on_contents] =
 % Defaults (no arguments)
 %  the horace_install script is located either in the folder, where Horace
 %  and  Herbert folders are extracted to (installation archive) or in
-%  /Horace/admin folder (cloned from the Github directly)
+%  /Horace/admin folder (cloned from the GitHub directly)
 %
 % Output parameters:
 %  Expected to be used in test mode only.
--- a/demo/demo_FM_spinwaves.m
+++ b/demo/demo_FM_spinwaves.m
@@ -27,8 +27,9 @@ warning off
 js=pars(1); delta=pars(2);
 gam=pars(3); temp=pars(4); amp=pars(5);
 
-omega0 = delta + ...
-    (4*js)*((sin(pi.*qh)).^2 + (sin(pi.*qk)).^2 + (sin(pi.*ql)).^2);
+%omega0 = delta + ...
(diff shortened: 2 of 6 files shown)


demo/horace_demo_script.m
```diff
@@ -176,7 +204,14 @@
 wadd=cc2a+27;
 plot(wadd);%notice the changed limits of the colour scale!
 
-
+for i=1:numel(spe_file)
```
cmarooney-stfc, line 207: nice just to have comment "delete all temporary files" here

demo/horace_demo_script.m
```diff
@@ -9,10 +9,16 @@
 %====================================
 %% Generate Horace data files
 %====================================
+demo_dir=pwd;
+indir=demo_dir;     % source directory of spe (or nxspe) files
+par_file=[indir,filesep,'4to1_124.par'];     % detector parameter file
+sqw_file=[indir,filesep,'fe_demo.sqw'];        % output sqw file
+data_source =sqw_file;
```
cmarooney-stfc, line 16: you may want to create temporary file sqw_file in the cwd but par file is definitely in the demo directory so either give them separate directories or ensure all ops take place in the demo directory.

Also comment that sqw_file is designed to be temporary and will be deleted at the end

demo/horace_demo_script.m
```diff
@@ -122,45 +124,71 @@
...
-
+J = 250;     % Exchange interaction in meV
+D = 0;      % Single-ion anisotropy in meV
+gam = 0.5;   % Intrinsic linewidth in meV (inversely proportional to excitation lifetime)
+temp = 10;  % Sample measurement temperature in Kelvin
+amp = 5;  % Magnitude of the intensity of the excitation (arbitrary units)
+
+%[wfit,fitdata]=fit_sqw(cc2a,@demo_FM_spinwaves,[250 0 2.4 10 5],[1 0 1 0 1],...
+%    @constant_background,[0.05],[1],'list',2,'fit',[0.001 30 0.001]);
+kk = multifit_sqw (cc2a);
```
cmarooney-stfc, line 142: It would be better to remove the commented out fit_sqw line so it is clear that the new parameters J...amp go with the multifit ops below along with the Do a fit commnet

demo/setup_demo_data.m
```diff
@@ -1,12 +1,12 @@
-function file_list=setup_demo_data()
+function file_list=setup_demo_data(sqw_ready)
```
cmarooney-stfc, line 1: need comment defining the input parm sqw_ready

demo/demo_FM_spinwaves.m
```diff
@@ -27,8 +27,9 @@
 js=pars(1); delta=pars(2);
 gam=pars(3); temp=pars(4); amp=pars(5);
 
-omega0 = delta + ...
-    (4*js)*((sin(pi.*qh)).^2 + (sin(pi.*qk)).^2 + (sin(pi.*ql)).^2);
+%omega0 = delta + ...
+%    (4*js)*((sin(pi.*qh)).^2 + (sin(pi.*qk)).^2 + (sin(pi.*ql)).^2);
+omega0 = delta + (8*js)*(1-cos(pi.*qh).*cos(pi.*qk).*cos(pi.*ql));
```
cmarooney-stfc, line 32: you have left the old omega0 formulation commented out in the code. Unclear if your new formulation is just  a condensed rewrite of the old or a correction. Prefer you comment on what the change means

admin/horace_install.m
cmarooney-stfc, line 1: Comment for the whole PR. It is not clear from the PR description that the fit_sqw function had previously been removed before this PR - if that is what happened. Please make that clear.

Also if you feel that fit_sqw should be restored - you my well be right - please ticket it and give its location so it is easy to restore.

Also if spinW does not give a correct fit to the data, even with multifit search, then please ticket it, with details o the difference between the calculations, and give the ticket number here

cmarooney-stfc (changes requested):

PR #1714, opened 2024-07-01
enable new nxspe files to be read wthout error

cmarooney-stfc:
Draft PR (still debug info in some source) to test reading of these new files with the existing nxspe reader. the changes test the ability to read. 

The code to populate instruments with the additional information has not yet been added

PR #1765, opened 2024-10-25
Enable nxspe reader for instrument data

cmarooney-stfc:
Draft PR (still debug info in some source) to test reading of these new files with the existing nxspe reader. the changes test the ability to read.

The code to populate instruments with the additional information has not yet been added

This supersedes PR #1714 which is corrupt because it still contains references to the too-large original nxspe test files. The ones here have had their energy bins reduced in size from 640 to 10 to fit within the github limit. The python script used to shrink the files has also been added


abuts (commented):
Comment: 
I do not understand the meaning of these changes. 
Either you want to read whole instrument from nxspe -- then `read_instrument` method should be present on nxspe loader, and a dumb `read_instrument` returning empty instrument on other loaders. and `read_instrument` shoild return whole insturument. 

Or you may want to read part of the instrument -- then bunch of the methods for components should be present with dumb equivalents on the interface. 

This is some unclear mixture which reads instrument but uses other fake `"read"` methods to construct instrument. The purpose of such excersize is absolutely unclear to me.

PR #1773, opened 2024-11-13
changes to crystal alighnment

abuts:
1) Alignment operations made commutative (or rather not commutable but see the comment at the end), and unit tests to prove this.
2) Minor  bug in `refine_crystal_strip_pars`
3) formal changes to names of some functions and methods. 

To reviewer:
To see actual changes, it would be reasonable to walk commit-by commit. There are few actual changes. There are a lot of formal changes related to function/methods renaming


_test/test_change_crystal/test_change_crystal_bragg_coarse.m
```diff
@@ -634,11 +705,12 @@ function test_rotmat_recovery(~)
 
             assertElementsAlmostEqual(corr.rotmat,rotmat_rec);
         end
+
         function test_corrections_work_in_parallel(obj)
             % testing if multifit works when parallel_miltifit option is true
-            % At the momentm no parallel is invoked in refine_crystal, 
+            % At the momentm no parallel is invoked in refine_crystal,
```
cmarooney-stfc, line 771: mometm --> moment

horace_core/sqw/@sqw/refine_crystal_strip_pars.m
```diff
@@ -34,10 +35,12 @@
...
 
-% Compute rotation matrix and new ub matrix
-rotmat = rotvec_to_rotmat2(rotvec);
-ub = ubmatrix(xtal.urot,xtal.vrot,bmatrix(alatt,angdeg));
-rlu_corr = ub\rotmat*xtal.ub0;
+% Compute rotation matrix and new ub matrix : unclear meaning of this code.
+% it seems the replacemant below 
+%rotmat = rotvec_to_rotmat2(rotvec);
+%ub = ubmatrix(xtal.urot,xtal.vrot,bmatrix(alatt,angdeg));
+%rlu_corr = ub\rotmat*xtal.ub0;
```
cmarooney-stfc, line 42: this makes the user even more uncertain. The sentence on L 39 is incomplete. Can we drop all the new comments and just let people look at crystal_alignment_info?

horace_core/sqw/PixelData/@pix_metadata/private/set_alignment_matr_.m
```diff
@@ -1,5 +1,8 @@
 function  obj = set_alignment_matr_(obj,val)
 %SET_ALIGNMENT_MATR_ helper property which checks and sets alignment matrix
+% if alignment matrix differs from unary matrix, it is stored in the class
```
cmarooney-stfc, line 3: unary --> unit

this is not quite correct; alignment_matr_ is always set below. What is added is a min threshold for the difference, otherwise the unit matrix is stored. There is no attempt to assume that e.g. alignment_matr_==[] will be expanded to eye on the fly

_test/test_change_crystal/test_crystal_alignment_info.m
cmarooney-stfc, line 1: I am worried that the PR says alignment ops are commutative, as non-infintesimal rotations are not commutative. An explicit comment on what the commutative property is and why it works despite the restriction on rotations would be welcome

abuts: yes, they are not actually commuttable, rather I can split an operation in halfs(dividing rotvec by 2) and it will be equivalent to doing it in one go. Or I can align crystal, then try to align it again and it will be small alignment but I can apply it to crystal again and the crystal will remain(better) aligned. I do not know how to call this operation, but this was not happening before (bug)

cmarooney-stfc: thanks Alex that makes more sense. And multiple rotations about the same axis are commutable, it's when you change axes. The important thing is that the code is well documented to say what is happening and any assumptions

horace_core/sqw/@sqw/refine_crystal_strip_pars.m
abuts: finshed this comment. Hopefully it is coherent now

PR #1775, opened 2024-11-19
bug in alignment

abuts:
While working with alignment code, I have identified another bug there.  There was a way to change pixels alignment directly in file, which has not been modified similarly to the changes for filebacker/memorybased `sqw` objects.
In other words, `change_crystal` filebacked operation with metadata was missing multiplication by new alignment matrix, added to operations with `sqw` objects earlier

This PR adds unit test demonstrating the bug and fixture for it.

### DIFF
--- a/_test/test_change_crystal/test_change_crystal_bragg_coarse.m
+++ b/_test/test_change_crystal/test_change_crystal_bragg_coarse.m
@@ -273,9 +273,7 @@ classdef test_change_crystal_bragg_coarse < TestCaseWithSave
         end
         %------------------------------------------------------------------
         %------------------------------------------------------------------
-        function test_revert_alignment_works_on_file_and_on_filebacked(obj)
-            corr  = crystal_alignment_info([5.0191 4.9903 5.0121], ...
-                [90.1793 90.9652 89.9250], [-0.0530 0.0519 0.0345]);
+        function test_revert_alignment_same_on_file_and_on_filebacked_object(obj)
             test_file_fb = build_tmp_file_name(obj.misaligned_sqw_file);
             copyfile(obj.misaligned_sqw_file, test_file_fb, 'f');
             test_file_file = fullfile(tmp_dir,'misaligned_file_to_check_reversion.sqw');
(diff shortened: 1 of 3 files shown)


horace_core/lattice_functions/bragg_positions.m
```diff
@@ -171,6 +170,7 @@
         bin_t=trans_bin_width;
         thick_t=trans_thickness;
     else
+        modQ=norm(proj.transform_hkl_to_pix(Qrlu(:)));   % length of Q vector in Ang^-1
```
cmarooney-stfc, line 173: need for clarification rather than flagging an issue. The move of old line 142 to 173 could be due to (a) calculate modQ only for efficiency reasons; do not evaluate if absolute binning OR (b) ensure that proj here picks up the value set on line 157 rather than pick up the value from line 131 for i==1 OR (c) both. 

Could we delete line 131, use img.proj rather than proj in lines 132-133, and fix the comments at 145-146 so that u and v are appropriately described. Also add comment into the PR description saying what has been fixed

_test/test_change_crystal/test_change_crystal_bragg_coarse.m
```diff
@@ -273,6 +273,46 @@ function test_finalize_alignment_on_file_keep(obj)
...
+            test_fb = sqw(test_file_fb,'file_backed',true);
+            assertTrue(test_fb.is_filebacked, ...
+                'This test validates filebacked operations but the object test_fb placed in memory')
+            test_fb = test_fb.set_as_tmp_obj();
+            assertTrue(test_fb.is_tmp_obj)
+
+
+            % TEST:
+            % apply alignment on FB object
+            test_fb_ref = change_crystal(test_fb, corr);
```
cmarooney-stfc, line 295: be nice if line 277 were placed just before this one so we can see quickly what corr means

_test/test_change_crystal/test_change_crystal_bragg_coarse.m
cmarooney-stfc, line 1: PR description says: "While working with alignment code, I have identified another bug there.

This PR adds unit test demonstrating the bug and fixture for it". I cannot see where this test demonstrates the bug. That should be detection of  failure resulting from it. The bits of the test should be commented to say what they do with respect to this bug fix

It would be good if the PR description actually says what the bug is so we know which bits of the changes do what.

cmarooney-stfc (changes requested):

_test/test_change_crystal/test_change_crystal_bragg_coarse.m
abuts: before the fixture, the test was demonstrating the bug. After the fixture, it shows that code works.

horace_core/lattice_functions/bragg_positions.m
abuts: Despite the changes I did are probably more correct than the previous one(picking up homoheneous ) , I better restore previous behaviour as do not have sufficient experience in using relative units.

Issue #1791, opened 2024-12-16
Problems with  unique references singleton implementation


abuts:
**ISSUE DESCRIPTION**

This ticket addresses two issues: Design flaw and bug. Bug can be actually sorted differently, but still mentioned here.

Simple demonstration illustrates this:

1) fix `list` option of the container to be able to explore  its contents, namely, add return value: `glc = glcontainer`; to row 862 of the `global_references_container` code base to be able to explore its contents.
2) clear memory (`clear classes`, `clear all`)
3) open `_test\test_miltifit\test_multifit_horace_1` and set up break point at row 32, immediatly after Horace loads reference cuts.
4) Run code and stop at break point
5) Do:
```
>>glc  = unique_references_container.global_container('list','GLOBAL_NAME_DETECTORS_CONTAINER')
    GLOBAL_NAME_INSTRUMENTS_CONTAINER: [1×1 unique_objects_container]
      GLOBAL_NAME_DETECTORS_CONTAINER: [1×1 unique_objects_container]
        GLOBAL_NAME_SAMPLES_CONTAINER: [1×1 unique_objects_container]


glc = 

  struct with fields:

    GLOBAL_NAME_INSTRUMENTS_CONTAINER: [1×1 unique_objects_container]
      GLOBAL_NAME_DETECTORS_CONTAINER: [1×1 unique_objects_container]
        GLOBAL_NAME_SAMPLES_CONTAINER: [1×1 unique_objects_container]

K>> glc.GLOBAL_NAME_DETECTORS_CONTAINER

ans = 

  unique_objects_container with properties:

         n_objects: 24
            n_runs: 24
          n_unique: 24
         baseclass: 'IX_detector_array'
               idx: [1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24]
    unique_objects: {1×24 cell}
      n_duplicates: [1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1]

K>> glc.GLOBAL_NAME_DETECTORS_CONTAINER.unique_objects

ans =

  1×24 cell array

  Columns 1 through 4

    {1×1 IX_detector_array}    {1×1 IX_detector_array}    {1×1 IX_detector_array}    {1×1 IX_detector_array}

  Columns 5 through 8

[...]

Issue #1147, opened 2023-07-25
Serializable equality method (eq) should be replaced by overloaded equal_to_tol


tgperring:
Should be part of common equal_to_tol refactoring for `PixelData`
also includes  is_cut_equal

Serializable has a method to check for equality, eq.m, that enables comparisons overloading == so that A==B is possible. However, by default it permits a tolerance of 1e-9 (relative and absolute) on numerical data. This has the potential to result in unexpected behaviour.

In fact, the use of eq.m for serializable objects at the moment appears only to be in tests, and the function has multiple optional input arguments and output to control tolerance, and return an error message, in contrast to the Matlab intrinsic function. Accordingly, eq.m is really just performing the function of an overloaded version of equal_to_tol. This should replace the current eq.m, and if it is desirable to overload the operators eq, ne etc they should have the precise correspondence with the Matlab intrinsic operation e.g. eq.m shoudl test true equality.

abuts:
Addressed by PR #1793. 
`eq_to_tol` overloads made consistent with the structure of Horace objects including PixelDataBase.
Made it easy to overload `eq_to_tol` to any other object.

PR #1800, opened 2025-01-01
unique ref container rebranded

abuts:
Addesses Re #1791 
Also the solution addresses  majority of Re #1157 and remaining issues have been moved to Re #1816

1) Unique storage extracted from `unique_references_container` and placed into its own class-standard signleton `unique_obj_store`.

2) `unique_objects_container`  was morphied into `unique_only_obj_container`  to do reference couting and be ready for unused objects deleteon. `unique_references_container` works with `unique_obj_store` which provides underlying storage to it.  Storage is based on `unique_only_obj_container`  and is ready for implementing ref-conting (see Re #1816 for details and problems associated with this)
Multiple additions of the object, equal to one in storage increases reference counters but deleteon is disabled for the time beeing.

3) `unique_objects_container` itself remain unchanged as has its benifits and its own area of applicability. The only change to it (only but the big one) is to take common for all containers code and put it into `ObjContainersBase`.  Algorithm for memory managment used in `unique_only_obj_container` may be deployed in `unique_objects_container` later if necessary, which may be nice but not important or udgent addition.

4) issue with multiple copies of the same objects in memory have been fixed earlier and already in master and this PR uses existing unit tests for the issue to validate it would not occur any more. Manual validation can be performed as follows:

```
clear all
>> unique_obj_store.instance()

ans = 

  unique_obj_store with properties:

              n_types: 0
            typenames: {1×0 cell}
    n_unique_per_type: [1×0 double]

>> runtests test_multifit_horace_1
35              data_dir = fileparts(mfilename('fullpath'));
K>> unique_obj_store.instance()

ans = 

  unique_obj_store with properties:

              n_types: 3
            typenames: {'IX_detector_array'  'IX_inst'  'IX_samp'}
    n_unique_per_type: [2 2 2]
K>> unique_obj_store.instance().get_objects('IX_detector_array')

ans = 

  unique_objects_container with properties:

         n_objects: 2
          n_unique: 2
         baseclass: 'IX_detector_array'
               idx: [1 2]
    unique_objects: {[1×1 IX_detector_array]  [1×1 IX_detector_array]}
      n_duplicates: [46 10]
```

4) As the results, tests failing due to luck of memory should be passing.

Remaining is the deletion of non-used unique objects from memory. 

Simplified:
To simplify the object design, unique object storage tags are selected to be equal to the names of the stored classes. This disabled the option of creating multiple unique storage instances with the same classes. I do not think this option is useful, so disabling it was partially design choice.


cmarooney-stfc (changes requested):
No specific changes at this point but this is a massive change and the original problem is easily fixed. We will discuss this at the Monday standup and the sprint review this week. In the meantime this is not to be merged

oerc0122 (changes requested):
Do not approve this going in.

Partial review concerns:
- What/why is `unique_only_obj_container`? What is its relation to new `unique_obj_store`, `unique_objects_container`, `unique_references_container` the whole hierarchy is very confusing. 

`unique_only_obj_container`, unlike `unique_objects_container`  contains unique objects and unique references to objects, while `unique_objects_container` contains unique objects and non-unique references. I hope `only` word fully highlights the difference, but open to other name suggestions.

The hirachy, differences and the reason for having it has been explained on two meetings dedicated to the issue and recorded in PR #1803. Extensive explanation and meeting so it is impossible to have it here. 

- There is a big risk with manually having to add `obj.clear_cache` everywhere, it seems like the sensible thing to do would be to override `subsasgn('.', ...)`, check if the data are in `hashableFields`
 
There is such a risk, which is mitigated by writing unit tests for each class, using this feature to validate that hash is actually cleared after changing any hashable property.  Such test is mandatory for every new `hashable` class. From other side, adding `clear_hash` to each property would allow one to use existing optimization of `serializable` property setters, not to clear hash multiple times if multiple properties are changed. 

BTW, such tests indeed identified number of omissions and allowed to identify and fix number of additional (independent) issues with `hashable` classes.


-  (still no idea why that's lowerCamel rather than snake_case like the rest of the code) and clear the cache when it is.
lowerCamel `hashableFields` name selected as similar to `saveableFields` in `serializable` as these classes are similar and the functions have similar purpose.

- Why test hash in every object test suite? If object is a valid `serializable` it should have a valid `hashable` or a separate test suite testing hashability of various classes. Otherwise each test suite is really testing `hashable` in addition to the classes in question.

Each hahsble objects got the test for correct usage of `clear_hash` property mentioned above. Nothing else is added to the tests.

- Having `eq` as the method in `@hashable` still violates Toby's request in #1147 which has been closed without addressing the problem.

Also subject of extensive discussion on the meetings with comments pro and contra and resolution for further development. 

- Seemingly random change in `set_mod_pulse` to skip run `119` seems dangerous and incorrect.

This is intentional test of `set_mod_pulse` which have been there before. It has deep meaning and intention, not addressed here but used and vigorously tested exactly because it is dangerous and have to be performed correctly. It test separate situation and branch of the code.

- Unused line added in `find_set_mode`
line removed 

Changes unrelated to this PR, must be split off or removed:

Thsese changes are actually from other PR which is already in master. It have been reviewed 

- Changes to `multifit_lsqr_ser`
- Changes to `equal_to_tol_shape`
- Changes to `resolution_plot`

_test/CMakeLists.txt
cmarooney-stfc, line 1: This comment is for the whole PR.

I do not understand the hash storage being used in this PR. It has examples of both build_hash on individual objects and stored_hashes in the container. If both are retained, and explanation of this must be given in the PR description.

abuts: Container is agnostic to `build_hash` function and the ways, you obtain hashes. It uses `build_hash` for obtaining a hash and if object is `hashable`, you get benifit of calculating hash once. If not, ok, you loose performance but you stil can put object in the container and use its hashes for comparison. 

The objects we place to container for experiment are `hashable` so this give necessary performance. 
Tests are often using some rubbish like `double` or `char` but performance is not important in these situations.

May be we should make containers agnostic to hashes too, and just use generic comparison, but because we can not properly overload `find` or `ismember` with custom comparison operations, this would be less efficient (C++ containers would certainly be implemented without separate hash storage, but here we should follow Matlab contstains)

It this answer what you are looking for?

PR #1801, opened 2025-01-06
fix fetch of glc

cmarooney-stfc:
Responding to #1791 this fixes unique_references_container for the existing implementation

A handle was used as a reference handle although it was CoW. Handle now refetched after every update.

### DIFF
--- a/_test/test_unique_objects_container/test_unique_references.m
+++ b/_test/test_unique_objects_container/test_unique_references.m
@@ -28,7 +28,81 @@ classdef test_unique_references < TestCase
             unique_references_container.global_container( ...
                 'CLEAR','GLOBAL_NAME_SAMPLES_CONTAINER')
         end
+        %------------------------------------------------------------------
+        function test_save_restore_global_state_two_urcs(~)
+            clWa = set_temporary_warning('off','HERBERT:unique_references_container:debug_only_argument');
+            function urc_clearer()
+                unique_references_container.global_container('CLEAR','multifit_issue')
+            end
+            clSt  = onCleanup(@urc_clearer);
 
+            urc1 = unique_references_container('multifit_issue','char');
+            urc1(1) = 'aaa';
+            urc1(2) = 'aaa';
+            urc1(3) = 'aaa';
+            urc1(4) = 'bbb';
+            urc2 = unique_references_container('multifit_issue','char');
+            urc2(1) = 'ccc';
+            urc2(2) = 'ccc';
+            urc2(3) = 'ccc';
+            urc2(4) = 'bbb';
+
+
+            glc = unique_references_container.global_container( ...
+                'value','multifit_issue');
+
+            savestr1_or = urc1.to_struct();
+            savestr2_or = urc2.to_struct();            
+            unique_references_container.global_container( ...
+                'CLEAR','multifit_issue');
+
+
+            urr1 = serializable.from_struct(savestr1_or);
+            urr2 = serializable.from_struct(savestr2_or);            
+            glr = unique_references_container.global_container( ...
+                'value','multifit_issue');
+
+            savestr_copy1 = urr1.to_struct();
+            savestr_copy2 = urr2.to_struct();            
+
+            assertEqual(savestr1_or,savestr_copy1);
+            assertEqual(savestr2_or,savestr_copy2);            
+            clear clSt;
+        end
+        
+        function test_save_restore_global_state(~)
+            clWa = set_temporary_warning('off','HERBERT:unique_references_container:debug_only_argument');
+            function urc_clearer()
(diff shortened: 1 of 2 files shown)


herbert_core/utilities/classes/@unique_references_container/unique_references_container.m
```diff
@@ -280,6 +280,9 @@
             for ii = 1:val.n_objects
                 obj = val.get(ii);
                 [obj,hash] = build_hash(obj);
+                % as this code updates self and hence glc, need to refetch it every
+                % time here
+                glc = self.global_container('value',self.global_name);
```
abuts, line 285: This is not a fixture. 

The previous implementation was adding 23 runs to the container. 
And this one clear it up at every header calulation. 
Experiment for this run comes with 1 run where should be 23 runs -- one unique and 22 duplicates. 

I am surprized that side-effects are not reflected in tests. Probably Tobyfit does not yet uses all these instruments properly and everything works on IX_experiment which is not a conteiner

abuts (changes requested):
This is not a solution. 
you just delete each IX_detector copy from memory immediately after loading it.

abuts (approved):
It fixes the major issue with memory and placing all unique object into it. Unit tests addded which have proven this. 

As it solves the main problem with this contaner, it should go ahead

_test/test_unique_objects_container/test_unique_references.m
```diff
@@ -28,7 +28,147 @@ function delete(~)
...
+            assertEqual(glr.n_objects,3)
+            assertEqual(glr.n_unique,3)
+            assertEqual(glr.unique_objects,{'aaa','bbb','ccc'})
+
+
+            savestr_copy1 = urr1.to_struct();
+            savestr_copy2 = urr2.to_struct();
+
+            % need to compare structures as initial urc is broken after
+            % clear
```
oerc0122, line 124: Is this still the case?
```suggestion
```

_test/test_unique_objects_container/test_unique_references.m
```diff
@@ -28,7 +28,147 @@ function delete(~)
             unique_references_container.global_container( ...
                 'CLEAR','GLOBAL_NAME_SAMPLES_CONTAINER')
         end
+        %------------------------------------------------------------------
+        function test_save_restore_global_state_with_replacement(~)
+            clWa = set_temporary_warning('off','HERBERT:unique_references_container:debug_only_argument');
```
oerc0122, line 33:
```suggestion
            % Tests that instances in the GRC are still stored in mem and only cleared on 
            % unique_references_container's save/load
            clWa = set_temporary_warning('off','HERBERT:unique_references_container:debug_only_argument');
```

_test/test_unique_objects_container/test_unique_references.m
```diff
@@ -28,7 +28,147 @@ function delete(~)
...
+            glr = unique_references_container.global_container( ...
+                'value','multifit_issue');
+            assertTrue(isa(glr,'unique_objects_container'))
+            assertEqual(glr.n_objects,2)
+            assertEqual(glr.n_unique,2)
+            assertEqual(glr.unique_objects,{'aaa','bbb'})
+
+            savestr_cp = urr.to_struct();
+            % need to compare structures as initial urc is broken after
+            % clear
```
oerc0122, line 165: Is this still the case?
```suggestion
```

_test/test_unique_objects_container/test_unique_references.m
```diff
@@ -28,7 +28,147 @@ function delete(~)
...
+
+            % need to compare structures as initial urc is broken after
+            % clear
+            assertEqual(savestr1_or,savestr_copy1);
+            assertEqual(savestr2_or,savestr_copy2);
+            clear clSt;
+        end
+
+        function test_save_restore_global_state(~)
+            clWa = set_temporary_warning('off','HERBERT:unique_references_container:debug_only_argument');
```
oerc0122, line 131: Still not sure what this is testing separately to the 1st one.

_test/test_unique_objects_container/test_unique_references.m
abuts: This is always the case. Global container destruction destroys enteties referring it.

_test/test_unique_objects_container/test_unique_references.m
abuts: always the case

_test/test_unique_objects_container/test_unique_references.m
abuts: operations after replacemement may go different way (codewise). Not in Chris case, but in my alternative solution it happens for efficiency. 

This is the difference, so this test checks alternative internal code path.

oerc0122 (approved):

PR #1804, opened 2025-01-08
Merge qspec funcs

oerc0122:
Merge separate implementation of computing `qw` from original data.

This is a potentially incomplete process as there may be other `qw` calculators scattered throughout the code.


abuts (approved):
looks like reasonable change

abuts:
This will be part of `fix-calc-qw-pix2` branch (Re #1152)

Issue #1808, opened 2025-01-10
bug in `faccess_sqw_v4` `get_exp_info`


abuts:
recent changes in detector arrays have moved them in into experiment info. 

when I call `faccess_sqw_v4` and invoke `get_exp_info` I expect to get fully complsed Experiment containing detectors.

This does not happen which causes werid side effects.

PR #1810, opened 2025-01-13
get_exp_info returns fully filled experiment

abuts:
Solves Re #1808. 

After detectors moved into `Experiment` , `get_exp_info` should return `Experiment` with detectors. 

Here I have modified  `get_exp_info` to return `Experiment` with array of detectors and other file-accessors to use this method to build `sqw`. 

I have also extracted old format conversion into new conversion routine into single place and provided unit tests for it.

Also, small bug in load/save `sqw` fixed to avoid  saving/loading detector information twice.

### DIFF
--- a/horace_core/sqw/@Experiment/set_instrument.m
+++ b/horace_core/sqw/@Experiment/set_instrument.m
@@ -2,7 +2,7 @@ function   obj = set_instrument(obj,instr_or_fun,varargin)
 % add or reset instrument, related to the given experiment class
 %
 if isa(instr_or_fun,'IX_inst')
-    if numel(instr_or_fun) == 1 %replace all instruments
+    if isscalar(instr_or_fun) %replace all instruments
         % in the container with single input instrument
         uoc = unique_objects_container('IX_inst');
         uoc = uoc.add(instr_or_fun);
@@ -35,7 +35,7 @@ elseif isa(instr_or_fun,'unique_objects_container') && strcmp(instr_or_fun.basec
     obj.instruments = instr_or_fun;
 
 elseif isa(instr_or_fun,"function_handle")
-    if numel(varargin)==1 && iscell(varargin{1})
+    if isscalar(instr_or_fun) && iscell(varargin{1})
         argi = varargin{1};
     else
         argi = varargin;
--- a/horace_core/sqw/file_io/@horace_binfile_interface/private/convert_old_det_forms_.m
+++ b/horace_core/sqw/file_io/@horace_binfile_interface/private/convert_old_det_forms_.m
@@ -1,0 +1,44 @@
+function det = convert_old_det_forms_(detpar,n_instances)
+% Method used to convert detector formats written by any previous version
+% of Horace into Horace 4.01 data form.
+%
+% Input:
+% detpar      -- an old or new format detector information.
+%                Normally obtained from binary sqw file.
+% n_instances -- number of run, this
+%
+% Returns:
+% det         -- detector information packed in
+%                unique_object_container container and
+%                distributed into approriate number of input
+%                runs.
+
+if isstruct(detpar) % we do not need do deal with struct array
+    % here as this have never been used and stored in
+    detpar = IX_detector_array(detpar);
+end
+if isa(detpar,'IX_detector_array')
+    det = unique_objects_container('baseclass','IX_detector_array');
+    det = det.add(detpar);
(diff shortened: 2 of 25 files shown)


horace_core/sqw/@Experiment/set_instrument.m
```diff
@@ -35,7 +35,7 @@
     obj.instruments = instr_or_fun;
 
 elseif isa(instr_or_fun,"function_handle")
-    if numel(varargin)==1 && iscell(varargin{1})
+    if isscalar(instr_or_fun) && iscell(varargin{1})
```
cmarooney-stfc, line 38:
```suggestion
    if isscalar(varargin) && iscell(varargin{1})
```
checking

horace_core/sqw/file_io/@horace_binfile_interface/private/convert_old_det_forms_.m
```diff
@@ -0,0 +1,44 @@
...
+    detpar = IX_detector_array(detpar);
+end
+if isa(detpar,'IX_detector_array')
+    det = unique_objects_container('baseclass','IX_detector_array');
+    det = det.add(detpar);
+    det = det.replicate_runs(n_instances);
+elseif isa(detpar,'unique_objects_container') || isa(detpar,'unique_references_container')
+    if detpar.n_objects == 1
+        det = detpar.replicate_runs(n_instances);
+    elseif ~isempty(n_instances) && detpar.n_objects == n_instances
```
cmarooney-stfc, line 27: does this deal with _instances==0 which is non empty

horace_core/sqw/file_io/@horace_binfile_interface/private/convert_old_det_forms_.m
```diff
@@ -0,0 +1,44 @@
...
+            detpar.n_objects,n_instances)
+    end
+else
+    error('HORACE:horace_binfile_interface:invalid_argument', ...
+        ['Unrecognized input class: "%s" provided to the conversion method\n' ...
+        'Allowed inputs are: "struc", "ID_detector_array", ' ...
+        '"unique_objects_container" or "unique_references_container" '],...
+        class(detpar))
+end
+end
```
cmarooney-stfc, line 44: the above code deals with detpar as either unque-objects_container or uniqu_references_container and returns a unique-objects_container. I would expect returning  references container. In either case can you document the logic

cmarooney-stfc (changes requested):

horace_core/sqw/file_io/@horace_binfile_interface/private/convert_old_det_forms_.m
abuts: Your suggestion would work, and I thought about it. 

But decided, that it is safe to have `unique_obj_container` here. You have `unique_references_container` in `Experiment` and asignment will check if you provide correct value and distribute it within references. 

The routine here just processes old data and check them for uniqness. It is easier to test it this way without addressing global storage.

Due to its global nature, `unique_references_contaienr` should be exposed outside of `Experiment` as little as possible.

horace_core/sqw/@Experiment/set_instrument.m
abuts: No, this would not work. 
The function signatrue is:
`obj = set_instrument(obj,instr_or_fun,varargin)` so the suggestion would break the code

horace_core/sqw/file_io/@horace_binfile_interface/private/convert_old_det_forms_.m
abuts: added checks and tests for '0' to deal with it reasonably

horace_core/sqw/file_io/@sqw_binfile_common/get_exp_info.m
```diff
@@ -62,9 +60,12 @@
 else
     [header,pos] = get_single_header(obj,n_header);
 end
-%
 exp_info = Experiment.build_from_binfile_headers(header);
-
+% Get detector parameters
+% -----------------------
+detpar  = obj.get_detpar();
+detpar  = detpar.replicate_runs(exp_info.n_runs);
```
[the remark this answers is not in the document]
abuts: checks are inside the routine. Inconsitent values will throw appropriate error.

cmarooney-stfc (approved):

PR #1823, opened 2025-02-05
fast map

abuts:
According to Re #1825 implemented fast map to use to access indices.

Adds class providing fast map functionality to the code to use in maping between pixels and IX_experiments.

fast_map_vs_map_performacne routine shows, that the class, if oprimized, works 100 times faster than standard MATLAB `containers.Map`. If not optimized, is still 5 times faster for search and 11 times faster for search and insertion.

```
Find & Add keys to UINT        map   takes 324.988sec
Find & Add keys FAST MAP       map   takes 27.3212sec
Find       keys in UINT        map   takes 89.579sec
Find    keys in FAST MAP       map   takes 19.6152sec
Find all keys in FAST MAP      map   takes 18.6721sec
Find keys in FAST MAP Opt      map   takes 0.803015sec
Find all keys in FAST MAP opt  map   takes 0.304346sec
```


abuts:
Solves #1825

herbert_core/utilities/classes/@fast_map/fast_map.m
```diff
@@ -0,0 +1,347 @@
...
+            %
+            obj = optimize_(obj);
+        end
+    end
+    %----------------------------------------------------------------------
+    % Overloaded indexers. DESPITE LOOKING NICE, adding them makes fast_map
+    % 40-60 times slower even without using indexes itself. Disabled for this
+    % reason, until, may be mex is written which would deal with fast part
+    % of indices or we fully switch to MATLAB over 2021a, where you may    
+    % overload subsasgn using inheritance and implemented abstract methods.
```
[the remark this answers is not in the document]
abuts: well, indexes look nice and you have bothered to implement them in `unique_***_containers` despite internal methods `add` and `get` would be sufficient to implement all containers functionality. 

Why did you bothered to implement them?

Its nice to have them here to, but it kills all performance, which denies the purpose of this map so I have removed them after measuring their price.

This is the comment on how to avoid inefficient usage in `fast_map` and unit tests give extensive examples

herbert_core/utilities/classes/@fast_map/private/optimize_.m
```diff
@@ -0,0 +1,15 @@
+function obj = optimize_(obj)
+%OPTIMIZE_  % place values into expanded array or cellarray, containing
+% NaN or empty where keys are missing and values where
+% keys are present. This array/cellarray is optimal for fast
+% access to the values as function of keys.
```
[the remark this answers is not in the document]
abuts: This is exactly what the method is doing and I have no idea what I can say more about its functionality. More detailed explanation is the code itself.

_test/test_config/test_hpc_config.m
```diff
@@ -17,51 +27,78 @@ function test_get_default_memory_from_config_store(~)
...
+                    % memory in bytes, allocate doubles, divide by 2 for
+                    % stability on Jenkins Windows where multiple job
+                    % running
+                    data = zeros(floor(mem/8/2),1);
+                    ok = true;
+                    clear data;
+                    break
+                catch
+                    ok = false;
+                    if obj.n_attempts>1; pause(1); end
```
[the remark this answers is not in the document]
abuts: I have added some comments but in fact I have no idea if 1 second is too short, too long or fine. Special expensive investigation is needed to identify this.

It is fine from the point of view of the  delay in the time of thest execution.

_test/test_utilities_herbert/fast_map_vs_map_performance.m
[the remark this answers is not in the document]
abuts: we are not usually writing extensive documentation to tests. I am adding some now, but there is huge change if we decide to do this for tests.

herbert_core/utilities/classes/@fast_map/fast_map.m
```diff
@@ -0,0 +1,347 @@
...
+        % The map optimization works by allocating large continuous array
+        % with places for keys as indices. Where keys correspond places in
+        % array, result contains values, and where no keys present, array
+        % contains nan-s. The property below shows how much more elements
+        % the optimization array should contain wrt the number of keys in
+        % the map. E.g. if empty_space_optimization_limit == 5 and there
+        % are 100 keys, optimization array would contain no more than 500 
+        % elements. If this ratio is not satisfied, i.e. 
+        % if max(keys)-min(keys) > number_of_keys*empty_space_optimization_limit,
+        % map optimization gets disabled and correspondence between keys
```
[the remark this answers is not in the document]
abuts: Its extreamly difficult to describe this algorithm while reading the code makes it much simple. 
I tried my best and have modified the description, but not sure if it is ok. Can not make it better though.

_test/test_config/test_hpc_config.m
```diff
@@ -17,51 +27,84 @@ function test_get_default_memory_from_config_store(~)
...
+                    % stability on Jenkins Windows where multiple job
+                    % running
+                    data = zeros(floor(mem/8/2),1);
+                    ok = true;
+                    clear data;
+                    break
+                catch
+                    ok = false;
+                    % wait for short period of time to allow other instance
+                    % of the test job releave its memory
```
cmarooney-stfc, line 49: I have no idea what "releave its memory" means

_test/test_utilities_herbert/fast_map_vs_map_performance.m
```diff
@@ -0,0 +1,141 @@
+% script to compare performance of custom fast_map class and
```
cmarooney-stfc, line 1: this is not a test. It is a benchmarking script. Regardless of the need for more documentation or otherwise of tests, this is a different beast and is impenetrable without these comments.

I will mark this reveiw "requires changes" and you will reply "understood" to indicate that you have taken on board this difference

_test/test_utilities_herbert/test_fast_map.m
```diff
@@ -0,0 +1,166 @@
...
+            assertEqual(fm.KeyType,'uint64');
+            assertEqual(fm.keys,uint64(10));
+            assertEqual(fm.values,1);
+        end
+        %------------------------------------------------------------------
+        function test_get_all_val_for_keys_optimized_no_checks(~)
+            n_keys = 100;
+            base_key = 10+round(rand(1,10*n_keys)*(10*n_keys-1));
+            base_key = unique(base_key);
+            n_keys = numel(base_key);
```
cmarooney-stfc, line 39: you have not said what this does, so the conversation is not resolved. Please add comments. You may refer to your benchmarking script in doing so

herbert_core/utilities/classes/@fast_map/private/optimize_.m
cmarooney-stfc, line 5: what is an expanded array/cellarray - what does it contain? what are we doing putting NaNs in it in places? Why is it optimal? Why do you assume that your code is sufficiently transparent to be self-explaining? Please document it.

cmarooney-stfc (changes requested):

_test/test_config/test_hpc_config.m
abuts: typo -- "relese"

_test/test_utilities_herbert/test_fast_map.m
abuts: "% generate specified number of random keys in the range
% exceeding  number of keys by 10. This approximate runids obtained from
% two  independent experiments provided in two cycles which are not too far
% from each other in time."
This is row abowe row 36. What else you want?

_test/test_utilities_herbert/fast_map_vs_map_performance.m
abuts: this is test of performance, which is critical in this case so this is a kind of test.  It is not a test itself as it does not start with `test_` keyword so it will not be run automatically. 

We do not have agreement on where to place such code so I put it alongside the code which tests class validity. If there are other suggestions -- I am ready to discuss and implement them if decided. 

What should I understand and what changes I should do wrt this?

herbert_core/utilities/classes/@fast_map/private/optimize_.m
abuts: optimal, because I have measured performance and found that this solution is the best. Here I state this fact, and this is how code description works. 

there is array, containing values where the keys are and there are nan-s where they are not. 

The description is already the size of the code itself -- I have described intentions and how it is done -- is in the code. I just do not understand what else I can write to make it clearer.

Issue #1828, opened 2025-02-24
L&D for RSE group. Horace/PACE presentation


abuts:
There is request to provide learning and development event for RSE group about Horace. 
It will be introduction to Horace and review of basic scripts used by  users. 

This material may be also useful for a new future Horace developers, so worth adding this material to Horace documentation itself.

abuts:
Closed as not approved for reviewer

PR #1829, opened 2025-02-25
Introduction  to Horace (RSI L&D event)

abuts:
Documentation to RSI learingn and development event according to Re #1828


documentation/presentations/HoraceIntro.pptx
cmarooney-stfc, line 1: This is a partial review of the slide set present here.

General: remove the ISIS background. The problem is the blue trajectories and scatterings on the right. On some slides they appear to be part of the presentation and are confusing.

Slide 2: Better split up into 3 or more slides to add all the necessary information.
The top left hand circle is the neutron source. This needs labelling. The label for the incoming proton needs rotating.
The functions of M1 and M2 are not labelled.
The nimonic chopper requires explaining. The fermi chopper probably also requires explaining. The alternative disc chopper will also require explaining. Reference this: (http://www.dataweb.stfc.ac.uk/ChiltonCatalog/1990.ral-open-days-handouts/what-type-of-choppers-do-we-use.pdf)
The ki/kf/Q diagram and definitions should go on a separate slide.

Slide 3. This is about the incoming spectrum of the neutrons after the moderator and needs a label to say it is hihgly non-monochromatic. The detector bank illustration is not needed here. The monochromtic spectrum after the chopper in slide 4 could be put here instead of having a separate slide

slide 5. The surface is not a 4D surface but a 2D contour plot. Say what the colours represent.

slide 6  The fermi chopper on the right is highly confusing and should be deleted. More labels for the other parts and trajectories are needed. What is the large purple thing top left and is it needed?

slide 7. This needs much more labelling. Better as at least 2 slides giving the left hand and right hand diagrams separately.

slide 8. Omit. This is an introduction to Horace, not a peer into the future of neutron scattering.

slide 9. Need a separate slide to describe the mantid box on the left. Separate slides for the mantid boxes at the bottom.
Need to define what S(q,w) is.

slide 10. This is puttng all of Horace on one slide. Separate it out into several slides

slide 11 I have no idea what this is showing; it needs proper explanation on the slide

cmarooney-stfc (changes requested):

abuts:
Thank you for reviewing this. Unfortunately, the presentation have been performed. Are you insisting on making them?

Issue #1830, opened 2025-02-28
Horace version 3 sqw files can not be loaded in Horace-4


abuts:
user reported the issue and provided the file (large 76GB). 
It fails while loading in Horace-4. 

This looks like a bug to fix

abuts:
fixed by Re #1831

PR #1831, opened 2025-02-28
 should fix issue with user request

abuts:
This is seems the solution of the user issue Re #1830

### DIFF
--- a/herbert_core/classes/instrument_classes/@Goniometer/private/check_and_set_uv_.m
+++ b/herbert_core/classes/instrument_classes/@Goniometer/private/check_and_set_uv_.m
@@ -5,7 +5,7 @@ function obj = check_and_set_uv_(obj,name,val)
 if isnumeric(val)
     if numel(val)==3
         cand = val(:)'; % make it row vector
-    elseif numel(val) == 1
+    elseif isscalar(val)
         cand = [val,val,val];
     else
         error('HERBERT:Goniometer:invalid_argument',...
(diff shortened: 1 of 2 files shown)


herbert_core/classes/instrument_classes/@Goniometer/private/check_and_set_uv_.m
```diff
@@ -5,7 +5,7 @@
 if isnumeric(val)
     if numel(val)==3
         cand = val(:)'; % make it row vector
-    elseif numel(val) == 1
+    elseif isscalar(val)
```
cmarooney-stfc, line 8: I am presuming that this is the actual fix but I have no idea why it works. To ensure that future development in this area does not accidentally revert this change, please add in the comments a full explanation of what it does. Thanks

herbert_core/classes/instrument_classes/instrument/@IX_experiment/IX_experiment.m
```diff
@@ -343,9 +343,12 @@
             old_fldnms = {'filename','filepath','efix','emode','en','cu',...
                 'cv','psi','omega','dpsi','gl','gs','uoffset','u_to_rlu'};
             obj = IX_experiment();
+            obj.do_check_combo_arg = false;
```
cmarooney-stfc, line 346: please explain in comments around this code why the checks need to be turned off at this poiint

cmarooney-stfc (changes requested):

herbert_core/classes/instrument_classes/@Goniometer/private/check_and_set_uv_.m
abuts: `isscalar` is the new Matlab recommended way of dealing with expressions like `row 8`. I am changing code to Matlab recommended as I go through code, but the actual solution is in other file.

cmarooney-stfc (approved):

Issue #1841, opened 2025-03-12
`disable_preselection` property on a generic projection


abuts:
`cut` using generic projection works in two stages. 

1) take image grids, related to source and target projections and calclulate their intersection to identify source grid cells which contain pixels which may contribute into target image.

2) go through source image cells containing pixels which may contribute to target projection, pick these pixels and place them into target grid cells calulating averages which form the target image. 

When one calculates instersection of two grids related to two random projections it may be difficult to judge, if the grids intersection is performed correctly and all pixels which may contribute into target image are indeed included in the selection. 

To do ultimate test that two grid intersections are indeed correct, one may want to make a cut over all pixels present in the source object. This may be inefficient but if this cut and the cut done with preselected image cells appear to be the same and contain the same pixels, one may be sure that the cell intersection is calculated correctly. 

The purpose of this ticket is to add to generic projection property, which, if true, would disable pixel preselection algorithm in the `cut` to allow cutting over all pixels contributing to source image. 

The property is expected to be used in testing and in special cases when it is clear that all pixels need to be included in the cut.

abuts:
fixed by Re #1842

PR #1842, opened 2025-03-12
disable pix preselection

abuts:
Fixes Re #1841 

Testing property added, code to use this property is included in `cut` and unit tests written to ensure it works as expected.


_test/test_algorithms/test_cut.m
cmarooney-stfc, line 1: Comment on PR as a whole. Please add description to PR saying how the source ticket was fixed. This can then be compared with the individual changes to be sure we understand what is happening.

horace_core/sqw/coord_transform/@aProjectionBase/aProjectionBase.m
```diff
@@ -144,6 +148,9 @@
         % methods for specific projections, but 4D transformation is
         % algorithmically simpler so actively used in tests.
         do_3D_transformation_ = true;
+        %if set to true, disables pixel preselection part of cut algorithm.
+        % Cut then works over all pixels present in the source image.
+        disable_pix_preselection_ = false;
```
cmarooney-stfc, line 153: state that by default we expect this var to be false as disabling is inefficient

horace_core/sqw/coord_transform/@aProjectionBase/aProjectionBase.m
```diff
@@ -171,6 +178,24 @@
             end
             [obj,par] = init(obj,varargin{:});
         end
+        function obj = copy_proj_param_from_source(obj,cut_source)
+            % a projection may need some information from the object,
+            % it used to process. For example, all projections used for cut
+            % need to know lattice parameters of sqw object used as source
```
cmarooney-stfc, line 184:
```suggestion
            % Load information into the projection from the object being processed, 
            % which is required to perform the operation. For example, all projections 
            % used for cut need to know lattice parameters of sqw object used as source
```
the original text says why you need the information but not what the method actually does with the information.

horace_core/sqw/coord_transform/@aProjectionBase/aProjectionBase.m
```diff
@@ -382,6 +414,12 @@
                 % available and enable if it available
                 targ_proj.targ_proj = obj;
                 obj.targ_proj = targ_proj;
+
+                if targ_proj.disable_pix_preselection_
+                    bl_start = 1;
```
cmarooney-stfc, line 419: add 
% select all pixels

cmarooney-stfc (changes requested):

_test/test_algorithms/test_cut.m
abuts: I hope description is sufficiently complete, though I am sure that ticket contains all necessary information and writing solves ticket Re #XXXX is sufficient enough

cmarooney-stfc (approved):

Issue #1776, opened 2024-11-25
kf_sphere_proj


abuts:
If one have a look at  inelastic stattering signal in direct (rather than reciprocal) coordinate system, one can see clear elastic line signal at low energy transfer range. The signal can be easily approximated by double-side exponential function (see fig.2). 
![image](https://github.com/user-attachments/assets/98d3c2df-c5af-4629-905f-2bb2c31a751d)
fig.1
![image](https://github.com/user-attachments/assets/ad85418e-c927-4233-a7da-1a985f12a94a)
fig.2

The coordinates of axis in fig 1 are X-axis -- dE, Y-axis -- scattering angle in direct coordinate system. 

There are currently no way to produce such image in Horace, as $`\|Q\|`$ along Y-axis changes from 0 to $\|k_i\|$.

Similar image may be produced using $`\vec{k_f}`$ coordinates, namely $`\vec{k_f}=\vec{Q}-\vec{k_i}`$

This is the ticet to build spherical projection which would produce Horace image in $`\vec{k_f}`$ coordinates, namely $`\|k_f\|`$, $`\theta`$, $`\phi`$ coordinates where $`\theta`$, is the angle between beam and $`\vec{k_f}`$ directions.

cmarooney-stfc:
BLOCKED by #1152

abuts:
finalized by Re #1845. Further development/refactoring is possible on request

abuts:
Minimally addressed by Re #1845. Further development is possible on request

PR #1845, opened 2025-03-26
instrument view cut

abuts:
The PR can be dealt with after Re #1838, #1842 and #1843 are merged as it depends and based on them, so it will be difficult to understand what belongs to this PR and what comes from the previous bunch.

closes Re #1776

The PR delivers thing, a bit different from the ticket requested, as `kf_sphere_proj` defined as described in ticket would not work with standard cut. Standard cut relies on a projection to be able to convert pixels to image and back image to pixels, getting initial values. The suggested projection collapses pixel information into 3-Dimensional hull, so the inverse transformation becomes ill defined. 

Despite that, limited usage of this projection appears to be useful, as it provides valuable diagnostic tool to identify issues with instrument. The results have been discussed with couple of Excitation group users and found to be useful, especially for diagnosis of old experimental data or sqw files. Further development in this direction (incluing modifying `kf_sphere_projection` and building proper projection) is possible, but will be performed after cousultation with customers and evaluation of cost/value ratio.

From software point of view, this PR delivers:

1) `kf_sphere_proj` class, which inherits from sphere projection and defines minimal transformation, necessary for such class and implementing the transformation, described in ticket. This class should not be advertized to users, as have many contstrains. Necessary set of unit tests for these transformations is also available. 

2) Minor refactoring of a projections interface, allowing to decrease code duplication (Re #https://github.com/abuts/Horace/commit/d1083f5ff5305227b96a015183f68dcfa2002578) and modify captions produced by cuts done with `kf_sphere_proj` without the need to overload `spherical_axes` class to provide axes for `kf_sphere_proj`. `sphere_axes` with modified plot captions is sufficient for the purposes, `kf_sphere_proj` is currently used.

3) User facing `instrument_videw_cut` algorithm which uses `kf_sphere_proj` and `cut` inside it and returns the results of the appropriate cut to analyze by users.


abuts:
This contribution has been mainly merged by PR https://github.com/pace-neutrons/Horace/pull/1843

Issue #1740, opened 2024-09-17
SQW compression methodology to speed up I/O


cmarooney-stfc:
Planning ticket for PACE 2 work. Please delete and replace this in the tracking epic #1738 if it is a duplicate

tgperring:
An ADD (11_sqw_compression.md) has been drafted that gives more detail and discussion about the compression objectives and methodology.

PR #1847, opened 2025-03-28
SQW file compression ADD

tgperring:
Adds an ADD that describes the reasons for having a sqw compression methodology and outlining a proposed solution to the folder \documentation\add

Refers to issue  #1740

### DIFF
--- a/documentation/add/11_sqw_compression.md
+++ b/documentation/add/11_sqw_compression.md
@@ -50,7 +50,7 @@ The array is presently stored as a float32 array i.e. 4 bytes per element. Thus,
 
 The proposed compression scheme exploits the following two facts:
 - `qx, qy, qz, en` are completely determined by {`irun, idet, ien`}, because those indices point to the experiment information that enables the value of the fixed energy, the detector element location, the energy bin (and hence energy transfer) and crystal orientation and lattice parameters that enable `qx, qy, qz, en` to be calculated.
-- In most single crystal experiments, the fraction of pixels that contain non-zero signal is very small. For example:
+-	In most single crystal experiments, the fraction of pixels that contain non-zero signal is very small. For example:
 	- Iron data: runs 15057 and 15097 (800 meV incident energy data): 0.13
 	- RbMnO3: a selection of runs: 0.010 – 0.012
 
@@ -95,7 +95,7 @@ In practice,
 **Table 2**
 
 | Fraction of non-zero signal | Bytes per pixel float32 signal | Percentage of current format | Bytes per pixel float64 signal | Percentage of current format
-|-----------------------------|-----------------|-----------------------------
+|---------------------|--------|-----|--------|---------------------------------
 |  f = 0  (best case) |  8     | 22% |  8     | 22%
 |  f = 0.011 (RbMnO3) |  8.09  | 22% |  8.18  | 23%
 |  f = 0.13 (Fe)      |  9.04  | 25% |  10.08 | 28%
@@ -138,10 +138,10 @@ Alex and Jacob have done some of benchmarking and profiling: Alex on total timin
 Profiling is important because past experience with accessing unique object storage, and of accessing detector and run information in Tobyfit, has shown that without careful optimisation these operations can become very time-consuming.
 
 
-### Why store the signal and variance as float64?
-The contents on disk and in memory will then be identical. A problem we have had is comparing data when debugging because pixels close to the edge of a bin can move from one bin to another, as well as the obvious 7 sig. fig. rounding that takes place on double to single conversion.
+### Do we need to store the signal and variance as float64?
+A problem we have had when comparing data during development and debugging arises because pixels close to the edge of a bin can move from one bin to another, and signal and error are rounded to 7 sig. fig. when these values are converted from double to single precision on saving to disk.
 
-The problem of pixel-to-bin mapping should disappear with on-the-fly computation of `qx,qy,qz,en`, as this will be done in memory from the run-detector-energy bin indices. That leaves just the question of the signal and error. Rounding to 7 sig fig may not be important. This should be considered further; particularly if the fraction of non-zero signal pixels is large for a significant proportion of experiments, then saving signal and error as float64 might result in a significant penalty (see Table 2). This will certainly be the case when simulating data with a model, when typically *all* pixels will have non-zero simulated signal. In this case the compressed format file will be 50% larger with float64 signal compared to float32 (and 2/3 the size of the current sqw file compared to 44% for float32).
+The problem of qx, qy, qz, en being rounded to single precision when an sqw object is saved to file is created should disappear with on-the-fly computation of qx,qy,qz,en on read, as this will be done reproducibly in double precision in memory from the stored run-detector-energy bin indices. That leaves just the question of the signal and error. Rounding these quantities to 7 sig fig may not be important. The disadvantage should be considered further] because there is a potential non-trivial disk space saving. If the fraction of non-zero signal pixels is large for a significant proportion of experiments, then saving signal and error as float64 might result in a significant penalty (see Table 2). This will certainly be the case when simulating data with a model for S(Q,w), when typically *all* pixels will have non-zero simulated signal. In this case the compressed format file will be 50% larger with float64 signal compared to float32 (and 2/3 the size of the current sqw file, compared to 44% for float32).
 
 
 ### The meaning of ‘non-zero signal’
@@ -151,11 +151,13 @@ By non-zero signal we actually mean zero signal and zero error: nothing was coun
 ### Values of nrun, ndet_max, ne_max
 In practice, we won’t always know the values of nrun, ndet_max, ne_max in advance of construction of an sqw file, for example, if the sqw file is built up during an experiment by accumulating the data of newly arrived nxspe files as they appear. In practice, however, we can use default values that will be larger than any feasible experiment in the lifetime of ISIS or other facilities:
 
+**Table 3**
+
 |          | Estimate of maximum value | Value | Comment
-|-----------------------------|-----------------|-----------------------------
+|----------|---------------------------|-------|--------
 | nrun     | 0.025 degree steps over 360 degrees (full rotation) | 1.44e4 |  x30 most ever done!
-| ndet_max | CSPEC at ESS if it gets the multigrid detectors | 1e6 | Very unlikely to get them!
-| ne_max   | Somebody once had 800 to my knowledge  | 1e3 | x20 finer than resolution
+| ndet_max | CSPEC at ESS if it gets the multigrid detectors     | 1e6    | Very unlikely to get them!
+| ne_max   | Somebody once had 800 to my knowledge               | 1e3    | x20 finer than resolution
 
 Result: N_max = 3.6e11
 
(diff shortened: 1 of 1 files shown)


documentation/add/11_sqw_compression.md
```diff
@@ -0,0 +1,201 @@
...
+### The principles of the scheme
+
+The proposed compression scheme exploits the following two facts:
+-	qx, qy, qz, en are completely determined by {irun, idet, ien}, because those indices point to the experiment information that enables the value of the fixed energy, the detector element location, the energy bin (and hence energy transfer) and crystal orientation and lattice parameters that enable qx, qy, qz, en to be calculated.
+-	In most single crystal experiments, the fraction of pixels that contain non-zero signal is very small. For example:
+	- Iron data: runs 15057 and 15097 (800 meV incident energy data): 0.13
+	- RbMnO3: a selection of runs: 0.010 – 0.012
+
+Accordingly, so long as we keep track of which pixels contain non-zero signal, we only need to keep the following information: 
+-	irun, idet, ien: for all pixels
```
mducle, line 58: If we restrict the specification to allow only a single fixed energy, one set of energy bins for all contributing runs and the same detector mask on all runs (this defines 100% of recent data from ISIS) you could get additional compression by simply not storing the zeros at all (not even storing `irun`, `idet`, `ien`), and computing them from the detector definition.

documentation/add/11_sqw_compression.md
```diff
@@ -0,0 +1,201 @@
...
+|  f = 0.25           |  14    | 39%
+|  f = 1 (worst case) |  20    | 56%
+		
+For small values of f, the storage is dominated by the space needed to hold {irun, idet, ien} for every pixel. For typical data, the storage requirement is reduced to about 1/3 that of the present sqw files. 
+
+
+
+### The proposed scheme and the achievable compression
+
+To improve on the above, special knowledge of the maximum values of the indices, namely {nrun, ndet, ne} can be used to reduce the space to hold {irun, idet, ien} for every pixel. With the typical values nrun = 500, ndet_max = 70,000, ne_max = 300 then N_max = 1.05e10. If {irun, idet, ien}  is converted into a linear index using the knowledge of nrun, ndet_max, ne_max, then the pixel indexing information can easily be held in a float64 (maximum integer 2^53 = 9e15) i.e. 8 bytes.
```
mducle, line 89: I don't understand why you want to store integers as `float64` - wouldn't an unsigned 32-bit integer be more appropriate (and take half the space)? The maximum value of an `uint32_t` is `2^32=4.29e9` which should be more than large enough to store any runs used over the lifetime of ISIS. You could also get away with an unsigned 16-bit integer for `ien` since it is unlikely people will use more than 65536 energy bins. (LET has 98312 detectors so you would need 32-bit integers for `idet` there).

If you need more than `4e9` indices then an unsigned 64-bit integer can store values up to `2^64=1.8e19`....

documentation/add/11_sqw_compression.md
```diff
@@ -0,0 +1,201 @@
...
+
+In practice, 
+-	With any feasible values for {nrun, ndet_max, ne_max} in the lifetime of ISIS and other facilities, 8 bytes is all that is needed (see additional notes below for details).
+-	Because the overall storage requirement in practice is dominated by {irun,idet,ien}, we could store the signal and error as float64 with little penalty in disk space compared to storing as float32: only for f>0.21 does the penalty exceed 15%. Thie advantage is that the representation in memory and on disk are identical, with little penalty in overall storage in bytes. The following gives the storage needs for the two case (and percentage size compared to the current sqw format):
+
+**Table 2**
+
+| Fraction of non-zero signal | Bytes per pixel float32 signal | Percentage of current format | Bytes per pixel float64 signal | Percentage of current format
+|-----------------------------|-----------------|-----------------------------
```
mducle, line 98: You need additional `|` in this line to match the line above and below, e.g.

```
|---------------------|--------|-----|--------|-----
```

documentation/add/11_sqw_compression.md
```diff
@@ -0,0 +1,201 @@
...
+
+### The meaning of ‘non-zero signal’
+By non-zero signal we actually mean zero signal and zero error: nothing was counted. This is distinct from the case of a background subtraction that has been performed in the data reduction such that there is zero signal and non-zero error. 
+
+
+### Values of nrun, ndet_max, ne_max
+In practice, we won’t always know the values of nrun, ndet_max, ne_max in advance of construction of an sqw file, for example, if the sqw file is built up during an experiment by accumulating the data of newly arrived nxspe files as they appear. In practice, however, we can use default values that will be larger than any feasible experiment in the lifetime of ISIS or other facilities:
+
+|          | Estimate of maximum value | Value | Comment
+|-----------------------------|-----------------|-----------------------------
```
mducle, line 155: You need additional `|` in this line to match the line above and below, e.g.

```
|----------|---------------------------|-------|---------
```

mducle (commented):
Just some comments... I'm not approving or requesting changes...

documentation/add/11_sqw_compression.md
mducle: Oh certainly we would need to know where the zeros are - what I suggest is that if all the runs have the same detectors and same energy bins then you don't need to *store* the zeros in the file - you can always compute them because you've got the detector map already from the instrument data, and you know the rotation angle of each run.

Obviously this doesn't work if the runs have different fixed energies, or different energy bins but we haven't made such scans in about 15 years... all single crystal experiments at ISIS since then have used the same fixed energy, energy bins and just rotated the crystal.

documentation/add/11_sqw_compression.md
abuts: 500*100000*500 = 2.5*10^10 while 2^32 (unit32)=4.29*10^9. I think Toby meanis uint64, which provides sufficient space for 2^64=1.8*10^19. (i.e. 1000*100 000*1000=10^11)

I did estimates and initial roitines to compress/uncompress indices using uint64 and this is 8 bytes. I think smaller values would create unnecessary complexity without much benifit

documentation/add/11_sqw_compression.md
abuts: hmmm... This is different from what we have discussed and may be similar to what Savickii did in Mantid... But I do not think it would work.

When you do cut you are loading page of pixels into memory and the page consist of `random`* selection of cells.  You would not know a-priory which indices have contributed into this page cells unless you calculated it beforehand. Indicating that bunch of these indices have zero signal -- fine, predicting they contribute into this cell beforhand -- well, some perevious calculations shoud suggest that part of detector's tragectories been there -- Savickii normalization -- as far as I remember, it duplicated the size of MD workspace and I do not fully understand how they do it.

*`random` -- means difficult to predict -- cut preselect cells which may contribute into resulting image

documentation/add/11_sqw_compression.md
tgperring: In the file we could just as well use uint64, but there may be an advantage in using float64 simply because we might want to store [irun,idet,ien,signal,error] as a single array (or rather, [irun_idet_ien, signal, error] as a 3 x N array. This becomes a detail depending on how we would hold the information (some thoughts were at the end of the document, but for unrelated reasons we may decide that signal and error are better stored as float32, not float64, in which case there is no advantage in holding a linear index that is decoded to [irun,idet,ien] as float64).

What i wanted to point out is that (i) float64 will be good enough to hold a linear index for any reasonable future needs, and (ii) in the rest of Horace I think the total number of pixels is limited to 2^53 anyway. as the default is for Matlab to hold all numerics as float64 and there will most likely be places where there are loops over pixel indicies (either explicitly, or implicitly as work is done on any array a chunk at a time).

documentation/add/11_sqw_compression.md
mducle: @abuts I think Andrei Savici's normalisation workspace is slightly different (and is probably more correct than what we do in Horace). Rather I think Andrei's normalisation constants is to ensure that you can always recover the raw count information and from that recompute the uncertainties accurately. 

What we're discussing here is just how to handle the zeros.

Now, please correct me, but I think the current cut algorithm is:

1. Determine the set of coarse bins which are within the limits the user sets
2. Load those sets of pixels into memory from file
3. Find out which actual pixels are within the limits and discard the others
4. Compute the rebinning with the pixels within the limits (in memory).

If we don't store the zeros in the `.sqw` files, we would have to modify it to:

1. Determine the set of coarse bins which are within the limits the user sets
2. Load those sets of pixels into memory from file
3. Find out which actual pixels are within the limits and discard the others
4. Determine zero bins:
4a. Loop through all runs and determine if the coarse bins for that run are within the limit
4b. Loop through all the possible pixels in those coarse bins and determine if 1) they are not in the non-zero list, and 2) are within the limits.
4c. If so add those zeros to the list of pixels loaded to memory.
5. Compute the rebinning with the pixels within the limits (in memory).

The calculations in step 4 is essentially what is being done in `gen_sqw` now, and I admit it is a significant amount of extra computation relative to the amount done at present but depending on the size of the cut it might not be that much extra processing in _absolute_ terms (e.g. it might still be quicker than loading all those zeros from disk).

tgperring: What you outline is correct (and of course can be done whether there is one or more detector files, different ei for each run etc.). However,  the reason I did not propose we do this is because of what you say in Step 4: the task is essentially what gen_sqw is doing, and every time one makes a cut the calculation will have to be repeated. Essentially , it means that each cut has a computational overhead of the same order as re-running gen_sqw for the entire set of runs.

Andrei Savici's algorithm reduces the work because I think that he computes where the the trajectory for a detector intersects the bounding volume of the output cut (or maybe the coarse bins in the input sqw file) and I believe it is coupled with a different data normalisation method as well. We could do the same, but it would turn a quick win in a clearly circumscribed part of the code (effectively a change to file format that reduces the file size by a factor 4, and taking advantage of existing algorithms) into a much larger task. This larger task would require creating new algorithms to perform computation of the intersection of detector trajectories, which would be more complex for cuts like the spherical cut (general cuts like energy-focussed?), and perhaps changing how we do our normalisation. It will turn the job into a longer and more open-ended one.

The speed of cuts has been highlighted as a priority by David, and for this reason I think the immediate concern should be to get the quick win.

mducle: Ok... fair enough.

abuts: completely agree with Toby

documentation/add/11_sqw_compression.md
tgperring: Hmm... the .md plugin that Notepad++ uses has no problem parsing to a nice table. I've just seen that the one in Github objects however. I'll make the required change.

documentation/add/11_sqw_compression.md
tgperring: See comment above; likewise.

documentation/add/11_sqw_compression.md
tgperring: BTW, I'll add a summary of the discussion above into the ADD so it isn't lost.

abuts (dismissed):
I disagree with the need to save signal/errors as `doubles` but otherwise all good. 

Assuming that both file formats have to be implemented anyway,  I approve the change -- we may discuss the details later -- software design plans have to be created anyway.

abuts (approved):
looks good

Issue #1860, opened 2025-04-24
Write ADD and other document on `sqw_op_bin_pixels`


cmarooney-stfc:
During the PACE standup on 24.4.25, Alex said that his work on #1854 had given him a much better understanding of the issues he had been investigating. This seems a good opportunity to document that improved understanding. Topics reported in that standup include background insertion and extraction, and role of non-orthogonal lattices.

Alex should document the parts of this work - whatever they are - where his understanding has most improved and add them to the ADD collection.

Some documentation is present in user documentation. A progress is possible only from feedback to this documentation

Issue #1854, opened 2025-04-11
`sqw_op` with binning pixels


abuts:
`sqw_op` allows to perform multiple operations over sqw object, but does not allow change pixels coordinates.

When one needs to change pixels coordinates one needs to run binning algorithm. The purpose of this ticket -- to modify `sqw_op` to be able to run binning algorithm when necessary. 

This ticket is simple and would deliver major part of the algorithm, necessary to implement focusing projection

abuts:
Solved by Re #1861

abuts:
This ticket have been implemented as part of #1861 job.

PR #1861, opened 2025-04-28
sqw_op_bin_pixels algorithm

abuts:
solves Re #1854

This PR provides `sqw_op_bin_pixels` algorithm and unit tests for it.

The algorithm itself is the combination of `PageOp`, `cut` and `combine` algorithms so contains very little of its own code.

user of the algorithm should provide correct function which may modify pixels coordinates and correct binning parameters, which work like cut algorithm working over modified coordinates.

More detailed  description of the algorithm will be provided within a separate ticket Re #1860


abuts:
This contribution has been mainly merged by PR https://github.com/pace-neutrons/Horace/pull/1843

Issue #1859, opened 2025-04-24
Write ADD document on Horace graphics


cmarooney-stfc:
In the process of fixing ticket #1787 @tgperring has determined it will be useful to write an ADD document on Horace graphics as a whole. As this task has not previously been ticketed, this ticket provides the tracking for the issue, to enable progress and prioritization to be recorded.

abuts:
fixed by Re #1869

Issue #1787, opened 2024-12-12
Unexpected behaviour when plotting arrays of IX_dataset_*d objects


tgperring:
There are a few minor issues that have arisen when plotting arrays of IX_dataset_1d and IX_dataset_2d objects
- the methods to convert an array of lower dimension IX_dataset_*d objects to higher dimensional objects contain bugs so that they do not work
- the utility function graph_range, used by the limits setting methods lx, ly, lz and lc, has bugs when plotting arrays of objects, and unexpected behaviour in some limiting cases

PR #1866, opened 2025-05-06
Refactor Horace graphics and add examples to ADD.

tgperring:
Refers to #1787 

The graphics algorithms have been refactored to make them more robust, fixing a number of bugs, simplifying the chain of nested calls to graphics management functions of the 'genie_figure' type, making full consistency of `sliceomatic` across `IX_dataset_*d`, `d*d` and `sqw`. The structure remains very similar. Included are a comprehensive set of tests.

Regarding documentation, an ADD has been created (12_Horace_graphics.md) and addressed by dependent ticket Re #1859 that explains how the graphics management and uniform interface operates (the latter was briefly described already in 07-sqw_redesign, and the same inheritance diagram is repeated) which moved into its own PR #1869. In addition, a folder has been created `/examples/` that contains a script `example_plotting.m` that demonstrates some very basic plotting (calling a function `data_for_plotting.m` in the same folder that returns some instances of `IX_dataset_*d`, `d*d` and `sqw` objects).


herbert_core/utilities/classes/@IX_dataset_3d/sliceomatic.m
```diff
@@ -26,100 +29,173 @@
...
 
-[par,keyword,present] = parse_arguments(varargin,arglist,flags);
-
-% Check input arguments
-% ---------------------
 if ~isempty(par)
     error('HERBERT:IX_dataset_3d:invalid_argument', ...
-        'Invalid sliceomatic arguments:\n %s',disp2str(par))
+        'Invalid sliceomatic arguments given. Only keyword arguments are permitted')
 end
```
abuts, line 46: I would disagree with this simplification. Previous state was showing what is provided invalid, The new one just infoms about some "keywords aruments".

horace_core/spaghetti_plot.m
```diff
@@ -644,3 +644,15 @@ function plot_labels(labels,xvals)
...
+function out = set_argout(nout, varargin)
+% set output arguments depending on number of output arguments
+% requested.
+%
+% expected usage is to set varargout, and because of this, no
+% output cell array is allocated (is this reasonable?)
+for i=1:nout
+    out{i} = varargin{i};
+end
+end
```
cmarooney-stfc, line 658: This does not feel comfortable. You should force the use of varargout, or provide a mechanism to swtich according to use. Suggest following the use of nargout(@function_handle) where the return value is negative if varargout is used.

herbert_core/utilities/classes/@IX_dataset_3d/sliceomatic_overview.m
```diff
@@ -0,0 +1,81 @@
+function varargout = sliceomatic_overview (w, varargin)
+% Plots an IX_dataset_3d object using sliceomatic viewed straight down one of the axes
+% When the slider for that axis is moved we get a series of what appear to be
+% two-dimensioonal slices.
```
cmarooney-stfc, line 4:
```suggestion
% two-dimensional slices.
```

cmarooney-stfc (changes requested):
just commenting on what I have reviewed; leaving the rest to Alex

herbert_core/utilities/classes/@IX_dataset_3d/sliceomatic.m
tgperring: I made the change because disp2str(par) can create huge output if e.g. par is a 1000x1000 numeric array it produces thousands of lines of print out, which in the error message would equate to thousands of red lines in the command window. It was that problem i wanted to avoid. However, looking more carefully at the class of variable `par` it is always a row cell array, so in fact it will produce limited amount of output. Accordingly, I'll change the line back to the original in the code.

horace_core/spaghetti_plot.m
tgperring: Ah... my laziness exposed. I got rid of a call to this function in many other places, but just copied the original into this function as the last bastion of a piece of code that also made me uncomfortable. Mea culpa and I am properly chastised. I'll get rid of it altogether in the revised PR.

herbert_core/utilities/classes/@IX_dataset_3d/sliceomatic_overview.m
tgperring: Done.

herbert_core/utilities/classes/@IX_dataset_2d/private/plot_twod.m
```diff
@@ -1,247 +1,289 @@
...
-            if isnumeric(par{5}) && isscalar(par{5}), zlo=par{5}; else bad=true; end
-            if isnumeric(par{6}) && isscalar(par{6}), zhi=par{6}; else bad=true; end
-        else
-            zlims=false;
+        alternate_cdata_ok = false;
+        w1_data_name = '';
+        w2_data_name = '';
+    case 'surface'
+        if ~is_string(default_fig_name)
+            default_fig_name = 'Herbert surface plot';
```
abuts, line 159: checking if property undefined using invalid property type is a bit convoluted.  Do you really want to distinguish between setting figure name to '' or [] at input? 
Ok -- I see that you are really do... Looks like overcomplication.

herbert_core/graphics/genieplot.m
```diff
@@ -0,0 +1,320 @@
...
+    %   EXAMPLE: >> genieplot.set('colors',{'b','r','g'})
+    %
+    % To retrieve a property:
+    %   >> value = genieplot.get(<property_name>)
+    %
+    %   EXAMPLE: >> val = genieplot.get('marker_sizes')
+    %           val =
+    %                6
+    %
+    % Get a structure with all properties:
```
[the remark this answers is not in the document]
tgperring: I used a singleton template model (from Yair Altman) that is consistent with the way graphics handle properties are altered or retrieved, and is also very simple in its usage: you directly refer to the name of the singleton (BTW, you can access all the properties as a structure to see what they are by using S=genieplot.get(), again consistent with Matlab graphics handle objects).

I've expanded further in a comment to the issue #1873 .

herbert_core/utilities/classes/@IX_dataset_2d/private/plot_twod.m
tgperring: Matlab allows figures with no name - when the corresponding figure handle property `Name` is set to ''. Therefore the undefined status has to be set to something different than ''. It was something that I found a bit irritating, for the same reason as you find it baroque, but when I thought it through at the time there were circumstances I could see that caused problems if I didn't distinguish between the two cases. For that reason I settled on unassigned as [] (in just the same way as Matlab sets cell array elements to [] when you do something like myvar = cell(1,5) ).

herbert_core/utilities/classes/@IX_dataset_3d/sliceomatic.m
abuts: I am moving suggested solution to this problem to ticket #1875

abuts (approved):
This is huge change, very difficult to review comprehensively.  Waste majority of it is ceartainly for the best.

I spent day of testing it and have not found any substantial problems, though may still look for longer, not sure it is reasonable time to spent. Do not expect any substantial bugs here.

 My suggestions, minor fixtures and small modifications are provided in folow-up tickets #1873 and #1875

tgperring:
@abuts regarding your comment about the inclusion of private functions in genieplot.m: the whole of the classdef file started out as only about 60 lines i.e. very lean. Having the private functions that are core to the operation as a singleton  in the classdef made it very readable as all the code was in one compact place. Later, I decided to add rigorous checks to the `set` methods, and it rather ballooned as the number of properties grew. If I was starting again now I would have the checks in separate private file(s), but I'd still leave the private functions that are core to the singleton implementation in the classdef (namely `getInstance()` and `initialise`) for readability.

You are right that there is style question to be agreed, as there are with other style questions about classes in general, but as you say, this should be a separate discussion.

cmarooney-stfc (approved):
approving as requested

Issue #1871, opened 2025-05-16
validate_horace problems when running tests locally.


tgperring:
`validate_horace` is in principle a very useful utility for running tests suites locally before making a pull request. It allows control over the test suites that are run through direct specification of one or more test folders, or multiple test folders choice of criteria (Herbert only, Horace only, all except MPI etc.),. However, there are a number of shortcomings that mean that it is in practice not very helpful:

- The test folders and test suites list in `validate_horace` needs to be manually edited to keep it synchronised with `CMakeList.txt` for the Jenkins tests that are performed when making a pull request. This has on several occasions (most recently just before writing this issue) resulted in `validate_horace` failing because a test folder name change has not been passed into `validate_horace`, or not running all tests.
- At the time of creation of this issue one or two particular test suite names have been cut-n-pasted from `CMakeList.tx`t but the syntax, while correct for the underlying call to `runtests` is incorrect for `validate_horace` and so causes a crash of `validate_horace`.
- Changes made when running the tests in one test folder to some global states e.g. Matlab warnings are sometimes retained for the entire remainder of the running of validate_horace, in contrast to the behaviour of the Jenkins tests, which runs each test folder in its own Matlab instance. This typically results in different test failures in the two cases.
- Useful documentation about troubleshooting the use of 'validate_horace' when there are multiple versions of Horace on a local machine (a common situation for developers) is not collected in one place, resulting in the tests being run with erroneous Horace configurations.

The consequence is test results that can depend on which tests are run and which order, or even failures of 'validate_horace'. These behaviours are also not necessarily reproducible. This issue is to address these problems.

In addition, there is a problem unrelated to 'validate_horace' which is that within a test folder the constructors for different test suites in some cases have interdependencies, or even (apparently) between different test folders. This is the subject of issue #1870.

tgperring:
Issue solved by PR 1903

PR #1872, opened 2025-05-19
unit test dependencies when running multiple tests in Matlab session

### LINKED ISSUE #1870: test interdependencies when tests are run from a Matlab session
when Horace tests are runned from `cmake`, each test folder is run under its own Matlab version and they are passing

If one tries to run all tests using `validate_horace`, many of them start to fail.

Possible reason for that may be the fact that test constructors modify some global Horace settings and these settings contradict each other. 

The purpose of this ticket is to remove test classes interdependencies and fix `validate_horace` script 

Related issue: #1871

abuts:
Partially addresses Re #1870.

There may be multiple issues with tests interdependencies not yet identified, but this PR fixes four of them.  Other issues may be identified later and will be fixed by other PR, so merging this should not close #1870

1) changes in unit tests folders not reflected in `validate_horace`. Other ticket (#1871) should deal with this properly but there is hotfix, making unit tests folders equivalent to the current state of `cmake` lists.

2) Strange design, tests and behaviour of `horace_paths` Usage of global variable for paths is descouraged. Changes to class implement and provide access to persistent variable for performance, though if this is really necessary and performance critical routine remains unanswered.

3) test_warnonce -- the test was failing if run in runtests. It looks like the  issue is that it uses `evalc` function and persistent variable initialized in function executed through `evalc` is different from the same variable,  initialized in a code.

4) bug in incorrect storing/restoring of global containers state in unit tests. The bug may be reproduced as follows:

4a -- ensure global containers are not empty and is in the specific state namely: 

```MATLAB
unique_obj_store.instance
ans = 
  unique_obj_store with properties:
              n_types: 9
            typenames: {1×9 cell}
    n_unique_per_type: [7 1 4 268 75 32 6 4 5]
```
This state is normally achieved when `validate_horace` was run over whole suite.

4b. Change to  `Horace\_tests\tests_sqw_file`  and run
```MATLAB
runtests('test_data_block:test_put_get_sqw_block', 'test_set_instrument_data:test_head_data')
```

This test is failing. After applying fix,  the test passes and `validate_horace` passes this test and all tests in `test_sqw_file` pass. The global containers state above is retained by `_tests\tests_sqw_file`


_test/common_functions/set_temporary_global_obj_state.m
```diff
@@ -15,19 +15,11 @@
...
+%unique_obj_store.instance().clear('IX_samp');
+%unique_obj_store.instance().clear('IX_detector_array');
 
 cleanup_handle = onCleanup(@() restore(mem_contents));
-% stop changes from being stored on disk.
-config_instance.saveable = false;
-
-for i = 1:2:numel(varargin)
-    config_field = varargin{i};
-    value = varargin{i + 1};
```
tgperring, line 28: Comment: I'm puzzled that these lines were ever present!

_test/shared/matlab_xunit/xunit/TestRunDisplay.m
```diff
@@ -169,14 +169,20 @@ function testRunFinished(self, did_pass)
 
             if did_pass
                 result = 'PASSED';
+                fh = 1;
```
tgperring, line 172: Change: 'fh ' should be expanded to a more meaningful name e.g. 'output_stream' or 'screen_outputID' I'd do something with a comment like:
`        screen_outputID = 1   % standard output to screen`
and
`        screen_outputID = 2   % standard error output to screen`

_test/common_functions/set_temporary_global_obj_state.m
abuts: I was intending to change configuration alongside with containers state, but this found to be unnecessary

_test/test_utilities/test_warnonce.m
```diff
@@ -6,13 +6,13 @@
         end
 
         function setUp(~)
-            warnonce('clear', 'HORACE:panic:panic')
-            warnonce('clear', 'HORACE:panic:arr')
+            evalc("warnonce('clear', 'HORACE:panic:panic')");
```
tgperring, line 9: Question: Why does this need to be changed to being run by `evalc` (and similarly in the rest of `function setUp` and `function tearDown)`? The reason why it is done elsewhere in the tests is to capture the output that would normally be written to the screen by haveing a return argument `out`: e.g. in `out = evalc("warnonce('HORACE:panic:panic', 'Panic at the disco')");`. if the return argument is not needed i don't understand why the `evalc` is needed.

herbert_core/horace_paths.m
```diff
@@ -1,58 +1,64 @@
...
-    methods
-
-        function herbert_path = get.herbert(obj)
-            global herbert_path
-            if ~exist('herbert_path', 'var') || isempty(herbert_path)
 
-                herbert_path = obj.get_folder('herbert_init');
-            end
+    methods
+        function her_ = get.herbert(obj)
```
tgperring, line 40: Change required: Please use more informative variable names. In this case, replace `her_` with `herbert_path` as the name gives an indication of what the method is doing. This is really important to help newcomer(s) understand what the code is doing.

herbert_core/horace_paths.m
```diff
@@ -1,58 +1,64 @@
...
+            her_= obj.herbert_;
         end
-
-        function horace_path = get.horace(obj)
-            global horace_path
-            if ~exist('horace_path', 'var') || isempty(horace_path)
-
-                horace_path = obj.get_folder('horace_init');
-            end
+        function hor_ = get.horace(obj)
```
tgperring, line 43: Change required: replace `hor_` with `horace_path`

herbert_core/horace_paths.m
```diff
@@ -1,58 +1,64 @@
...
+            hor_= obj.horace_;
         end
-
-        function root_path = get.root(obj)
-            global root_path
-            if ~exist('root_path', 'var') || isempty(root_path)
-                root_path = fileparts(obj.get_folder('horace_init'));
-            end
+        function rootp = get.root(obj)
```
tgperring, line 46: Change required: See earlier comment about variable names, and change the name here accordingly.

tgperring (dismissed):

tgperring (approved):

tgperring (approved):

PR #1877, opened 2025-05-23
Tests dependencies

abuts:
Solves Re #1870 and addresses substantial part of Re #1871

All Horace tests constructors were checked for dependencies. The identified dependencies have been removed. 

Changes from this PR  remove all dependencies between unit tests caused by changes in global state settings, i.e. configurations or warning states.

After applying these changes one shoule be able to use `validate_horace`  for testing Horace.

This PR is convenient to review looking at each commit separately. There are big number of very similar changes to unit test classes and each commit reflects such changes.

There are multiple changes to new `GlobalStateTrace` class done during debugging  to found most convenient way of using it and enhance its usage.

### DIFF
--- a/_test/common_functions/clean_up_file.m
+++ b/_test/common_functions/clean_up_file.m
@@ -9,7 +9,7 @@ if ~exist(file_path, 'file')
 end
 
 % Get list of all open file IDs and the names of the files
-if verLessThan('MATLAB','24')
+if verLessThan('MATLAB','24.1')
     all_open_fids = fopen('all');
 else
     all_open_fids = openedFiles();


_test/common_functions/clean_up_file.m
```diff
@@ -9,7 +9,11 @@ function clean_up_file(file_path)
 end
 
 % Get list of all open file IDs and the names of the files
-all_open_fids = fopen('all');
+if verLessThan('MATLAB','24')
```
tgperring, line 12: Required change: Make the Matlab version explicit. R2024a is 24.1, R2024b is 24.2. Do you mean 24.1 or 24.2? It is confusing because '24' is not a formal Matlab version (see the Table in https://en.wikipedia.org/wiki/MATLAB). In my opinion, to make the condition readable, it is much better to use 'R2024a' or 'R2024b', except possibly where code is performance critical, which is not the case here.

_test/shared/matlab_xunit/xunit/TestSuite.m
```diff
@@ -142,6 +142,13 @@ function add(self, component)
             else
                 self.TestComponents{end + 1} = component;
             end
+            difr= GlobalStateTrace.instance.trace();
```
tgperring, line 145: Required: Please use a more informative name for the variable `difr`. I do not have any clear picture of what the variable is.

_test/shared/matlab_xunit_ISISextras/GlobalStateTrace.m
```diff
@@ -0,0 +1,131 @@
...
+            % Trace changes in global configuration and return difference
+            % in configuration if global configuration changed.
+            if ~obj.trace_enabled
+                difr = [];
+                return;
+            end
+            current_state = struct();
+            current_state.warning_state = warning;
+
+            hc = hor_config;
```
tgperring, line 57: Required change: This singleton, that is is part of the xunit ISIS_extras, refers specifically to the state of three Horace configuration classes. This is bad design. The unit test framework should be generic, and not refer to specifics of one particular project - in this case Horace. The tests that are specific to Horace should be isolated into Horace, and the test framework should be independent of Horace. I realise that we have the xunit test framework in Herbert, but this is contingent. For example, we might want to move to using the Matlab test framework (the Matlab  unit test framework based on the xunit test framework that we got from the file exchange originally.

I think you should find a way of having a generic optional state test in xunit_ISISextras, which can then be directed to point to a specific instance e.g. by overloading a generic method.

tgperring (approved):
Following a conversation with Alex about the principles of the changes, I have dipped into various test suites to understand the changes, and can see that they are essentially careful temporary disabling of warnings and resetting on exit from tests, and similarly with Horace configurations. It all looks fine to me.

We agreed that the idea of a generic globalStateTrace class was valuable, but that engineering it to be generic rather than specific to Horace requires a little bit of thought about how to implement (just) one function call robustly, and should become a separate ticket. It is not needed for this PR, as it was a tool used to help locate the Horace state/configuration changes and that work has been done now. Its value is for assisting in locating future problems in Horace, and test suites for other projects.

**One preferred change, but not required**: Alex, please can you look over the changes to the tests and make the naming of variables that hold configurations more consistent. For example
- In test_change_crystal_bragg_coarse.m, two object properties to hold configurations have been added:`hpc_config_holder` and one called `hpc_conf_to_restore`; for consistency the second should be `hpc_config_to_restore`

- In test_equal_to_tol_dnd.m a previously exisiting property is `old_config`, and an new one has been added called `cfg_holder`. The new one should be renamed `config_holder` for consistency and clarity.

It may sound trivial, but consistency with naming - just like abbreviations that are readable and meaningful -  makes it easier for people to read what the code is doing.

PR #1880, opened 2025-05-28
Fix bug in converting Python output in get_syms.m

mducle:
Fixes a bug in `get_syms.m` - this calls a Python script which writes out all the symmetry operators for a particular spacegroup - it is intended to work with the symmetrisation routines, e.g.

```matlab
[info, symops] = get_syms('Fd-3m');
proj = ortho_proj([1 1 0], [-1 1 0]);
w1 = cut_sqw(sqwfile, proj, [0.01], [-1, 1], [-1, 1], [0.2], symops)
```

However, the output from the Python script needs to be converted into Matlab which it was not. This PR fixes the conversion of the Python outputs.

----

To test - check that:

```matlab
[info, symops] = get_syms('Fd-3m');
```

returns an `info` string and a cell array of `symop` objects. Before this PR this will give an error.

You will need to have Python installed and `pyenv` set up to a Python interpreter with `spglib` installed. On DAaaS the euphonic environment `/mnt/ceph/auxiliary/excitations/isis_direct_soft/euphonic_env/bin/python` has this.

----

This fix is applied on IDAaaS already.

### DIFF
--- a/horace_core/admin/get_syms.m
+++ b/horace_core/admin/get_syms.m
@@ -25,12 +25,19 @@ function [info, sym] = get_syms(arg)
 % in the Python environment you supply to Matlab using the pyenv function.
 % On DAaaS you can use the Mantid environment:
 %
-%>>pyenv('Version', '/opt/mantidworkbenchnightly/bin/python')
+%>> pyenv('Version', '/opt/mantidworkbenchnightly/bin/python')
 %
 % or Euphonic one:
 %
-%pyenv('Version', '/mnt/ceph/auxiliary/excitations/isis_direct_soft/euphonic_env/bin/python')
+%>> pyenv('Version', '/mnt/ceph/auxiliary/excitations/isis_direct_soft/euphonic_env/bin/python')
 %
+% Usage example:
+% [info, symops] = get_syms('Fd-3m');
+% proj = line_proj([1 1 0], [-1 1 0]);
+% w1 = cut(sqwfile, proj, [0.01], [-1, 1], [-1, 1], [0.2], symops)
+%
+% Note, that ranges defined in "cut" have to be transformed to each other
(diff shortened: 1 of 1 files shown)


horace_core/admin/get_syms.m
```diff
@@ -1,4 +1,4 @@
-function [info, ops] = get_syms(arg)
+function [info, sym] = get_syms(arg)
```
cmarooney-stfc, line 1: please add comments for meaning of input and output arguments

I cannot see where sym is set

horace_core/admin/get_syms.m
```diff
@@ -13,10 +13,12 @@
...
     end
 
-    [info, ops] = py.get_syms.main(arg);
+    outpy = py.get_syms.main(arg);
+    info = string(outpy(1));
+    evalc(string(outpy(2));
     if nargout < 2
         disp(info)
-        disp(ops)
+        disp(sym)
```
cmarooney-stfc, line 21: cannot see where sym is set

horace_core/admin/get_syms.m
mducle: The [python code](https://github.com/pace-neutrons/Horace/blob/259dba6ae7525c8e93464e5fff988040eea484e5/horace_core/admin/get_syms.py#L892) returns a string in the second return value which begins `sym = {` - this is evaluated by `evalc` into the function's workspace and returned directly to the user as `sym`.

horace_core/admin/get_syms.m
mducle: Added comment above addressing this.

abuts (dismissed):
Its a pity that spglib is python only (is it the part of standard python installation?), but this is certanly other ticket and here I see very useful function many may want to use.

abuts (approved):

Issue #1873, opened 2025-05-20
add a standard signleton interface to `genieplot` class


abuts:
`genieplot` is a singleton which defines Horace/Herbert graphics. 

For some reason, it has specific to `genieplot`  interface to work with singleton and is not providing standart one, present in all Horace singletons.

This ticket is to add standard signleton interface to the gieneplot one, for pieple with different experience to work with familiar interface to the class.

The advantage of the standard interface is that calling `genieplot.instance()` gives one access to all graphic properties unlike custom graphics interface which requests user to know about its properties.

tgperring:
I agree that uniformity has merits. However, before making any changes, we should discuss what uniformity we want to aim for. I used a singleton template model (from Yair Altman) that is consistent with the way graphics handle properties are altered or retrieved, and is also very simple in its usage: you directly refer to the name of the singleton  (BTW, you can access all the properties as a structure to see what they are by using S=genieplot.get(),  again consistent with Matlab graphics handle objects).

The template I came across in Horace, `Singleton.m`, didn't give me much clue as to how it was used in practice. I downloaded the original version from the Matlab file exchange (Bobby Nedelkovski), which looks somewhat different, and it came with an example usage script that seemed a bit baroque, where you have to grab an instance before one can change a property, which personally I find a bit confusing. Perhaps the design pattern has more flexibility I haven't appreciated?

We should discuss the pros and cons. What I think would be a bad idea is to implement two different interfaces to genieplot.

abuts:
The change is better even for main reason:

If I type:
`genieplot.get` I got a structure, chaning which is useless. I need to remember to do `genieplot.set('propertyname',somethng)` to change something
if I do:
`genieplot.instance`, a class is returned. For this class you may use standard Matlab settings interface, tab autocompleteon etc.

Additional advantages and pro-arguments are:

1) there are bunch of other singletons in codebase, accessable through `class_name.instance`. Why somebody would want to use this singleton differently? 
2) old genie interface is there. If you want to use `genieplot.set('propertyname',somethng)` -- you welcome to do that. 
3) The standard interface code relies on Matlab's standard setters/getters. For old genie interface you need to write pretty complex custom glue code doing `obj.(property_name) = value` which looks uglier. It is there and is unit tested, so no problem of keeping it there, but if the question is on removing one interface, I would prefer the Matlab's standard one.

PR #1882, opened 2025-05-30
1873- singleton interface only

abuts:
This solves Re #1873 and adds standard singleton interface previously removed by PR #1874.

The changes have the following advantages:

1. It makes `geineplot` singleton class similar to any other current Horace classe, exposing its public interface to user and hiding private properties behind Matlab `get/set` class methods, unlike previous implementation, which hides public class interface.
2. it exposes `instance` method, similarly to any other singleton class present in Horace (there are at least 5 there already). A new Horace user does not need to learn new Toby's specific way of working with singleton, which provides no benifits but different. 
3. A handle, returned by `instance` method may be operated like any other graphical handle availible in modern Matlab.
4. The change retains previous interface which provides `get/set` interface reminiscent to older MATLAB graphics.
5. The change retains  useful finding (like `reset` method) which is recommended to distribute to other Horace singleton classes.


_test/test_IX_classes/test_plot_singleton.m
```diff
@@ -25,391 +31,611 @@
...
-            
+
             obj.S_default = orderfields(S);
         end
-        
+
+        function setUp(obj)
+            obj.current_graphics_settings = genieplot.get();
+        end
+        function tearDoan(obj)
```
tgperring, line 41: The function name is mispelt: it should be "tearDown". I think this means that the function will not  be run and so the graphics settings will not be reset.

_test/test_IX_classes/test_plot_singleton.m
```diff
@@ -2,16 +2,22 @@
...
     % Tests the setting, getting and resetting of the properties
-    
+    %
+    % In addition, tests the interoperability between the two different syntaxes
+    % for use of genieplot.
+
     properties
         S_default       % Structure with default property values
+        % store current graphics settings in the structure back to revert it back
+        % after test completeon
```
tgperring, line 12: Misspelt: replace "completeon" with "completion"

_test/test_IX_classes/test_plot_singleton.m
```diff
@@ -2,16 +2,22 @@
...
     %
     % Tests the setting, getting and resetting of the properties
-    
+    %
+    % In addition, tests the interoperability between the two different syntaxes
+    % for use of genieplot.
+
     properties
         S_default       % Structure with default property values
+        % store current graphics settings in the structure back to revert it back
```
tgperring, line 11: Unnecessary repetition of "back"; also, the graphics settings will be stored in the object property, not a structure? Change:
"Store current graphics settings in the structure back to revert it back..."
to
"Store current graphics settings so they can be to reverted..."

tgperring (changes requested):

PR #1890, opened 2025-07-10
nopix in pageop

abuts:
This is formal PR which have already been merget to master but made just to keep bookmarking going. 

The ticket should be about very simple change to code, when `sqw_op` algorithm may run with `-nopix` option, which, similarly to `-nopix` in `cut` would  drop used pixels and produce image only which would be much simple operation.


abuts:
This contribution has been merged from PR #1843

PR #1899, opened 2025-10-07
add to release instructions

cmarooney-stfc:
Update instructions for creating a release in line with Alex' briefing.

In progress

### DIFF
--- a/documentation/smg/10_build_release_process.md
+++ b/documentation/smg/10_build_release_process.md
@@ -156,6 +156,11 @@ On the build page, fill in the following fields:
 (the number of the build on the first line of Stage View), semi-colon; and
 on line 2 repeat for the `Release-windows-VM-Win10` pipeline.
    - release body: this should be a copy of the release notes.
+   - draft: (check-box) see below
+   - prerelease: (checkbox) see below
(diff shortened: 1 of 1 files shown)


abuts (changes requested):
The document describes release/bugfix process we were using when delivering Horace-4 and supporting Horace-3. 

Then indeed, bugs identified in Horace-3 need to be placed in separate bugfix branch provided to users and possibly ported or not ported into Horace-4 branch as Horace-4 was introducing breaking changes. 

Workplan for further Horace-4 development does not assume any breaking changes (at least as far as existing epics stand), so old release/bugfix process looks unnecessary complicated. 

We need a discussion if we indeed want to support it or follow simpler development process with released branch and nightly development branch (master) which includes bugfixes and new development features, released as soon as ready in master and available as nightly. As far as I am aware, all RSE projects  and MANTID follow this process. 

This simplifies process allows additional early testing of new features, faster bugfix releases and eliminates unnecessary philosophical questions like the one may be asked about Re #1907 -- is this a bug or a new feature and where it should go.


Also need to describe the process of upgrading documentation with releases, which is not included here. But this may be another ticket.

Issue #1908, opened 2025-12-16
Confusing input in `symop` and possible inefficiencies in its implementation when crytal lattice is non-orthogonal


abuts:
When you define `SymopRotation` operation, one defines rotation axis as e.g. `[1,0,0]` or `[0,0,1]` etc. A user who see this would assume that rotation axis is defined in `hkl` units. This is incorrect for non-rectangular lattices, as in fact in non-rectangular lattice, e.g. `[0,0,1]` axis provided for symop currently defines an axis, which is orthogonal to `[1,0,0],[0,1,0]` plane so it is not in `hkl (rlu)`
This is confusing, as user would expect `[0,0,1]` in non-orthogonal lattice to be parallel to $\vec{c^*}$.

In addition, offset for rotation (and may be folding) is defined in Crystal Cartesian system of coordinates (Angstrom and orthogonal), while interface and description suggests that it is `hkl`.
(this issue is sorted by Re #1922)

Probably need discussion on how to best present e.g. rotation axis in non-orthogonal (rotation plane defined by two vectors in `hkl`?)

This ticket is for fixing these issues and modifying `Symop` description to be valid for non-orthogonal lattices.

abuts:
finaly solwed by Re #1923

Issue #1822, opened 2025-02-05
global issue with symmetrization.


abuts:
User have identified issie with symmetriszation framework.  It looks like it is substantial bug to be fixed.

I have one more question. I tried to rebin a cut with symmetry operation and loose lots of data even with "0" as argument:

```MATLAB
dq = 0.005
r1 = cut_sqw(sqw_file, struct('u', [1.0,1.0,1.0], 'v', [1.0,1.0,-2.0], 'type', 'rrr', 'uoffset', [0.0, 0.0,0.0,0.0]), [0.5-dq, 0.5+dq], [-dq,dq], [-dq,dq], [-10.0,0.1,30.0], final_symmetry_operation_list)

plot(r1)
```

![Image](https://github.com/user-attachments/assets/0032118f-1ff5-4d18-862c-d7d476ca14ca)

```MATLAB
r2 = rebin(r1, 0.5)
```

![Image](https://github.com/user-attachments/assets/6891a2ee-de57-4342-82c8-78eabf80d6e1)

```MATLAB
r2 = rebin(r1, 0)

plot(r2)
```

![Image](https://github.com/user-attachments/assets/2585925d-83ef-4857-9675-2d57fe8faa89)

cmarooney-stfc:
briefing:

Rebin delegates to “cut”

Symmetrise transforms pixel coordinates to make it as though pixels had always been in their symmetrical positions, i.e. the positions in the irreducible section of q-space. Thus all pixels are now in the irreducible section; there are no pixels elsewhere in q-space. 
cm:[Picture showing -x-->x reflection symmetry; a pixel in the -x side of the diagram is moved to the +x side and is no longer in the -x side.]

My suspicion is that cut is somehow recomputing some [atts???] of these leading to pixels out of range.

Suggest setting up a small trial with generate_cube_sqw and seeing if issue can be reproduced. 

Following this inspect sqw coordinates pre and post cut.

cm:[Using generate_cube_sqw
e.g. as in test_sym_op::test_cut_sym_cube

final_sym_op_list not defined in query]

Issue #1668, opened 2024-05-08
Feature Request: Define `SymopReflection` as plane-normal


oerc0122:
Add option to define a `SymopReflection` from a plane normal.

PR #1919, opened 2026-02-06
`symop` in `cut` looses pixels and incorrect offset addressed.

abuts:
follows #1914; Solves Re #1822 

Issue complains about pixels not being propagated into symmeterized cuts appropriately and strangenes happening with offsets. The solution brings much more changes than expected as all pixels transformation ideology implemebted in Horace-4  were faulty and only pixel-image transformation in projections was correct for 1D and 2D images.

This contribution fixes the following issues:

1) Pixels coordinates are correctly modified within a `cut` to correspond symmeterized image., so if you do a `cut` from symmeterized `sqw` object, the `cut` can be perfomed and is correct. To achieve that, `Symop.transform_proj` has been substantially modified to generate correct transformation. The linear projection classes were modified to accept these transformation and incorporate them into transformation, applied within a projection during a cut operation.

2) total mix-up with coordinate systems used in symmetry transformation has been resolved. `Symop` obtained new property, which contains `b-matrix` used when necessary to convert from `rlu` to Crystal Cartesian coordinate system and vise-versa.

3) Offsets applied to `Symop` are now expressed in correct coordinate systems and properly propagated to projection when necessary. 

4) `Symop` are allowed again for `sqw` objects with non-orthogonal crystal lattice, though input for such lattices needs proper understanding and caution from user. This issue will be resolved later, as part of work on issues #1668 and #1908.


horace_core/symop/@SymopRotation/SymopRotation.m
```diff
@@ -57,99 +67,101 @@
...
+        function selected = in_irreducible(obj, coords, tolerance)
+            % Compute whether the coordinates in `coords` (Q) are in the irreducible
+            % set following the symmetry reduction under this operator
+            %
+            % For a rotation `R` about axis `n` of angle `theta`:
+            %
+            % For any `u` not parallel to `n` and v = R*u;
+            % The planes defined by UN, VN encapsulate the reduced region
+            % And thus any coordinate `q` from `Q` where
+            % q*(n x u) > 0 && q*(v x n) > 0
```
cmarooney-stfc: (1) checking that x is cross product. ^ would be less ambiguous.
(2) what is q*? please explain

horace_core/sqw/axes_blocks/@AxesBlockBase/private/bin_pixels_.m
```diff
@@ -95,7 +95,7 @@
 % collapse first dimension, all along it should be ok for pixel be ok
 if is_pix
     % Add filter to ignore pix, used previously through symmetry and binning operation
-    ok = all(coord>=r1 & coord<=r2,1) & pix_cand.detector_idx >= 0;
+    ok = all(coord>=r1 & coord<=r2,1) & pix_cand.detector_idx >= 0; % use explicit valye of selected pixels tag (negative det-id)
```
cmarooney-stfc:
```suggestion
    ok = all(coord>=r1 & coord<=r2,1) & pix_cand.detector_idx >= 0; % use explicit value of selected pixels tag (negative det-id)
```

cmarooney-stfc (changes requested):

horace_core/symop/validate_and_generate_sym.m
```diff
@@ -90,4 +99,37 @@
...
+% check if all offsets, present on symmetry transformation are the same
+if iscell(sym)
+    offsets = cellfun(@(x)(x.offset),sym,'UniformOutput',false);
+else
+    offsets = arrayfun(@(x)(x.offset),sym,'UniformOutput',false);
+end
+is_zero_offset = true;
+max_err = 4*eps('double');
+for off = offsets
+    same = all(abs(off{1}-offsets{1})<max_err);
```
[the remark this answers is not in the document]
abuts: no -- `off` is iterator over cellarray (see python influence), so each `off` is cell of 1 element running over all elements. `offsets` from other side is cellarray of number of elements so the expression is correct

horace_core/symop/@SymopRotation/SymopRotation.m
abuts: This is Jackob's text which looks pretty clear to me. I have added definitions for `x` and `*` for newbies.

cmarooney-stfc (approved):

Issue #1918, opened 2026-02-02
ADD on symmetry operations


abuts:
Current implementation of symmetry operation contains bunch of classes interacting with `symmeterisze_sqw` through `page_op` algorithm and `cut` algorithm through `projection`.  

The interaction is not-trivial so needs ADD to describe principles and the details of the design. 

The ticket is to provide this ADD.

abuts:
fixed by Re #1924

Issue #1850, opened 2025-04-01
Issues with symmetrisation EPIC


abuts:
Users complan and developers are uncertain about current implementation of symmetrization options in Horace. 

There are currently three ticket dedicated to the issue:
 
- [ ]  Re #1822  no pixels folding in symop.
- [ ]   Re #1848 rotation in `symmeterize_sqw`
- [ ]   Re #1849  inefficiency in `symmeterize_sqw`

More tickets are possible after deeper code investigation.
These ticket should be resolved and current symmetrisation operations should be independently validated.

PR #1924, opened 2026-03-10
symmetry docs

abuts:
Solves Re #1918. 

ADD and user documentation on current state of Horace symmetry code.

Should be merged after Re #1922; Re #1923, though does not depend on them but rather describes changes made there and in bunch of previous PR-s already merged to master. 

Ths PR completes Symmetrisation Epic Re #1850.


cmarooney-stfc (approved):
approving as it's documentation and corrections can be done as part of the 4.2 review process

Issue #1935, opened 2026-04-08
update spaghetti plot description.


abuts:
iron task involves generation of spaghetti plot pannels obtained using extensive data processing and plotting these pannels using `spaghetti_plot`. 

This possibility is currently available in `spagetti_plot` but existing description does not mention it.

This ticket is for updating `spaghetti_plot` description to mention possibility to plot set of `d2d` objects representing spaghetti plot pannels.

abuts:
sorted by Re #1936

PR #1936, opened 2026-04-10
spaghetti plot documentation update

abuts:
Addresses Re #1935.

Small updates to spaghetti plot documentation.


cmarooney-stfc (approved):
approving as it's documentation and corrections can be done as part of the 4.2 review process

Issue #1939, opened 2026-04-15
documentation to draw mask


abuts:
new subalgorighm `draw_mask` is available and this ticket is for describing this algorithm in release notes and user documentation.

abuts:
sorted by Re #1940

PR #1940, opened 2026-04-16
`draw_mask` algorithm documentation

abuts:
Addresses Re #1939

User documentation about `draw_mask` algorithm.


cmarooney-stfc (approved):
approved as documentation - to be checked after use

Issue #896, opened 2022-11-11
Enable full mexing options in cut


abuts:
Cut using mex is currently disabled (only some mex cut options are working) 

need to fix and enable test mex/nomex for cut. 

Sample test for that is: 
`test_algorithms->test_cut:test_you_can_take_an_out_of_memory_cut_with_tmp_files_with_mex`

abuts:
Closed through PR #1946

PR #1944, opened 2026-05-07
mex transformation when binning pixels

abuts:
Partially addresses Re #896 

Long standing ticket to perform binning pixels using C++ code while doing cut with linear projection, like it was done in Horace-3.  

Mex code is fully implemented to work within linear projection and used when `use_mex` option is enabled. 
Contains performance tests and unit tests. 

The work is not completed yet as OMP parallelization is not yet enabled, but code runs 2-times faster so, despite being not so good as expected, should not be frowned-off. 

OMP will be addition, in addional templates and function tables so this PR contains completed chunk of work.


abuts:
superseeded by Re #1964 which contains and slightly improves thses changes + adds OMP modes

Issue #1950, opened 2026-05-20
small utility to check omp code performance


abuts:
productivity of OMP code depends on number of OMP threads used. There is need to identify the performance of OMP code as function of number of threads and used algorithm. 

There is already performance test in test_mex_nomex folder which investigates performance of omp code as function of algorithm. 

This ticket is for creating on the basis of existing perfomance test the function which would test performance of an algorithm as the function of number of OMP threads which execute this algorithm.

abuts:
Solved by Re #1951

PR #1951, opened 2026-05-20
omp thread testing utility

abuts:
Addresses Re #1950 

Simple piece of code, which calculates performance of OMP cutting algorithm as function of number of threads used in processing data. 

Figure below shows result of using this utility to investigate performance as function of number of threads and size of dynamic step (this variable is not plotted and is colour-parameterised. Non-optimal value surely harms performance) for one of subalgorithms on a machine with 20 cores. Results for another subalgorithm are provided for comparison (yellow line) 

image


cmarooney-stfc (commented):
appreciating that this is a testing utility rather than production code; nevertheless it is difficult to follow without reverse engineering it, because the new functions introduced in this PR have no comments describing what they do, where they fit in, or what this arguments are, and how the various arguments are processed. It would be good to have some commenting to make clear how this works for anyone picking it up in future

cmarooney-stfc (approved):

PR #1955, opened 2026-05-30
Updated build system to run newer MATLAB versions on Jenkins

abuts:
The change modifies Horace build scripts to use `gcc` compiler appropriate to corresponding Matlab version(s).
It was identified that Matlab R2024a needs native gcc10 and higher Matlab versions need gcc-12. 
The Jenkins script, `cmake` scripts and unix shell modified appropriately to support this type of build. 

`rand_like` generator works differently in Matlab 2021a and later Matlab versions. The test which validates this generator modified accordingly. 
Jenkins on Linux successfully runs against higher Matlab versions.


mducle (commented):
So, I had a look at some old data and it turns out that the problem mentioned in this issue was already solved and the `run_id`s were already de-conflicted, and the issue in [this doc](https://pace-neutrons.github.io/Horace/unstable/manual/Data_diagnostics.html#instrument-view-cut) doesn't occur for continuous rotation data.

Now, the "meat" of the change here is [line 365 of gen_sqw.m](https://github.com/pace-neutrons/Horace/pull/1955/changes#diff-61a99f5c39c6b6123c5344dbab0665daf1823732b43285e68b41fae5eb950ea0L365-R368) which forces `run_files` to pass through `update_duplicated_rf_id` - _but_ this is already done `rundatah` [in line 67](https://github.com/pace-neutrons/Horace/blob/master/horace_core/sqw/%40rundatah/rundatah.m#L67) which is called here in [line 286](https://github.com/pace-neutrons/Horace/blob/master/horace_core/sqw/%40rundatah/rundatah.m#L67) to create the `run_files` cell-array - hence all the `run_id`s in `run_files` are already unique (before the current changes).

So, I don't think this change is needed but then again it doesn't really hurt...

cmarooney-stfc (commented):
I cannot follow what is going on here. The only Horace functionality is in gen_sqw, making the update_duplicated_rf_id call unconditional. It does not look as if it does what you say in the PR comment description. Please could you make it clear *where* the additional run_ids are added. I would expect to see modifications to update_duplicated_rf_id or similar here as well.

Separately, you make it clear that this is a hack which will need refining later. Please make a ticket for this work if you have not already done so, and reference it in the PR description.

Thanks

abuts:
Duc comment state that despite this hack make no harm, its already there. From this point of view, this PR is unnecessary. I do agree with this. 

At the moment I use this PR as a placeholder for fixing build for new MATLAB versions as it curries out no actual changes to code. If succeed, I will reformat the PR into the fixture for build system. 

Ticket for proper solution of the problem is already there and I am working on it. 

At the end it should not contain any change to MATLAB code -- only cmake and Jenkins changes.

cmarooney-stfc (approved):

